PR #1, opened 2016-06-04
Testing

jschueller:
This fixes a bug with CBLAS enabled where an unused string replace cmake command is given an empty argument:
 ```
-- CBLAS enable
CMake Error at CBLAS/testing/CMakeLists.txt:19 (string):
  string sub-command REPLACE requires at least four arguments.
Call Stack (most recent call first):
  CBLAS/testing/CMakeLists.txt:62 (add_cblas_test)
```

I added a travis config file in case you want to use continuous integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: cpp
+
+addons:
+  apt:
+    sources:
+      - george-edison55-precise-backports # cmake
+    packages:
+      - cmake
+      - cmake-data
+      - gfortran
+
+os:
+  - linux
+  - osx
+
+script:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jschueller** [comment]
This fixes a bug with CBLAS enabled where an unused string replace cmake command is given an empty argument:
 ```
-- CBLAS enable
CMake Error at CBLAS/testing/CMakeLists.txt:19 (string):
  string sub-command REPLACE requires at least four arguments.
Call Stack (most recent call first):
  CBLAS/testing/CMakeLists.txt:62 (add_cblas_test)
```

julielangou:
Awesome...i did not know about Travis...i just enabled it.
We will see how it goes.
Thank you so Julien for all your help
Julie

PR #2, opened 2016-07-04
Improve c make building

hjmjohnson:


hjmjohnson:
@sebastien-villemot @julielangou 

I've provided two patches to 1) clarify and improve the dual building with cmake and Makefiles, 2) provide more testing for cases when the cmake found compilers for fortran and C are not compatible with each other.

I have a few notes of other relatively minor cmake build improvements that I will hold off on until this set of changes is accepted/rejected.

Thank you,
Hans

LAPACKE/CMakeLists.txt
```diff
@@ -58,7 +55,7 @@ else (USE_XBLAS)
 endif(USE_XBLAS)
 
 lapack_install_library(lapacke)
-INSTALL( FILES ${LAPACKE_INCLUDE} DESTINATION include )
+INSTALL( FILES ${LAPACKE_INCLUDE} ${LAPACK_BINARY_DIR}/include/lapacke_mangling.h DESTINATION include )
```
hjmjohnson: I had missed adding  ${LAPACK_BINARY_DIR}/include/lapacke_mangling.h to the installed files in the earlier commit.  

This was a problem when using default cmake "find_package(LAPACKE REQUIRED)" from an external package.

julielangou:
Thank you very much Hans for your contribution.
Looks like the CMAKE build was successful in our test environment.

julielangou:
Not very sure why, but the author of this commit is me while it should be you.

hjmjohnson:
@julielangou I am still the author of the commit that fixes the bug, you are the author of the merge commit.    Git is careful about giving attributes (blame) to each person involved.  You merge commit is giving you attribution for accepting the changes provided by me.

PR #3, opened 2016-07-05
BUG: Fix pathing to include files

hjmjohnson:
During building of external packages, only the build tree or install
tree files should be used (the source tree may not be available from
the binary package).  Set to use the build tree locations for the
configuration files.


hjmjohnson:
@sebastien-villemot @julielangou Is this the correct github repository for proposing code fixes?

Thanks,
Hans

langou:
Yes, this is the correct github repository for proposing code fixes to LAPACK. Reviewing them now. Thanks! Cheers, Julien.

langou:
Hi Hans, Yes, yes, we have your two pull request. This is the correct github repository. Julien.

hjmjohnson:
Spectacular!  Thank you for considering these patches.

PR #9, opened 2016-07-27
?LALN2: fix doc for the D2 parameter

vladimir-ch:
D is a NA×NA diagonal real matrix. Therefore, its (2,2) element D2 is not used if NA=1. There is no relation to NW.

Also reported at http://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=13&t=4997


langou:
Thanks Vladimir. Good catch! Cheers, Julien.

PR #14, opened 2016-07-27
COMP: LAPACKE and CBLAS require C language support

hjmjohnson:


jschueller:
-1: they're already enabled as needed in subdirs

hjmjohnson:
@jschueller I missed that!  Thanks.

PR #15, opened 2016-07-27
Prevent bad build configs

hjmjohnson:


julielangou:
thank you @hjmjohnson , this is a very good idea. Unfortunately, the Travis build failed on this.
Also, I would maybe make it Platform independent.
A lot of our users are using CMake when building the LAPACK library on a Windows machine.

hjmjohnson:
Travis builds were not as robust as they need to be.  I'm working on that.   Tests seem to be fragile.

hjmjohnson:
@julielangou Just a follow up comment now that I have a keyboard.  

1) The travis builds were failing, becuase they are doing insource builds, and this patch explicitly prevents insource builds!

2) Travis builds were not testing CBLAS, and CBLAS testing succeeded on clang build, but failed on gcc build.  I think the failure is related to the the testing harness, not a problem with the code.  I'm currently investigating.  Once the travis builds in #16 are made more robust, then this should pass.

3)  This code was taken from the www.itk.org project where it has been used for many years on many platforms (including windows).  It has been extensively tested in ITK, VTK, ANTs, BRAINSTools, Slicer3D, and many other projects.  I don't think there are any windows issues associated with it.

julielangou:
Awesome! Thank you for your hard work.

PR #16, opened 2016-07-27
Let travis submit dashboards

hjmjohnson:


hjmjohnson:
@julielangou This should be ready to merge now.

Issue #17, opened 2016-07-28
CBLAS tests failing with GCC compiler.


hjmjohnson:
These tests are failing on GCC compilers 4.9.1
version 4.9.1 20140922 (Red Hat 4.9.1-10) (GCC) 

https://travis-ci.org/Reference-LAPACK/lapack/jobs/147765341

	 13 - CBLAS-xscblat1 (Failed)
	 14 - CBLAS-xscblat2 (Failed)
	 15 - CBLAS-xscblat3 (Failed)
	 16 - CBLAS-xdcblat1 (Failed)
	 17 - CBLAS-xdcblat2 (Failed)
	 18 - CBLAS-xdcblat3 (Failed)
	 19 - CBLAS-xccblat1 (Failed)
	 20 - CBLAS-xccblat2 (Failed)
	 21 - CBLAS-xccblat3 (Failed)
	 22 - CBLAS-xzcblat1 (Failed)
	 23 - CBLAS-xzcblat2 (Failed)
	 24 - CBLAS-xzcblat3 (Failed)

I verified these failures on a RHEL6 machine using These tests are failing on GCC compilers 4.9.1
version 4.9.1 20140922 (Red Hat 4.9.1-10) (GCC).

PR #22, opened 2016-07-29
Enable cblas testing

hjmjohnson:


hjmjohnson:
@julielangou @jschueller @vladimir-ch 

This patch set only turns on tests that were previously disabled.  It demonstrates that there are long standing test failures (that were just not run previously).

It would be great if these test could be made to work, and then the test run on the dashbaords.

Hans

julielangou:
Related to #17

langou:
Hi Hans, I am going to close this `pull request`. Thanks! Julien.

PR #28, opened 2016-08-04
?TREXC: fix range checks for IFST and ILST

### LINKED ISSUE #25: ?TREXC: unavoidable error return when N=0
The documentation and the code indicate that N=0 is allowed. However, in that case the error checking for IFST (and ILST) leads to an unavoidable failure:
```
      ELSE IF( IFST.LT.1 .OR. IFST.GT.N ) THEN
         INFO = -7
      ELSE IF( ILST.LT.1 .OR. ILST.GT.N ) THEN
         INFO = -8
      END IF
```

vladimir-ch:
Fixes #25


langou:
Thanks Vladimir!

PR #31, opened 2016-08-05
LAPACKE_?trexc_work: check ldq conditionally on compq

### LINKED ISSUE #26: LAPACKE_?trexc_work: ldq must be always >= n
In row-major case, `LAPACKE_?trexc_work` requires `ldq >= n` even when `compq` is not `'v'`, that is, when the matrix Q is not referenced. The `ldq` parameter is always used conditionally only if `compq` is `'v'`, so the `ldq >= n` check should be also done only in that case.

In ?TREXC the corresponding check is:
```
IF( LDQ.LT.1 .OR. ( WANTQ .AND. LDQ.LT.MAX( 1, N ) ) ) THEN
   INFO = -6
```

vladimir-ch:
Fixes #26


langou:
Thanks Vladimir!

PR #36, opened 2016-08-29
Fix incorrect call to GEMM in ctrevc3.f and ztrevc3.f: alpha should have complex type

eugenechereshnev:


langou:
Thanks Eugene! Thanks a lot. This is a great fix. This helps a lot.

PR #40, opened 2016-09-01
External definitions not called

zerothi:
I have opened this issue on the forum:

http://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=13&t=5022

However, I just saw that you have moved development here. Hence, I could easily create the PR.


langou:
Thanks for removing the extra xLASET external declarations. 
And thanks for adding a P to DGEQR2 in the documentation indeed. 
All good. Thanks a lot. Julien.

PR #50, opened 2016-09-15
maint: removed unused variables

zerothi:
Removed 3 unused variables from 2 different routines. 

I have compiled with `-Wunused` and these where the only ones which popped up.


langou:
Thanks!

PR #51, opened 2016-09-16
?TREVC?: fix definition of left eigenvectors in docs

vladimir-ch:


langou:
Good point Vladimir. 

On a side note. Oh my. I see that we have comments with this `**` for denoting exponent. There was a debate about whether we should use `^` or `**` a while ago. Julie homogenize everything and doing some `grep`-ping now I see that we are again inconsistent. Something to thing about. 

Thanks Vladimir for the correction.

PR #54, opened 2016-09-18
bug: fixes always failing tests for *TREXC which was erroneous tests

zerothi:
There where checks for *TREXC for the argument of N = 0 with a return value
of -7. However, the documentation of the *TREXC routines specify that
N = 0 is a valid argument with instant return.

Hence the checks have been removed and the quick returns established in the
affected routines.

Also comments for [dz]errec.f files are fixed.

**NOTE**: I have instantiated quick returns as `N.LE.1` which is equivalent to the [SD]TREXC checks. Please remark that these codes does not check for `N<0` and return an error code accordingly. This check is beyond what I am comfortable inserting as it may not be desired.

### DIFF
--- a/SRC/ctrexc.f
+++ b/SRC/ctrexc.f
@@ -182,7 +182,7 @@
 *
 *     Quick return if possible
 *
-      IF( N.EQ.1 .OR. IFST.EQ.ILST )
+      IF( N.LE.1 .OR. IFST.EQ.ILST )
      $   RETURN
 *
       IF( IFST.LT.ILST ) THEN
--- a/SRC/ctrexc.f
+++ b/SRC/ctrexc.f
@@ -57,6 +57,7 @@
 *> \verbatim
 *>          N is INTEGER
 *>          The order of the matrix T. N >= 0.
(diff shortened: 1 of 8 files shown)


zerothi:
Perhaps I should expand on the test removed. 

It seemed that the test should return an error if `IFST>N`. However, from the documentation it seems that the intend for `N=0` is an immediate return.

Either way, I think the changes made in `[cz]trexc` are still correct.

langou:
Hi Nick, 

(1) Allowing the case `N=0` => I am OK with this. The quick return as you did is good. We should probably add in the description of  IFST/ILST that the case N=0 is handled by a quick return. 

(2) You remove the test for `N=0`. Why don't we `simply` pass on `-1` instead of 0 in the existing test? that way we make sure that the test is there for `N<0`. Does it make sense? 

Otherwise allowing the case N=0 in [cz]trexc is a good idea. Yes. Can you revise? 

Cheers, 
Julien.

zerothi:
1) I have added a description at the `N` argument. Would this suffice?

2) Done. :)

PR #55, opened 2016-09-18
Doc: fix xyyEQUB documentation

cconrads-scicomp:
- fix description of purpose, parameters
- cite the paper "Scaling by Binormalization"

### DIFF
--- a/SRC/cheequb.f
+++ b/SRC/cheequb.f
@@ -37,12 +37,11 @@
 *> \verbatim
 *>
 *> CHEEQUB computes row and column scalings intended to equilibrate a
-*> Hermitian matrix A and reduce its condition number
-*> (with respect to the two-norm).  S contains the scale factors,
-*> S(i) = 1/sqrt(A(i,i)), chosen so that the scaled matrix B with
-*> elements B(i,j) = S(i)*A(i,j)*S(j) has ones on the diagonal.  This
-*> choice of S puts the condition number of B within a factor N of the
-*> smallest possible condition number over all possible diagonal
+*> Hermitian matrix A (with respect to the Euclidean norm) and reduce
+*> its condition number. The scale factors S are computed by the BIN
+*> algorithm (see references) so that the scaled matrix B with elements
+*> B(i,j) = S(i)*A(i,j)*S(j) has a condition number within a factor N of
+*> the smallest possible condition number over all possible diagonal
 *> scalings.
 *> \endverbatim
 *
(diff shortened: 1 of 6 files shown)


cconrads-scicomp:
# Introduction

There are multiple matrix equilibration routines within LAPACK, e.g., xPOEQUB, xHEEQUB, or xGEEQUB, and each of them uses a different mathematical method. In this comment, I will explain that xSYEQUB, xHEEQUB
* is a misnomer, and
* that these methods are indeed an implementation of the BIN algorithm.

# xSYEQUB, xHEEQUB is a Misnomer

The difference between xyyEQU and xyyEQUB is usually that the latter avoids round-off errors by returning scaling matrices who entries are powers of the floating point radix and the [xGEEQUB](http://www.netlib.org/lapack/explore-html/dd/d9a/group__double_g_ecomputational_gade6ad2e443f1d6686471d5861a2c3e6e.html#gade6ad2e443f1d6686471d5861a2c3e6e) documentation explains the difference very well. The [xPOEQU](http://www.netlib.org/lapack/explore-html/d1/d7a/group__double_p_ocomputational_gae83b3e94a9fce827c8a7c547d134d9e6.html#gae83b3e94a9fce827c8a7c547d134d9e6) and [xPOEQUB](http://www.netlib.org/lapack/explore-html/d1/d7a/group__double_p_ocomputational_ga771aa24ef6fe09d905031c56adf64a62.html#ga771aa24ef6fe09d905031c56adf64a62) obey the same naming rule but the documentation does not highlight the difference so you have to look at the source code to understand the difference. Obviously, the "Purpose" paragraph of the xPOEQU/xPOEQUB documentation was copied verbatim to xSYEQUB, xHEEQUB. Seeing that xSYEQUB, xHEEQUB do not return radix powers, the routines are misnomers.

I am currently preparing a patch for the xPOEQUB documentation that highlights the difference to xPOEQU.

# xSYEQUB, xHEEQUB implement the BIN algorithm

If you look at the patches, you will notice that the
reference to the Livne/Golub paper was already present for xSYEQUB in
LAPACK 3.4.0. Now if you view SRC/dsyequb.f and SRC/cheequb.f
side-by-side it is obviously the same code with `ABS(x) = |x|` replaced by `CABS1(x) = |Re(x)| + |Im(x)|`. Clearly, all these functions implements the same algorithm and the BIN
algorithm does generalize to Hermitian matrices (see §5 of the [Livne/Golub paper](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.3.1679)).

Furthermore, at least DSYEQUB is indeed an implementation of the BIN
algorithm:
* dsyequb.f, lines 236-265 implement the computation of the average $
\overline{\beta}$ in the BIN paper,
* dsyequb.f:265-272 is the computation of Eq. (7) in the paper,
* dsyequb.f:273 is the termination test from step 3 of Algorithm BIN in
the paper, and
* the loop in dsyequb.f:275-318 implements steps 4, 5, 6 from Algorithm
BIN, e.g., Eq. (12) from the paper is solved in lines 276-281 and the
variable `D` in line 281 is $\Delta$ in §3.4 and used to detect
unscalable matrices.

langou:
Thanks Christoph for taking the time to fix the fix documentation of xyyEQUB and taking the time to explain on github. Julien.

PR #60, opened 2016-09-29
LAPACKE: fix wrong direction in LAPACKE_?larfb_work

### LINKED ISSUE #59: LAPACKE_?larfb_work: wrong direction
Apart from #37, there is another bug in `LAPACKE_?larfb_work`. When transposing the input matrices, in the last, fourth, case when `storev` is `'r'`, the check for `direct` should compare it to `'b'`, not to `'f'`.

vladimir-ch:
Fixes #59.


langou:
Thanks Vladimir.

PR #63, opened 2016-10-01
Faster GSVD with QR factorizations, 2-by-1 CS decomposition (REAL(8) only)

christoph-conrads:
The generalized singular value decomposition can be computed directly with xGGSVD, xGGSVD3 (see [LAWN 46](http://www.netlib.org/lapack/lawnspdf/lawn46.pdf)) or by computing QR factorizations and a 2-by-1 CS decomposition. In the following, let us call the former approach _direct GSVD_ and the latter _QR+CSD_. In my master's thesis I used both approaches and it turned out that the QR+CSD was always signifcantly faster than the direct GSVD for pairs of square matrices with up to 3600 columns (see Section 3.5 of my [master's thesis](https://christoph-conrads.name/masters-thesis/)).

I am convinced that the speed difference is persistent across platforms because the iterative phase of the solver is much faster for QR+CSD. The direct GSVD reduces both matrices A, B to upper triangular form before the iterative solver xTGSJA is called whereas the iterative phase of QR+CSD operates on a 2-by-1 isometric matrix Q^T = [Q1^T, Q2^T]^T, where Q1, Q2 are bidiagonal (A isometric <=> A^T A = I). Since the matrix Q is always well-conditioned, the CSD solver xBBCSD needs only few iterations to compute a solution. For comparison, the maximum number of iterations per singular value/eigenvalue is
- 6 for the CSD (see variable MAXITR in xbbcsd.f, line 356),
- 30 for the symmetric eigenvalue problem with Francis QR (MAXIT, dsteqr.f:154), and
- 40 for the GSVD (MAXIT, dtgsja.f:402).

I implemented the QR+CSD solver named xGGQRCS for real double precision floats as a proof of concept. Computing the GSVD for a pair of 1000-by-1000 matrices and entries uniformly distributed in [-1, +1) takes 27.9s with the QR+CSD (and debugging statements enabled) and 243s with the direct GSVD on my computer. You can repeat these measurements by calling the program `bin/ggsvd_vs_ggqrcs` in the build directory.

This pull request contains the following changes:
- a real double precision QR+CSD solver in `SRC/dggqrcs.f` with debugging statements,
- tests using C++14, [Boost.Test](http://www.boost.org/doc/libs/1_62_0/libs/test/doc/html/index.html) and [Boost uBLAS](http://www.boost.org/doc/libs/1_62_0/libs/numeric/ublas/doc/index.html) in `TESTING/cpp` , and
- a benchmark program in `TESTING/cpp/ggsvd_vs_ggqrcs.cpp`.

If there is enough interest, then I will implement the other variants of the QR+CSD solver ({C,S,Z}GGQRCS) and remove the debugging code.


poulson:
This is very impressive work, but I am skeptical of any performance conclusions based on boost.ublas, as it is notoriously slow. It might be worthwhile to switch the tests to depend on either the reference BLAS, OpenBLAS, ATLAS, BLIS, MKL,  or one of their ilk.

cconrads-scicomp:
@poulson Boost uBLAS is used only for (defect) testing and more convient matrix construction. The solver itself is implemented in pure Fortran, independent of Boost, and uses the reference BLAS in this repository.

poulson:
In that case, I would recommend removing the boost dependency in the PR...Boost it is a very heavy dependency to only be used for convenience in one place.

For what it's worth, the Travis build is failing due to boost not being installed by default. Thankfully, should you want to keep the boost dependency, Travis whitelists boost (https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise), so it would be a trivial change to the `.travis.yml` to install it.

gaming-hacker:
what's status on this?

cconrads-scicomp:
This is not top priority for me currently, especially since the maintainers have not commented on this algorithm yet.

poulson:
@cconrads-scicomp Please don't be discouraged or take it personally. It's great work and there is a significant time investment involved in handling this type of change set.

gaming-hacker:
@cconrads-scicomp  i pulled into my repo and see how it goes

gaming-hacker:
i think this last test would have passed if an updated apple compiler was choosen as the gcc choosen is very old.  travis bombed with
error: invalid value 'c++14' in '-std=c++14'

martin-frbg:
@gaming-hacker did you get around to doing any tests ? While I do not feel competent to review this code it should be possible to accept it into OpenBLAS to give it more exposure if desired (at least OpenBLAS does not strive to provide a _reference_ implementation of anything). (@christoph-conrads did you receive any criticism of your thesis work in the meantime, and/or did any further publications arise out of it ?)

gaming-hacker:
@martin-frbg i ran some basic tests on the algorithm, nothing intense or complex after building with the clang-5.0 compiler and well nothing has bombed or crashed at this point.

christoph-conrads:
@martin-frbg Soon I will have a computer at home again and then I will finish the implementation for matrices with complex values, say within the next month. No further publications arose out of it because my code is based on a transformation well known in the literature, see for example Section 8.7.4 and Section 8.7.5 in the fourth edition of _Matrix Computations_ by Golub and Van Loan or in [this paper](https://conservancy.umn.edu/handle/11299/1875) by Bai. So far, no one implemented this because there was no implementation of the CS decomposition in LAPACK until recently and I happened to compare these two approaches.

PR #66, opened 2016-10-06
Improve pkgconfig file generation

turboencabulator:
The prefix and libdir lines at the top of the .pc files are variable
declarations.  Set these variables appropriately, reference them in the
rest of the file, and prevent cmake from expanding ${}-style references.
Switch back to @prefix@ and @libdir@ for compatibility with autoconf.

Make the descriptions consistent and update the URLs.


julielangou:
Thank you

PR #72, opened 2016-10-17
Add [out] in docs

mgates3:
Correct docs in *larcm, *lacrm, *lasd0 where an array is modified to be [out].


langou:
Thanks Mark!

PR #73, opened 2016-10-17
make SY/HE imat match LAWN 41 (different than PO)

mgates3:
LAWN 41 (Table 1) lists these matrix types for PO, PP and SY, SP, HE, HP:

```
                            PO, PP      SY, SP, HE, HP
Diagonal                    1           1
Upper triangular            
Lower triangular            
Random, k = 2               2           2
Random, k = sqrt(0.1/eps)   6           7
Random, k = 0.1/eps         7           8
First column zero           3           3
Last column zero            4           4
Middle column zero          5           5
Last n/2 columns zero                   6
Scaled near underflow       8           9
Scaled near overflow        9           10
Random, unspecified k               
Block diagonal                          11
```

Note that, going from PO => SY, matrix 6 => 7, ..., 9 => 10.

zchksy.f seems to follow this numbering for SY, zeroing entries for matrices 3-6. However, zlatb4.f, which zchksy.f uses, had the exact same code for both PO and SY. Thus for SY, matrix 6 ended up both with cond = sqrt(0.1/eps) _and_ zeroing Last n/2 columns.

This fixes that, making latb4 have a separate condition for SY, SP, HE, HP type, with imat cases changed from 6, 7, 8, 9 to 7, 8, 9, 10.

### DIFF
--- a/TESTING/LIN/clatb4.f
+++ b/TESTING/LIN/clatb4.f
@@ -340,12 +340,10 @@
             ANORM = ONE
          END IF
 *
-      ELSE IF( LSAMEN( 2, C2, 'PO' ) .OR. LSAMEN( 2, C2, 'PP' ) .OR.
-     $         LSAMEN( 2, C2, 'HE' ) .OR. LSAMEN( 2, C2, 'HP' ) .OR.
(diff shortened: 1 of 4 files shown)


julielangou:
Thank you

PR #74, opened 2016-10-19
Tsqr

sydha:
Tall skinny and short wide routines added.


langou:
Thanks Syd. We will have a look. Julien.

zerothi:
May I come with a few comments:

1) the .DS_store file should be removed from the revisions
2) The *_aasen routines have been removed from the Makefile, however, they should remain, no?
3) It seems like the ilaenv has been copied from a previous revision (i.e. no merging), version numbers etc. have changed to previous values, also particularly the 'ESV' check is removed in this revision.

julielangou:
I did not look at the code yet, but CMake Makefile needs to be modified too.

julielangou:
I did not look at the code yet, but CMake Makefile needs to be modified too. 
Since Travis is using the CMake Build, tests are failing too.

PR #88, opened 2016-11-19
Revert " adding the 2stage symmetric eigenvalue routines drivers checking"

julielangou:
Reverts Reference-LAPACK/lapack#83


poulson:
What was the reason for the reversion?

julielangou:
I have a couple of concerns in term of the code but I just deleted the revert...so we are going to move forward and improve the code in the coming days...

PR #93, opened 2016-12-06
bug: fixed compilation error

cmoha:


langou:
Hi @cmoha, this is great. Thanks. This is same as your pull request #87 and commit
`748287abe5e80a8bd141417acbae1bea9f362fde`. Sounds good to me. Julien.

cmoha:
OK, I'll dig into these files to find any other bugs.

PR #94, opened 2016-12-06
fix for some bugs

cmoha:
This is a supplementary PR to #87 & #93 
- fix for wrong definition of some voids.
- C/fortran binding.
- 2 reduntant variables.


julielangou:
Thanks a lot!

PR #104, opened 2016-12-29
Patches applied to Debian package for LAPACK 3.7.0

sebastien-villemot:


langou:
Thanks Sébastien. All good with me. Cheers, Julien.

PR #107, opened 2016-12-29
Fix complex to int conversion

jschueller:
This PR fixes these:
```
lapack/LAPACKE/src/lapacke_zsytri_3.c: In function ‘LAPACKE_zsytri_3’:
lapack/LAPACKE/src/lapacke_zsytri_3.c:63:5: error: aggregate value used where an integer was expected
     lwork = (lapack_int)work_query;
```


langou:
I see. Thanks for the pull request. I think that there are similar problems in lapacke_zgelq.c, lapacke_zgemlq.c, lapacke_zgeqr.c, lapacke_zhetri_3.c, lapacke_zstegr.c, lapacke_zstemr.c, and lapacke_zsytri_3.c. And I suspect ditto for the C versions as well. I will try to fix these as well as a follow up. Thanks. Cheers, Julien.

PR #108, opened 2017-01-04
Readme markdown

nschloe:
There's still some dubious information in there, e.g.,

> A list of known problems, bugs, and compiler errors for LAPACK is
maintained on netlib.

but this is definitely a step up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,29 +1,29 @@
-==================
-LAPACK README FILE
-==================
+# LAPACK
 
-VERSION 1.0   :  February 29, 1992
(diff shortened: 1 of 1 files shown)


langou:
Thanks Nico. Marking down the README file is a good idea. I realized now that we should have done it earlier. So thanks for the pull request. The README.md looks much nicer than the README on the project frontpage. Cheers, Julien.

nschloe:
Sure. Note that it's probably worthwhile for you to go through the contents once again.

PR #113, opened 2017-01-21
Split SRC_OBJ list to avoid build problem with mingw/msys 

### LINKED ISSUE #111: LAPACKE build problem with mingw
Over at OpenBLAS we received a report https://github.com/xianyi/OpenBLAS/issues/1056 stating that mingw fails to build LAPACKE as the list of object files in LAPACKE/src/Makefile is too long for the ar on that platform to handle. Would you consider a PR to split the object list as it is currently done in https://github.com/xianyi/OpenBLAS/blob/develop/lapack-netlib/LAPACKE/src/Makefile ?

martin-frbg:
Fixes #111, "argument list too long" error from msys ar

### DIFF
--- a/LAPACKE/src/Makefile
+++ b/LAPACKE/src/Makefile
@@ -32,9 +32,12 @@
 ##############################################################################
 # makefile for LAPACKE, used to build lapacke binary.
 #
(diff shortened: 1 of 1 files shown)


langou:
Super. Thanks @martin-frbg

PR #114, opened 2017-01-23
Assorted Makefile fixes

turboencabulator:
Here is part 3 of Makefile cleanup.  Also includes fixes for some Doxygen warnings, and removes a stale reference to libm.


langou:
Thanks @turboencabulator.

PR #119, opened 2017-02-04
Linking fixes

turboencabulator:
Round 4 of makefile cleanup.  This time the theme is linking problems.

- Fix more omitted or duplicated source files, mostly when building with certain flag combinations.
- Avoid linking transitive dependencies (overlinking).
- Add a version to the CBLAS and LAPACKE shared libs.


langou:
Thanks @turboencabulator

tonytheodore:
> Avoid linking transitive dependencies (overlinking).

The change to `Requires.private` in the *.pc.in files doesn't really avoid overlinking since the libraries are always required and now must be manually added to the pkg-config invocation. I thought the point of pkg-config was to pick up transitive dependencies?

turboencabulator:
Does your application use both LAPACK and BLAS routines?

If yes, then you should always link against both libraries (`pkg-config --libs lapack blas` to yield "-llapack -lblas").  Otherwise your application is underlinked, basically just hoping that its BLAS dependency is fulfilled by -llapack which might not always be the case.  This is probably the situation for why you're having to add extra libraries.

If it just uses LAPACK, then it's only necessary to add -llapack (with `pkg-config --libs lapack`), and that library knows about its own dependency on -lblas (and potentially others) and everything works out.  Unless it doesn't know about its own dependencies (for example with static libs or other file formats that don't have that dependency info), then you would add the `--static` flag to pkg-config to get all its transitive dependencies.  You don't want to unnecessarily add -lblas and others where they're not needed, otherwise your application's dependencies become a mess.

In both cases, you shouldn't have to care if LAPACK internally uses BLAS.  You only have to care whether your application uses LAPACK, BLAS, or both.

tonytheodore:
Thanks for the explanation! That makes a lot more sense now. I don't use these libraries myself, just maintain cross-compiled mingw packages for [MXE](http://mxe.cc). We have link tests to ensure the provided configuration works and to act as examples of usage. It happens that the failing tests use routines from multiple libraries.

My only minor question (being unfamiliar with the API), is why the last example in http://www.netlib.org/lapack/lapacke.html#_calling_code_cgeqrf_code_and_the_cblas ("Calling CGEQRF and CUNGQR to compute Q with workspace querying") needs to also link `-lblas`. I can see the others (lapacke,lapack,cblas) but can't see any direct usage of blas.

langou:
There is no direct usage of BLAS. There is indirect usage of BLAS. LAPACK and CBLAS relies on BLAS. There should be no direct usage of LAPACK either. The code only directly uses CLAPACK and CBLAS (for the interfacing), it indirectly uses LAPACK and BLAS.

tonytheodore:
Got it now, thanks. The extra libs aren't actually required, removing them from pkg-config call works for both shared and static builds.

PR #120, opened 2017-02-04
lapacke_*(bb|or|un)(csd|bdb): forward calls to LAPACK without conversion

echeresh:
1) Ignore TRANS parameter for LAPACKE since it has the same meaning
   as matrix_layout parameter.
   TRANS = 'T' means matrices are handled in row-major format.
   TRANS != 'T' means matrices are handled in col-major format.
2) So conversion from/to row-major layout can be removed in LAPACKE.
   All the cases can be forwarded to LAPACK calls with corresponding TRANS parameter.
3) nrows_* variables can be safely removed. NaN checkers for
   different TRANS values can be called with corresponding layout.

LAPACKE calls are forwarded in the following way:
trans != 'T', col-major: call LAPACK(trans = 'N') - col-major
trans != 'T', row-major: call LAPACK(trans = 'T') - row-major
trans  = 'T', col-major: call LAPACK(trans = 'T') - row-major
trans  = 'T', row-major: call LAPACK(trans = 'T') - row-major

### DIFF
--- a/LAPACKE/src/lapacke_cbbcsd.c
+++ b/LAPACKE/src/lapacke_cbbcsd.c
@@ -47,17 +47,18 @@ lapack_int LAPACKE_cbbcsd( int matrix_layout, char jobu1, char jobu2,
     lapack_int lrwork = -1;
     float* rwork = NULL;
     float rwork_query;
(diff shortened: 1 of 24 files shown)


julielangou:
Thank you Eugene

PR #127, opened 2017-02-19
LAPACKE: fix wrong number of columns in ?ormlq

### LINKED ISSUE #122: LAPACKE: bad transpose in lapacke_?ormlq_work
The matrix `a` is not transposed correctly in `lapacke_?ormlq_work` when row-major. Its size depends on `side`:
```
k×m if side == 'l'
k×n if side == 'r'
```
However, when doing the transpose, the number of columns is fixed to `m`:
```
LAPACKE_dge_trans( matrix_layout, k, m, a, lda, a_t, lda_t );
```

Also, the allocation of `a_t` can be simplified by using `r`, although `r` is an unusual variable name for the number of columns.

vladimir-ch:
Fixes #122


langou:
Super Vladimir. Thanks!

PR #128, opened 2017-02-21
LAPACKE: fix wrong matrix size in lapacke_?ormbr

### LINKED ISSUE #121: LAPACKE: bad transpose in lapacke_?ormbr_work
The matrix `a` is not handled correctly in `lapacke_?ormbr_work`. It's size depends on `vect` but the distinction is not applied consistently.

The correct size is:
```
nq×min(nq,k) if vect == `q`
min(nq,k)×nq if vect == `p`
```

However,
- `lda` is checked as
```
if( lda < MIN(nq,k) )
```
independently of `vect`.
- `a` is allocated as
```
if( LAPACKE_lsame( vect, 'q' ) ) {
          a_t = (double*)LAPACKE_malloc( sizeof(double) * lda_t * MAX(1,k) );
```
Here, `lda_t` has the correct value, but `MAX(1,k)` is wrong.
- `a` is transposed as
```
LAPACKE_dge_trans( matrix_layout, r, MIN(nq,k), a, lda, a_t, lda_t );
```
Here, `r` has the correct value, but the number of column is not always correct.

vladimir-ch:
Fixes #121

### DIFF
--- a/LAPACKE/src/lapacke_dormbr.c
+++ b/LAPACKE/src/lapacke_dormbr.c
@@ -42,7 +42,7 @@ lapack_int LAPACKE_dormbr( int matrix_layout, char vect, char side, char trans,
     lapack_int lwork = -1;
     double* work = NULL;
     double work_query;
-    lapack_int nq, r;
+    lapack_int nq, ar, ac;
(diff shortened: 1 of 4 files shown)


langou:
Thanks Vladimir. Thanks Intel. Cheers, Julien.

PR #129, opened 2017-02-21
Added Lapack code coverage

Gjacquenot:
This PR adds Lapack code coverage with codecov services.

It tells us that 82% of Lapack is covered with existing tests.

How it works:
It uses gcov to create coverage data, and continuous integration with Travis sends this data to codecov service. codecov parses the result and proposes a HTML view of the coverage.

Once this PR accepted, one should create a codecov account and update the README.md for the coverage badge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,25 +13,50 @@ os:
   - linux
   - osx
 
+env:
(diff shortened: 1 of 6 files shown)


langou:
Super. Merci. J.

PR #130, opened 2017-02-22
Fix missing comma on EXTERNALs that span multiple lines

martin-frbg:
In a few files from 3.7.0, an EXTERNAL declaration of subroutines missed a comma before the continuation line. This caused concatenated names to appear as unresolved external dependencies in the objects and library at least when compiled with ifort. See https://github.com/xianyi/OpenBLAS/issues/1069


langou:
Hi Martin, Thanks for adding the four extra commas. There were needed. Compilers these days are way too permissive! ;) Much better with the extra commas. Best wishes, Julien.

PR #131, opened 2017-02-22
TESTING/EIG: fix spurious EXTERNAL references to nonexistent functions

martin-frbg:
According to https://github.com/xianyi/OpenBLAS/issues/1069#issuecomment-281664864 the Windows version of the Intel Fortran compiler treated these as unresolved external references

### DIFF
--- a/TESTING/EIG/zchkhb2stg.f
+++ b/TESTING/EIG/zchkhb2stg.f
@@ -373,7 +373,7 @@
 *     ..
 *     .. External Subroutines ..
       EXTERNAL           DLASUM, XERBLA, ZHBT21, ZHBTRD, ZLACPY, ZLASET,
-     $                   ZLATMR, ZLATMS, ZHBTRD_HB2ST, ZSTEQR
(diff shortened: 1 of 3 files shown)


jeffhammond:
@martin-frbg I reproduced these issues on Mac or Linux, so they are not specific to Windows.

martin-frbg:
@jeffhammond thanks - I must have missed them in my test build then (very likely as I did not read the log that carefully after I got the library to build)

langou:
@martin-frbg: thanks for the fix.

Thanks to the Windows version of the Intel Fortran compiler for being rigorous and flagging these programming mistakes. I wish more compilers would be like this. 

@jeffhammond: I understand that you are saying that the Intel Fortran compiler is flagging these programming mistakes on linux and mac as well. Great. Good to hear. Thanks for checking. (If you meant another compiler, please let us know, and please let us know the flag used.)

Julien.

jeffhammond:
@langou Yes, you are correct.  I wasn't specific enough.  I reproduced the issues with Intel Fortran on one of my machines, which are all Mac or Linux.

PR #132, opened 2017-03-03
Update makefile prerequisites

turboencabulator:
Fix a few problems with missing or incorrect prerequisites, for instance some toplevel targets needing tmglib but would fail if it hadn't been built yet.  Various simplifications, such as using $^ to refer to all prerequisites, and giving all the clean targets consistent behavior.


langou:
Hi @turboencabulator, thanks a lot! Julien.

Issue #92, opened 2016-11-29
xLASWP: bug for negative INCX


vladimir-ch:
In DLASWP, `INCX` is the increment for `IPIV` and can be positive or negative. If negative, the swaps are applied from `K2` down to `K1`, otherwise from `K1` to `K2`. From the documentation it is understood that the same elements of `IPIV` are accessed in both cases.

However, the implementation behaves differently. When `INCX>0`, the first accessed element is `K1`, then `K1+INCX`, ... until `K1+(K2-K1)*INCX`. When `INCX<0`, the first accessed element is
```
1+(1-K2)*INCX
```
when in fact it should be
```
K1+(K1-K2)*INCX
```
It seems that the `1` is missing a `K` in front. All the calls to DLASWP with `INCX=-1` luckily use `K1=1` in LAPACK.

langou:
I have been starring at this line ( `1+(1-K2)*INCX` ) for quite some time (like 10 years) and could not figure out the meaning of this line and put in context with the subroutine description.

Yes, it really seems that you are correct and that the intent was to write: `K1+(K1-K2)*INCX` and what was written was `1+(1-K2)*INCX`. 

As you said I think we only use `xLAWSP` with `INCX = 1` or `INCX = -1`. So the other cases are not really tested.

In the `INCX = -1` case (only negative `INCX` case used in `LAPACK`), the line (`1+(1-K2)*INCX`) is good enough and returns the correct value (`K2`). But in other cases, (not used in `LAPACK`,) this line (`1+(1-K2)*INCX`) is clearly not correct and does not match the subroutine description.

This might be one of the oldest in LAPACK I know of. Oh my. I cannot check LAPACK 1.0 and others to see if the bug was present initially, or it was introduced at some point. The bug is in LAPACK 3.0. So since 2000 for sure.

langou:
Three proposed changes in xLASWP:

(1) In description, change 
`
          K2 is INTEGER
          The last element of IPIV for which a row interchange will
          be done.
`
to
`
          K2 is INTEGER
          (K2-K1+1) is the number of elements of IPIV for which a row
          interchange will be done.
`

(2) In description, change 
`
          IPIV is INTEGER array, dimension (K2*abs(INCX))
          The vector of pivot indices.  Only the elements in positions
          K1 through K2 of IPIV are accessed.
          IPIV(K) = L implies rows K and L are to be interchanged.
`
to
`
[...]

PR #133, opened 2017-03-07
Fix documentation for ?LASWP routine

echeresh:
It seems comments in LASWP routine doesn't fully correspond to implementation.

If I understand correctly, ?LASWP routine works as follows:
When INCX > 0:
```
DO I = K1, K2
   swap rows A(I,*) and A(IPIV(K1 + (I - K1)*INCX),*)
END DO
```

When INCX < 0:
```
DO I = K2, K1, -1
   swap rows A(I,*) and A(IPIV(K1 + (K1 - I)*INCX),*)
END DO
```

In both cases row I and `IPIV(K1 + (I - K1)*abs(INCX))` are interchanged. The only difference is that when INCX < 0 the pivots are applied in reverse order: `I = K2, K2-1, ..., K1`

### DIFF
--- a/SRC/claswp.f
+++ b/SRC/claswp.f
@@ -79,14 +79,15 @@
 *> \verbatim
 *>          IPIV is INTEGER array, dimension (K1+(K2-K1)*abs(INCX))
 *>          The vector of pivot indices. Only the elements in positions
-*>          K1 through K1+(K2-K1)*INCX of IPIV are accessed.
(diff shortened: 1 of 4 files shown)


langou:
Hi Eugene, Thanks for the correction in the comments of LASWP. All good with me. Feel free to have a look at the discussions of issue #92 and logs of commit #2ceda6db0f046b1b676030f9b6d92541e6222e3c as well. Julien.

PR #137, opened 2017-03-30
Add quick return for *LARR* routines

echeresh:


zerothi:
A comment before this is merged, the info variable is not defined in all cases of quick return. This, should be fixed before merge (see e.g. slarrr) 

Also, in some cases the default info is 1,while in others it is 0, what should it be for quick return?

echeresh:
@zerothi Thank you for catching that!

I updated *LARRK and *LARRR routines so that they set INFO to 0 before quick return.
I believe `INFO = 0` is more appropriate for both cases there because it is default for *LARRR:
```
*> \param[out] INFO
*> \verbatim
*>          INFO is INTEGER
*>          INFO = 0(default) : the matrix warrants computations preserving
*>                              relative accuracy.
*>          INFO = 1          : the matrix warrants computations guaranteeing
*>                              only absolute accuracy.
*> \endverbatim
```
and for *LARRK it means "Eigenvalue converged" - when matrix is empty, it is supposed that no any errors occurred:

```
*> \param[out] INFO
*> \verbatim
*>          INFO is INTEGER
*>          = 0:       Eigenvalue converged
*>          = -1:      Eigenvalue did NOT converge
*> \endverbatim
```

zerothi:
Great! :+1:

langou:
Looks good to me. Thanks Eugene. Thanks Nick.

PR #138, opened 2017-03-31
Fixed formatting (indentation, spaces, style) in some routines

kbergstr:
cgeevx, sgeevx: fixed indendation
cgejsv.f, zgejsv.f: fixed style
sbdsdc.f, dbdsdc.f: fixed spacing, corrected workspace indexing. When ICOMPQ == 2, LASR call at the end of the function uses the first 2N-2 elements of WORK, and so these elements should be unchanged.


langou:
Hi Kirana. Thanks for the patch. I reviewed it and I am OK with it. On line 311, we set WSTART at 1. (WSTART is useless in the (ICOMPQ.EQ.0) case.) On line  319, we shift WSTART by 2*N-1 in the ( UPLO.EQ.'L' ) case. This is to accommodate the storage of the Givens's rotations (CS and SN), see line 329-330, this is only relevant for the (ICOMPQ.EQ.2) case since for the (ICOMPQ.EQ.1) case, they are stored in Q. So all good with me. (And of course no problem with indentation, and such.) Thanks for everything. Julien.

PR #140, opened 2017-04-15
Replace last occurrences of `include` install paths

SoapGentoo:
* Distributions might need to change the header dir
* Also change the template paths in the `.pc` files to
  the idiomatic CMake `GNUInstallDirs` full paths,
  which are always correct, regardless of whether the
  user specified relative or absolute paths. This makes
  the build system somewhat easier and more idiomatic.

### DIFF
--- a/BLAS/blas.pc.in
+++ b/BLAS/blas.pc.in
@@ -1,5 +1,5 @@
-prefix=@prefix@
-libdir=@libdir@
+libdir=@CMAKE_INSTALL_FULL_LIBDIR@
+includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
 
 Name: BLAS
 Description: FORTRAN reference implementation of BLAS Basic Linear Algebra Subprograms
--- a/CBLAS/CMakeLists.txt
+++ b/CBLAS/CMakeLists.txt
@@ -28,7 +28,7 @@ endforeach()
 endmacro()
 
 append_subdir_files(CBLAS_INCLUDE "include")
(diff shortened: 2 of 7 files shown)


SoapGentoo:
Probably also interesting to @turboencabulator

langou:
This change is all good with me. Thanks. Julien.

SoapGentoo:
@langou another thing: I think lapack should be installed into its own directory (`${CMAKE_INSTALL_INCLUDEDIR}/lapack` or something). While with the solution in this PR we can modify it, I still think the default is bad as it pollutes the overall root of `include/` with a bunch of files. Modifying it upstream is better, as it will lead to a more homogenous ecosystem. Your thoughts?

turboencabulator:
Looks fine to me, though I'm not a fan of putting CMake-specific variables in a template and thus tying the template to one particular build system.

Regarding moving the headers to a subdirectory of `${CMAKE_INSTALL_INCLUDEDIR}`, are you thinking just one directory, or separate directories for CBLAS and LAPACKE?  Can CMake handle the latter?

SoapGentoo:
All build systems use different variables, so I don't see how this changes anything. It's not like you can port CMakeLists.txt to SCons in 2 minutes. Also, the build system is using GNUInstallDirs already, this just replaces the last occurrene of handcoded variables.

Yes, I think `BLAS` should be (default) installed into `/usr/include/BLAS`, `CBLAS` into `/usr/include/BLAS` etc. This prevents the pollution of the global include namespace and any build system using pkg-config will not notice anyway.

SoapGentoo:
@langou could we get this merged?

langou:
@SoapGentoo: Thanks for following up with me. This is merged. Cheers, Julien.

PR #154, opened 2017-05-21
Avoid conversion between layouts in lapacke_?lange_work

echeresh:
?LANGE routine is  memory-bounded. It means that additional copying in column-major layout will decrease efficiency of LAPACKE interface significantly. So it's better to avoid such a conversion and call Fortran routine for the same matrix but with swapped rows and columns.

### DIFF
--- a/LAPACKE/src/lapacke_clange_work.c
+++ b/LAPACKE/src/lapacke_clange_work.c
@@ -39,37 +39,41 @@ float LAPACKE_clange_work( int matrix_layout, char norm, lapack_int m,
 {
     lapack_int info = 0;
     float res = 0.;
+    char norm_lapack;
     if( matrix_layout == LAPACK_COL_MAJOR ) {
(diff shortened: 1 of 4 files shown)


langou:
Super. Thanks Eugene. I read your code. 
```

[ . . . ]

        if( LAPACKE_lsame( norm, '1' ) || LAPACKE_lsame( norm, 'o' ) ) {
            norm_lapack = 'i';
        } else if( LAPACKE_lsame( norm, 'i' ) ) {
            norm_lapack = '1';
        } else {
            norm_lapack = norm;
        }

[ . . . ]

        res = LAPACK_clange( &norm_lapack, &n, &m, a, &lda, work_lapack );
```
Super. So compute the 1-norm of the transpose for the infinite norm, and compute the infinite norm of the transpose for the 1-norm, else compute the norm of the transpose. All good with me. 

Cheers, Julien.

PR #156, opened 2017-05-22
lapack_testing.py: print results in one line

nareshkamboju:
The purpose of the patches is to print lapack results in a single line
test-case-name sub-tests-running pass/fail status and count

Modified sub test case name to single word
Nonsymmetric Eigenvalue Problem to Nonsymmetric-Eigenvalue-Problem

Signed-off-by: Naresh Kamboju <naresh.kamboju@linaro.org>

### DIFF
--- a/lapack_testing.py
+++ b/lapack_testing.py
@@ -232,12 +232,12 @@ for dtype in range_prec:
     letter+"bb","glm","gqr",
     "gsv","csd","lse",
     letter+"test", letter+dtypes[0][dtype-1]+"test",letter+"test_rfp"),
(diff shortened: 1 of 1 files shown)


langou:
Hi @nareshkamboju. Thanks for the change. Glad that this helps for your work on LAVA and having LAPACK building, running and testing on ARM64 target boards. Great. Thanks a lot. Cheers, Julien.

nareshkamboju:
@langou Thanks for all your support in merging this patch.
May i know when would be next release ?

langou:
We are shooting for a June 2017 release. This would be 3.7.1 and have the current git repo in. Cheers, Julien.

PR #160, opened 2017-06-03
more cleaned up comments

thielema:


langou:
Thanks!

thielema:
In commit
https://github.com/Reference-LAPACK/lapack/pull/160/commits/366f9b3541705231cfbb1724b221b8857ac768a8
I replaced in the file SRC/cuncsd.f in the parameter comment for `U1` the array dimension `(P)` by `(P,P)` because the parameter description suggested a `PxP` matrix.
Maybe this was inappropriate. I think the correct dimension must be `(LDU1,P)`.
At least, in most other functions the array dimensions in the FORTRAN code and in the Doxygen parameter comments match, except in the last dimension.
If this invariant shall hold universally, I may alter all functions accordingly.

langou:
I think that, by `U1 is COMPLEX array, dimension (P)`, the author of SRC/cuncsd.f meant `U1 is COMPLEX array, order (P)`. The size of U1 is clear by the description thereafter. But, you are correct: this is not standard. Yes, please, can you please propose a pull request with `(LDU1,P)`? Cheers, Julien.

PR #165, opened 2017-06-08
Some final line-break rearrangements that assist my parameter comment parser.

thielema:


langou:
Thanks. Glad that this is useful for your parser. Moreover, it reads better like this. I think we might get over the 72-character per line but since this happens in the comment section, I am not worried. Cheers, Julien.

PR #166, opened 2017-06-12
fixing the definition of "info" in Aasen's

iyamazaki:


langou:
Super. Thanks @iyamazaki

PR #167, opened 2017-06-12
Added and straightened Doxygen comments on BLAS parameters.

thielema:
The new comments are mostly copies from existing comments.
Please review them!


langou:
Hi, the two linux-box  travis-CI checks failed. Looking at the logs, it seems that this was because the servers were unresponsive (?) and since I reviewed the changes of the PR and the changes are in the documentation, all good with me. If others want to review, please go ahead. That said I am a little worried with the fact that continuous integration seems broken. (Maybe because we have a pre-release tag?) In any case, thanks for the changes, xheers, Julien.

PR #168, opened 2017-06-26
Added LAPACKE interfaces for [c/z]larcm and [c/z]lacrm

kbergstr:
We have added LAPACKE interfaces for these auxiliary functions to MKL in response to a customer request


langou:
Thanks Kirana!

PR #170, opened 2017-06-30
cblas: add include directory to target

NeroBurner:
add include directories to target. With this change in a cmake project one can link cblas with just the following

```
find_package(cblas CONFIG REQUIRED)
add_executable(foo foo.cpp)
target_link_libraries(foo cblas)
```

### DIFF
--- a/CBLAS/src/CMakeLists.txt
+++ b/CBLAS/src/CMakeLists.txt
@@ -120,5 +120,9 @@ set_target_properties(
   VERSION ${LAPACK_VERSION}
   SOVERSION ${LAPACK_MAJOR_VERSION}
   )
+target_include_directories(cblas PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
+  $<INSTALL_INTERFACE:include>
+)
 target_link_libraries(cblas PRIVATE ${BLAS_LIBRARIES})
 lapack_install_library(cblas)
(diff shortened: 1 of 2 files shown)


CBLAS/src/CMakeLists.txt
```diff
@@ -120,5 +120,9 @@ set_target_properties(
   VERSION ${LAPACK_VERSION}
   SOVERSION ${LAPACK_MAJOR_VERSION}
   )
+target_include_directories(cblas PUBLIC
```
ruslo: CMake minimum version is set to `2.8.10`:
* https://github.com/Reference-LAPACK/lapack/blob/e774b4339ebea509161ec9a75656662f64cd177e/CMakeLists.txt#L1

Command `target_include_directories` is not available:
* https://cmake.org/cmake/help/v2.8.10/cmake.html#section_Commands

This can be fixed by updating minimum requirement to 2.8.12:
* https://cmake.org/cmake/help/v2.8.12/cmake.html#command:target_include_directories

If update to `2.8.12` is acceptable, alternatively you can use `INCLUDES DESTINATION include` here:
* https://github.com/Reference-LAPACK/lapack/blob/e774b4339ebea509161ec9a75656662f64cd177e/CMakeLists.txt#L102-L107

Docs:
* https://cmake.org/cmake/help/v2.8.12/cmake.html#command:install

NeroBurner: I think `INCLUDES DESTINATION` is also a v2.8.12 feature. I didn't find it in the v2.8.10 documentation
https://cmake.org/cmake/help/v2.8.10/cmake.html#command:install

ruslo: > I think INCLUDES DESTINATION is also a v2.8.12 feature

Yes, that's what I meant by "If update to 2.8.12 is acceptable, alternatively..." :wink:

PR #171, opened 2017-06-30
lapacke: add include directory to target

NeroBurner:
add include directories to target. With this change in a cmake project one can link lapacke with just the following

```
find_package(lapacke CONFIG REQUIRED)
add_executable(foo foo.cpp)
target_link_libraries(foo lapacke)
```


NeroBurner:
rebased on current master to include the push of the minimum CMake version as stated in #170

PR #177, opened 2017-07-20
Added an ILAENV2STAGE, fixed ILAENV in [CZ]HEEVR_2STAGE

zerothi:
The `ILAENV` routine forces the entire LAPACK library to be compiled with `-frecursive` or the equivalent for non-GNU compilers.

This merge fixes this by adding a separate `ILAENV2STAGE` routine to query information related to the 2stage solvers, only. 
I.e. there are now 2 `ILAENV` routines for retrieving machine dependent variables.

Secondly, this merge fixes `ILAENV` for the 2stage routines `[CZ]HEEVR` for their block-size queries (`DSY*` was used). _I know this should have went in another PR, but that would inflict with this PR_.

Tests are also updated to enable these changes.

### DIFF
--- a/INSTALL/make.inc.gfortran
+++ b/INSTALL/make.inc.gfortran
@@ -20,9 +20,9 @@ CFLAGS = -O3
 #  should not compile LAPACK with flags such as -ffpe-trap=overflow.
 #
 FORTRAN = gfortran
-OPTS    = -O2 -frecursive
+OPTS    = -O2
 DRVOPTS = $(OPTS)
-NOOPT   = -O0 -frecursive
+NOOPT   = -O0
 
 #  Define LOADER and LOADOPTS to refer to the loader and desired
 #  load options for your machine.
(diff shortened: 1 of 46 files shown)


zerothi:
@langou gentle ping, could you please comment on whether this seems like a good idea or not?

My main motivation was the removal of the `-frecursive` flag. 

If this has no interest I can make a new PR with the fixes for the `ilaenv` calls in the complex routines and close this one.

langou:
Hi Nick, I was waiting on some feedback for some stakeholders in your fix. I did not hear from them. So this is all good. Yes, thanks for the fix. I do agree that this is better like this. Thanks for the PR. Cheers, Julien.

zerothi:
Great it could be of use. You are welcome. :)

pv:
This removes the `-frecursive` flags. In view of stuff like this http://icl.cs.utk.edu/lapack-forum/viewtopic.php?t=1930, are you sure this it won't result to LAPACK builds losing threadsafety?

zerothi:
Yes, however in http://icl.cs.utk.edu/lapack-forum/viewtopic.php?t=1930#p5457 Julien explains that

zerothi:
I see now that the recursive flag is for exactly that bug. It should be re-added.

zerothi:
@pv I have made a PR to fix this, thanks for noticing.

PR #181, opened 2017-08-30
Fix for ?GEBRD: WS should be declared as INTEGER

echeresh:


langou:
All good with me. I am waiting for the checks to complete to merge the pull request. 
I wonder whether WS was set as large as ( M+N )*NB. So not that big. 
I think the pull request is all fine.

echeresh:
@langou You are right, it is not related to the possible overflow of WS. But conversion INTEGER -> REAL in general is not safe because of possible rounding errors (if a value is big enough: > ~16M). So we have a special auxiliary function to provide safe ceil rounding (it is called for all cases where we want to convert INTEGER to REAL, just in case) and it expects INTEGER as an input, but here WS was declared as REAL and we observed some strange behavior.

langou:
Hi @echeresh, thanks for following up and explaining. Cheers, Julien.

PR #186, opened 2017-10-10
fixed path in installed package config files for CBLAS and LAPACKE.

raovgarimella:
These did not get caught in the cleanup of installation directory names (#e64606d). Fixes #21


julielangou:
hey it looks like the Travis build (automatic build) is failing with this update....

raovgarimella:
@julielangou - It seems that the master is also failing at the same place so maybe it is not because of this update?

raovgarimella:
Yay! Thank you

PR #188, opened 2017-10-12
Reinserted frecursive due to prior bug-reports

zerothi:
See 3a87b2df45b999b6da5a4ee3f54c766c7b1e8f4d for details.

Also, see comment by @pv in #177.


langou:
Super. Thanks. I should not have accepted the initial pull-request in the first place. Thanks for the second pull request. I am letting Travis CI do its job and then I will process the merge. Cheers, Julien.

zerothi:
No, I should have noticed. Sorry for the mistake.

zerothi:
@langou it fails on MacOSX, but the errors seem un-related?

julielangou:
ok, Yes indeed. Thanks

PR #189, opened 2017-10-16
Fixed (alot) missing external subroutines and removed "dead" variables

zerothi:
This commit is purely of cleaning up missing externals. Currently there
is no check on EXTERNAL's not used.

Basically there is no code changes other than declarations.


zerothi:
rebased to see if CI works now.

julielangou:
Thanks!

PR #190, opened 2017-10-20
Correct the argument order in the documentation for convf family

andreasnoack:
This probably happened because the `conv` and `convf` don't have `e` and `ipiv` in the same order.


julielangou:
Thank you! Nice catch

Issue #196, opened 2017-11-08
Infinite loop for infinity arguments in the ?larfg family of functions


martin-frbg:
Over at OpenBLAS, a recent issue https://github.com/xianyi/OpenBLAS/issues/1343 brought to our attention that DLARFG and related functions can enter an endless loop trying to rescale the input vector if its elements are "inf". 
While the fix would appear to be trivial - exiting the loop if the already present counter KNT exceeds some suitably large threshold here https://github.com/Reference-LAPACK/lapack/blob/6bcfa0d632158cb047a9b722da1a2f22c509ba12/SRC/dlarfg.f#L173-L174
I wonder if the circumstances would be seen as too contrived to warrant attention ?

langou:
Hi Martin. 

We fixed something related in ZLARTG in 2013:

> bug110 :: possible infinite loop in ZLARTG when input vector has NaN
> CORRECTED, rev 1437
> Present in LAPACK 3.4.2
> Bug report by Mark Hoemmen and Jack Poulson
> See http://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=5&t=3267&p=9067
> 
> Mark Hoemmen writes: Jack Poulson found that ZLARTG has an infinite loop related to NaN .EQ. 0. It’s the line below "ELSE IF( SCALE.LE.SAFMN2 )" between 10 and 20. When G is NaN, then the loop goes forever.
>
> Corrected by Rodney on Oct 5 2013

You wrote: 

> I wonder if the circumstances would be seen as too contrived to warrant attention ?

if ( ( (nan in input) => (infinite loop) ) .or. ( (inf in input) => (infinite loop) ) ) then we would like to be safe against this. This is not too contrived and this warrants attention. Thanks to OpenBLAS for pointing this.

martin-frbg:
Thanks for the quick reply. If you agree with the non-elegant solution of exiting the loop after something like 100 futile iterations I will prepare a PR soon. (I am not aware of a check for inf that is both portable and reliable ?)

weslleyspereira:
I think this issue was solved with #197. Am I right?

martin-frbg:
I think so too - and probably #308 as well

langou:
Thanks @martin-frbg and @weslleyspereira. Let us close this issue.

PR #197, opened 2017-11-10
Break out of potentially infinite rescaling loop after 1000 iterations

martin-frbg:
inf values in the input vector will survive rescaling, making the code hang. The limit of 1000 iterations is arbitrarily chosen with the intention to never interfere with regular behaviour. Fixes issue #196

### DIFF
--- a/SRC/clarfg.f
+++ b/SRC/clarfg.f
@@ -175,7 +175,7 @@
             BETA = BETA*RSAFMN
             ALPHI = ALPHI*RSAFMN
             ALPHR = ALPHR*RSAFMN
-            IF( ABS( BETA ).LT.SAFMIN )
+            IF( ABS( BETA ).LT.SAFMIN .AND. KNT .LT. 1000)
      $         GO TO 10
 *
 *           New BETA is at most 1, at least SAFMIN
--- a/SRC/clarfg.f
+++ b/SRC/clarfg.f
@@ -175,7 +175,7 @@
             BETA = BETA*RSAFMN
             ALPHI = ALPHI*RSAFMN
             ALPHR = ALPHR*RSAFMN
-            IF( ABS( BETA ).LT.SAFMIN .AND. KNT .LT. 1000)
+            IF( (ABS( BETA ).LT.SAFMIN) .AND. (KNT .LT. 20) )
      $         GO TO 10
 *
 *           New BETA is at most 1, at least SAFMIN
--- a/SRC/clarfgp.f
+++ b/SRC/clarfgp.f
@@ -197,7 +197,7 @@
             BETA = BETA*BIGNUM
             ALPHI = ALPHI*BIGNUM
             ALPHR = ALPHR*BIGNUM
-            IF( ABS( BETA ).LT.SMLNUM )
+            IF( ABS( BETA ).LT.SMLNUM .AND. KNT .LT. 1000 )
      $         GO TO 10
 *
 *           New BETA is at most 1, at least SMLNUM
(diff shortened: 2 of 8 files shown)


langou:
Thanks Martin. I will review some more. 

Another way to go is to check for infinity in BETA. I wonder if we cannot use the intrinsic ieee_is_finite(). I wonder how widely the intrinsic ieee_is_finite() is supported by compilers. This is part of Fortran 2003 standard. We do not use this intrinsic in LAPACK yet. To check for infinity, we could also check whether ABS( BETA ) is larger than the largest finite number or not.

Three questions in my mind. (1) What behavior do we want for LARFG if BETA is INF? (2) Does the code with (KNT .LT. 1000) behaves as the behavior in (1)? ( I think so. ) (3) Do we want 1000? Sounds like 20 is enough.

For sure if BETA is INF, we have an infinite loop and your fix removes this infinite loop. So this is great.

So, thanks a lot for the pull request. 

I will also try to update and change 
+            IF( ABS( BETA ).LT.SAFMIN .AND. KNT .LT. 1000)
with
+            IF( ( ABS( BETA ).LT.SAFMIN ) .AND. ( KNT.LT.1000) )
(so with parenthesis, this is more LAPACK notation like, really no big deal.)
I will need to learn how to edit a pull request. I'll see for this. 

Great. Thanks. 
Julien.

martin-frbg:
I can update the pull request with added parentheses if you like (I am not sure if I would know how to edit someone else's pull request either), but feel free to ignore this altogether if you have a more elegant solution (or rather, an elegant solution at all - I am quite aware that mine is just a dirty hack to make things work). 
I had a quick look around the internet and I think the issue with the ieee intrinsic was that some older version of gfortran did not support it. Not sure if this would be a concern here (and I do not recall the version number anyway, probably something relatively ancient like 4.3).

langou:
Let us accept this pull request for the coming release and revisit the issue later. 
I am merging the PR and creating an issue.

martin-frbg:
Thanks for merging this despite its obvious shortcomings. Issue #204 is sort-of a duplicate of my #196 now I think ?

langou:
Indeed. I thought the merge of your fix would close the issue, and so I opened a new issue. Indeed there is no point in this. Thanks for pointing this out. Closing #204.

vladimir-ch:
Sorry for commenting on a merged PR but I've been wondering whether this change actually does anything if `X` contains infinite values.

If `X` contains infinities, then `XNORM` will be infinite, `DLAPY2( ALPHA, XNORM )` will be infinite, and so `BETA` will be infinite. Finally, `ABS( BETA ).LT.SAFMIN` will be false and the branch with the loop will not be executed. Am I missing anything?

vladimir-ch:
On the other hand, if `X` contains infinities, `TAU` will be NaN, which is what may have been causing the issue in the original OpenBLAS ticket.

martin-frbg:
This issue was meant to be revisited, which is why #196 was kept open. My fix was certainly not elegant and probably not all-encompassing, but I was not sure how to (reliably, portably) check for INF and "the loop was there", so limiting its counter was trivial. Checking TAU for NaN looks like the clever thing to do, but at least currently it is calculated only after the problematic loop so would not help. (Whatever ABS(BETA) was at that point, it did appear to be .LT.SAFMIN in practice, though maybe it was actually NaN and the comparison was undefined)

vladimir-ch:
Sorry, I should have better commented on #196. I will take the discussion there.

PR #212, opened 2017-12-01
Fix misc. typos

luzpaz:
Mostly trivial source code comments.


langou:
@luzpaz : thanks!

musvaage:
@luzpaz 

Would you have an opinion on this extract?

in the `directoryso` of the

```
$ ed -s lapack/CMAKE/Findcodecov.cmake <<<'119,122p'
  # If expression was found, SOURCEFILE is a generator-expression for an
  # object library. Currently we found no way to call this function automatic
  # for the referenced target, so it must be called in the directoryso of the
  # object library definition.
$ 
```

musvaage:
I've been preparing a PR and will add this fix.

I'll assume "directory" is the intention in the extract.



in the `directory` of the

PR #214, opened 2017-12-16
Missing LAPACKE source files

turboencabulator:
I've found that the LAPACKE Makefile is missing a few source files.


martin-frbg:
This should already be covered by my PR #208 (which now also adds a fix for "ar" aborting on macOS)

langou:
I have just merged #208. Sorry for the delay. 
@turboencabulator can you follow up with @martin-frbg comment?

turboencabulator:
I've just merged the two, looks like @martin-frbg found all the same missing files that I did.  All that is left in my branch is some sorting to make it easier to compare the Makefile to CMakeLists.txt.

PR #218, opened 2017-12-23
Improvements for distribution packaging

besser82:
It comes in handy being able to suffix some files, if they are build with a certain flavour, like FORTRAN 64-bit int.  tmglib should be versioned as well.


cdluminate:
Any updates?

langou:
Thanks for the PR. We'll close this PR, the package has evolved a lot since then. There are discussions about packaging at #556.

PR #221, opened 2017-12-28
Little description mistake.

TarcioV:


langou:
> When SIDE = 'L', the transformation takes the form A := P*A and when SIDE = 'R', the transformation takes the form A := A*P**T

Weird. Yes. You are correct. No idea why there is this typo in the description of the array A. Thanks!

PR #225, opened 2018-01-24
Fixed #182 for N < 1

### LINKED ISSUE #182: Possible undefined output in delaruv
Am I missing something or could `ISEED` be garbage on output if `dlaruv` is called with `N` < 1?

https://github.com/Reference-LAPACK/lapack/blob/afd02ae8d40eefe454e004f506b878a4e48b517b/SRC/dlaruv.f#L438-L441

zerothi:
Fixes a possible call to laruv routines when N < 1.

See #182.

### DIFF
--- a/SRC/dlaruv.f
+++ b/SRC/dlaruv.f
@@ -386,6 +386,12 @@
 *     ..
 *     .. Executable Statements ..
 *
+
+*     Quick return for N < 1
+      IF ( N < 1 ) THEN
+         RETURN
+      END IF
+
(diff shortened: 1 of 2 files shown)


langou:
Hi Nick. Thanks for looking into this! No need for a commit, I guess then. Cheers, Julien.

martin-frbg:
Anybody recall what happened here ? I see that a fix was prepared but closed rather than merged, but I see nothing in this conversation (or the original issue) that explains the "no need for a commit, I guess"

PR #234, opened 2018-02-23
Update dbdsqr.f

TarcioV:
work needs at most 4 elements (2 cosines and 2 sines) per entries in the sub-(super)-diagonal, that is, N-1 entries.


langou:
Thanks @TarcioV. Julien.

PR #236, opened 2018-02-24
Correcting doxygen warning message

albert-github:
Correcting message: "BLAS/SRC/sdsdot.f:165: warning: Member sdsdot found in multiple @ingroup groups! The member will be put in group single_blas_level1, and not in group complex_blas_level1" by removing incorrect ingroup command.


langou:
Thanks.

Issue #211, opened 2017-11-30
nanchecks in lapacke larfb assume row-major


mgates3:
The nanchecks in lapacke larfb routines assume the V matrix is stored row-major.

For instance, this accesses the V(k,0) as v[k*ldv].
```
         if( LAPACKE_lsame( storev, 'c' ) && LAPACKE_lsame( direct, 'f' ) ) {
             if( LAPACKE_str_nancheck( matrix_layout, 'l', 'u', k, v, ldv ) )
                 return -9;
             if( LAPACKE_sge_nancheck( matrix_layout, nrows_v-k, ncols_v, &v[k*ldv], ldv ) )
```
It should be v[k*lrv] where lrv is the row stride, either lrv=1 for col-major or lrv=ldv for row-major.
```
             if( LAPACKE_sge_nancheck( matrix_layout, nrows_v-k, ncols_v, &v[k*lrv], ldv ) )
```

Additionally, the last two cases have the same if condition. The last one should be direct=b instead of direct=f.
```
         } else if( LAPACKE_lsame( storev, 'r' ) && LAPACKE_lsame( direct, 'f' ) ) {
             ...
         } else if( LAPACKE_lsame( storev, 'r' ) && LAPACKE_lsame( direct, 'f' ) ) {
```
I'll submit a fix.

Issue #210, opened 2017-11-30
wrong vector size in lapacke lacgv, lassq, larfg nancheck


mgates3:
In the routines below, the vector length is wrong. The arrays have those dimensions (except lassq), but the logical vector length is either n (for lacgv and lassq) or n-1 (for larfg). If abs(incx) > 1, these checks access the array x out-of-bounds.

```
LAPACKE/SRC> grep nancheck *.c | grep inc
lapacke_clacgv.c:        if( LAPACKE_c_nancheck( 1+(n-1)*ABS(incx), x, incx ) ) {
lapacke_clarfg.c:        if( LAPACKE_c_nancheck( 1+(n-2)*ABS(incx), x, incx ) ) {
lapacke_classq.c:        if( LAPACKE_c_nancheck( 1+(n-2)*ABS(incx), x, incx ) ) {
lapacke_dlarfg.c:        if( LAPACKE_d_nancheck( 1+(n-2)*ABS(incx), x, incx ) ) {
lapacke_dlassq.c:        if( LAPACKE_d_nancheck( 1+(n-2)*ABS(incx), x, incx ) ) {
lapacke_slarfg.c:        if( LAPACKE_s_nancheck( 1+(n-2)*ABS(incx), x, incx ) ) {
lapacke_slassq.c:        if( LAPACKE_s_nancheck( 1+(n-2)*ABS(incx), x, incx ) ) {
lapacke_zlacgv.c:        if( LAPACKE_z_nancheck( 1+(n-1)*ABS(incx), x, incx ) ) {
lapacke_zlarfg.c:        if( LAPACKE_z_nancheck( 1+(n-2)*ABS(incx), x, incx ) ) {
lapacke_zlassq.c:        if( LAPACKE_z_nancheck( 1+(n-2)*ABS(incx), x, incx ) ) {
```

Additionally, lapack/SRC/*lassq.f documents the wrong dimension for X. It should be dimension (1+(N-1)*INCX), instead of dimension (N). It doesn't allow incx < 0.

```
*> \param[in] X
*> \verbatim
*>          X is COMPLEX*16 array, dimension (N)
*>          The vector x as described above.
*>             x( i )  = X( 1 + ( i - 1 )*INCX ), 1 <= i <= n.
*> \endverbatim
```

mgates3:
I'll submit a fix.

PR #249, opened 2018-05-07
Lapacke fixes

mgates3:
Fix several issues:
1) Issue #210 (lapacke_*lacgv, lassq, larfg)
2) Issue #211 (lapacke_*larfb)
3) lapacke_*larfx could fail nancheck or segfault for side = right and m > n, since it accesses V out-of-bounds.
4) lapack_*mtr could fail nancheck or segfault for side = right and m > n, since it accesses tau out-of-bounds.


langou:
Awesome. Thanks Mark!

PR #251, opened 2018-05-17
Add -b (bin) option to lapack_testing.py

malcolmroberts:
When building with cmake, the test executables and data files may not
be in the location where lapack_testing.py.  By adding the -b option,
we can specify this manually.  For example, if we do
  mkdir build && cd build && cmake -DBUILD_TESTING=ON .. && make
we can then call
  ./lapack_testing.py -r -b bin -d ../TESTING
which now runs the tests.

### DIFF
--- a/lapack_testing.py
+++ b/lapack_testing.py
@@ -12,8 +12,8 @@ import os, sys, math
 import getopt
 # Arguments
 try:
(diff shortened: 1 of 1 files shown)


langou:
@malcolmroberts: Thanks!

PR #252, opened 2018-05-29
Doc fixes

drwells:
I used the documentation to generate some code recently and made a few small fixes.


langou:
Sounds good. Thanks for the comments. I am always feeling awkward to label the work array argument as being ``output`` since it really does (in general and in these cases) not contain anything of value in output; but for sure, labelling it as ``output`` is better than as it was labelled before as ``input``. The other fixes are great as well. Thanks. J.

PR #253, opened 2018-05-31
Fixes for Aasen's LAPACKE interfaces & workspace sizes, + some other small fixes

kbergstr:
- LAPACKE interfaces for Aasen's functions now call ?sytrf_aa and ?hetrf_aa instead of ?sytrf and ?hetrf
- When minimal workspace is given in ?hesv_aa, ?sysv_aa, ?hesv_aa_2stage, ?sysv_aa_2stage, now no error is given
- Quick return for ?laqr1
- Some other small fixes to documentation, testing, etc.


langou:
Thanks Kirana!

PR #256, opened 2018-06-18
sgeqrf: documentation of WORK without MAX(1,...)

amigalemming:
Documentation of WORK: removed MAX(1,...) because LWORK>=1 is required in the description of LWORK anyway.
MAX(1,...) is still correct but confusing.


zerothi:
`LWORK` may be -1 for a work size query.

Issue #262, opened 2018-07-02
DSYEVR_2stage illegal value at parameter 7


uihsnv:
When I use the 2stage routine of dsyevr, I get the following error message:

` ** On entry to DGEQRF parameter number  7 had an illegal value`

The same usage works well with Intel's MKL.

EDIT: I forgot to mention that I'm using it via its `LAPACKE_dsyevr_2stage()` interface.

EDIT2: I'm calling it with the following parameters:

```
MATRIX_LAYOUT = LAPACK_COL_MAJOR
JOBZ = 'N'
RANGE = 'I'
UPLO = 'L'
N = 150
A = <pointer to a 150x150 sized calloc'd double>
LDA = 150
**VL = 0.0**  -> This is parameter number 7; Why is it "illegal"?
VU = 0.0
IL = 4
IU = 4
ABSTOL = 1e-12
M = <pointer to a malloc'ed int>
W = <pointer to a 150 sized malloc'd double>
Z = <pointer to a malloc'd double>
LDZ = 150
ISUPPZ = <pointer to a malloc'ed int>
```

zerothi:
Not that I have a solution but it is the routine `DGEQRF` that gets the wrong argument (not `dsyevr_2stage`). This suggests an erroneous work-array allocation that is passed through the call-tree.

langou:
Hi all, Azzam is having a look at this right now. (Thanks Azzam.) He identified that the problem comes from https://github.com/Reference-LAPACK/lapack/commit/7b651690c7bc24878a1ee92d916ea7e0c9d825fe

zerothi:
I have fixed the bug (it was due to missing conversions in the old #177).

This should clear out the remaining calls. I have tested this small snippet which fails on 3.8.0 but succeeds with #271.
```fortran
program test
  implicit none

  integer, parameter :: N = 150, M = 1
  real(8) :: A(N,N), W(N), Z(N,M)
  real(8) :: ISUPPZ(2*M)
  real(8), allocatable :: WORK(:)
  integer, allocatable :: IWORK(:)
  integer :: LWORK, LIWORK
  integer :: mm, IL, IU
  integer :: i, j

  IL = 4
  IU = IL + M - 1

  A = 0
  do i = 1, N
    A(i,i) = 1
  end do

  allocate(WORK(1), IWORK(1))
  LWORK = -1
  LIWORK = -1
  call dsyevr_2stage('N','I','L',N,A,N,0._8, 0._8, &
[...]

PR #270, opened 2018-08-12
 fixing LAPACK issues #262 by rolling back to use ILAENV instead of ILAENV2STAGE

haidarazzam:


zerothi:
#271 should superseede this one.

PR #271, opened 2018-08-12
Fixed #262 by fixing #177

### LINKED ISSUE #262: DSYEVR_2stage illegal value at parameter 7

zerothi:
There were some missing `ILAENV` -> `ILAENV2STAGE` conversions
missing.
This commit fixes all calls to `ILAENV` that uses the parameters
17, 18, 19, 20 (the old `ILAENV`).

I have also tested this for the issue reported in #262.

### DIFF
--- a/SRC/chetrd_hb2st.F
+++ b/SRC/chetrd_hb2st.F
@@ -280,8 +280,8 @@
 *     ..
 *     .. External Functions ..
       LOGICAL            LSAME
-      INTEGER            ILAENV 
-      EXTERNAL           LSAME, ILAENV
+      INTEGER            ILAENV2STAGE 
+      EXTERNAL           LSAME, ILAENV2STAGE
 *     ..
 *     .. Executable Statements ..
 *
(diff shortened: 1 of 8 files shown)


haidarazzam:
@zerothi please check the latest commit, I had 2 commit, the first one was missing some files the next commit was adding all the files. Let me know if this is OK for you.

zerothi:
@haidarazzam, all I can say is that I think having recursive as a prerequisite for the subroutine is not really necessary. This was my main motivation for creating the initial PR.

I.e. I would be in favour of retaining the ILAENV2STAGE method.

haidarazzam:
@zerothi , why are you saying that the way it was it requires recursive? 
I am saying it similar to HSEQR that use ISPEC 11:15 and call IPARAMQ.
I followed the same model.
I thought that recursive was added to accommodate the recursive panel of LU.
Please let me know why frecursive will be needed n case of 1 ILAENV.
BTW you can contact me directly though haidar@icl.utk.edu
Thanks
Azzam

zerothi:
@haidarazzam, because the old `ILAENV` called `IPARAM2STAGE` which in turn called `ILAENV`.

victorliu:
I reported errata bug 61 (*), in which some large local arrays ended up being allocated on the heap using static buffers, resulting in multi-threaded calls to some calls (like ZGEEV, which eventually calls ZGEHRD) would thrash each others workspaces. The flag -frecursive is a mitigation so that these large arrays would be allocated on the stack, as a temporary fix at least for GCC. I am partial to leaving -frecursive in the Makefile, as I happen to use Lapack in multithreaded scenarios, and package maintainers apparently use the default flags without checking for thread safety.


(*) http://www.netlib.org/lapack/Errata/errata_3.4.0.html#_strong_span_class_red_bug0061_span_strong_zgehrd_f_is_overflowing

zerothi:
@victorliu please note that `-frecursive` hasn't been removed (and won't).

This is merely discussing the recursive calls that was originally in the `ILAENV` routine.

langou:
I am closing this PR. There will be no merge and no follow up. @zerothi and @haidarazzam worked in parallel and essentially provided the same PR. #271 and #272. Both are essentially the same PR. We will merge #272 and close #271. That said, thanks @zerothi. Cheers, Julien

PR #277, opened 2018-08-15
cmake: Fix LAPACKE include directory associated with the build interface

jcfr:
This commit ensures that the <root>/LAPACKE/include directory is associated
with the LAPACKE target and it fixes configuration error when building project
against a build tree of the LAPACK project.

Co-authored-by: Pablo Hernandez <pablo.hernandez@kitware.com>


jcfr:
cc: @langou

PR #280, opened 2018-09-07
Fixed #279 by adding IF clauses

### LINKED ISSUE #279: swap VS only if JOBVS=V in s/dgeesx
In S/DGEES, there is a check before swap VS. For example, SGEES, line 473-475.

But in S/DGEESX, it always do swap VS (line 586). But if JOBVS = N, VS should not be referenced.

zerothi:
Added IF clauses


langou:
Thanks Nick. This commit looks good to me. Julien.

Issue #261, opened 2018-06-27
Accuracy issue in norms (lange, lansy, lantr, etc.)


mgates3:
There's an accuracy issue in the LAPACK and ScaLAPACK Frobenius norms, which exhibits at large sizes, particularly in single precision. Ignoring the scaling for the moment, it computes a sum:

    sum = 0;
    for j = 1:n
        for i = 1:m
            sum = sum + abs(A(i,j))^2;
        end
    end

For large matrices, clearly sum >> abs(A(i,j))^2, making the addition inaccurate. A better, though not perfect, solution would be to sum each column individually, and accumulate the column sums:

    sum = 0;
    for j = 1:n
        colsum = 0;
        for i = 1:m
            colsum = colsum + abs(A(i,j))^2;
        end
        sum = sum + colsum;
    end

For extremely tall matrices (say, 1e6 rows), there would still be an issue. I've implemented this fix in ScaLAPACK *lange, *lansy, *lantr, and can easily do the same changes for LAPACK. It adds only a few multiply-adds and a divide per column to deal with the scaling, so has trivial performance impact.

We noticed this issue in developing SLATE. Because SLATE computes results per tile and then sums up tile results, its answer is more accurate than ScaLAPACK's. LAPACK computes the same answer as single node ScaLAPACK (except Intel MKL appears to have fixed the issue already). Below is output for genorm (i.e., lange) before and after changes. The relative error is computed as
    (||A||_lapack - ||A||_slate) / (sqrt(m n) ||A||_lapack),
with a tolerance of error < 3*eps to pass. Note that the error grows with the size of the matrix. Also, the same matrix is used in single and double, so the norms should match.

    [note: output modified for clarity and brevity]
    slate/test> ./test genorm --norm fro --dim 500:4000:500,10000 --verbose 1 --type s,d

                                            SLATE       ScaLAPACK  relative
    type     norm       m       n            norm            norm     error  status
    single    fro     500     500  2.88898743e+02  2.88892822e+02  4.10e-08  pass
    single    fro    1000    1000  5.77679993e+02  5.77577881e+02  1.77e-07  pass
    single    fro    1500    1500  8.66120239e+02  8.65607910e+02  3.95e-07  FAILED
    single    fro    2000    2000  1.15464795e+03  1.15303625e+03  6.99e-07  FAILED
    single    fro    2500    2500  1.44320764e+03  1.44007178e+03  8.71e-07  FAILED
[...]

PR #290, opened 2018-11-06
Update Frobenius norms for better accuracy.

mgates3:
This fixes issue #261.

I added `scombssq` and `dcombssq` to merge two sum-of-squares quantities. These functions are taken from ScaLAPACK (TOOLS/pdtreecomb.f), just the documentation is updated to match LAPACK's style.

Updated norms were tested via the LAPACK++ tester, comparing single to double precision on identical matrices, and results from new code with results from old code.

For instance, original results (output abbreviated):
```
lapackpp/test> ./test lange --norm fro --dim 4000 --verbose 1 --type s
                                            LAPACK++        Ref.
type     norm       m       n      error    time (s)    time (s)  status
norm_tst = 3.0072395019531250e+03
   s      fro    4000    4000   0.00e+00      0.0443      0.0557  pass

lapackpp/test> ./test lange --norm fro --dim 4000 --verbose 1 --type d
norm_tst = 3.0541497549406699e+03
   d      fro    4000    4000   0.00e+00      0.0694      0.0810  pass
```
Using double precision (type d) as reference, this gives a relative error for single precision (type s) of 1.5360e-02.

With updated code:
```
lapackpp/test> ./test lange --norm fro --dim 4000 --verbose 1 --type s
                                            LAPACK++        Ref.
type     norm       m       n      error    time (s)    time (s)  status
norm_tst = 3.0541506347656250e+03
   s      fro    4000    4000   0.00e+00      0.0449      0.0524  pass

lapackpp/test> ./test lange --norm fro --dim 4000 --verbose 1 --type d
norm_tst = 3.0541497549404239e+03
   d      fro    4000    4000   0.00e+00      0.0688      0.0808  pass
```
Again using double-precision as reference, this gives a much better relative error for single precision of 2.8808e-07. Note also that the double-precision result is close to that generated by the old code.

For reference, in LAPACK++ test I ran:
```
python ./run_tests.py --aux-norm --norm fro --verbose 1 --medium --large --type s > snorm.txt
python ./run_tests.py --aux-norm --norm fro --verbose 1 --medium --large --type d > dnorm.txt

python ./run_tests.py --aux-norm --norm fro --verbose 1 --medium --large --type c > cnorm.txt
python ./run_tests.py --aux-norm --norm fro --verbose 1 --medium --large --type z > znorm.txt
```
and compared those files. I modified the LAPACK++ matrix_generator.cc to generate *exactly* the same matrix for single and double precision, e.g., for all i, j:
```
    A( i, j ) = rand() / float(RAND_MAX);
```
instead of using `larnv`.

I did not address RFP matrix norms (`dlansf`, ...). Nor tridiagonal matrices (`dlangt`, ...), which due to having fewer entries (only O(n) instead of O(n^2)), are less affected by the accuracy issue.

### DIFF
--- a/SRC/CMakeLists.txt
+++ b/SRC/CMakeLists.txt
@@ -150,7 +150,8 @@ set(SLASRC
    stplqt.f stplqt2.f stpmlqt.f
    ssytrd_2stage.f ssytrd_sy2sb.f ssytrd_sb2st.F ssb2st_kernels.f
    ssyevd_2stage.f ssyev_2stage.f ssyevx_2stage.f ssyevr_2stage.f
-   ssbev_2stage.f ssbevx_2stage.f ssbevd_2stage.f ssygv_2stage.f)
+   ssbev_2stage.f ssbevx_2stage.f ssbevd_2stage.f ssygv_2stage.f
+   scombssq.f)
 
 set(DSLASRC spotrs.f sgetrs.f spotrf.f sgetrf.f)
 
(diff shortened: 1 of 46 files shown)


julielangou:
Cool! Will close #261 
We may need some dedicated testing at some point... but seems the unit testing went fine.
Thanks @mgates3

PR #291, opened 2018-11-07
Lapacke fixes

mgates3:
Fixes a few minor issues (const, unused params) in LAPACKE and minor documentation issues. These were causing errors when using LAPACK++ lapack.h header, which has correct const.


langou:
Thanks. Nice clean up. Thanks Mark! Julien.

PR #293, opened 2018-11-07
gesvdq

mgates3:
Adds liwork and workspace queries for iwork, rwork.


julielangou:
Looks good! Thanks @mgates3

PR #295, opened 2018-11-09
BLAS++ and LAPACK++ info

julielangou:
Just to let know our users about those two projects and their possibly future external integration into the LAPACK library

@jcfr and @mgates3 is that simple enough?

### DIFF
--- a/SRC/CMakeLists.txt
+++ b/SRC/CMakeLists.txt
@@ -150,7 +150,8 @@ set(SLASRC
    stplqt.f stplqt2.f stpmlqt.f
    ssytrd_2stage.f ssytrd_sy2sb.f ssytrd_sb2st.F ssb2st_kernels.f
    ssyevd_2stage.f ssyev_2stage.f ssyevx_2stage.f ssyevr_2stage.f
-   ssbev_2stage.f ssbevx_2stage.f ssbevd_2stage.f ssygv_2stage.f)
+   ssbev_2stage.f ssbevx_2stage.f ssbevd_2stage.f ssygv_2stage.f
+   scombssq.f)
 
 set(DSLASRC spotrs.f sgetrs.f spotrf.f sgetrf.f)
 
@@ -341,7 +342,8 @@ set(DLASRC
    dtplqt.f dtplqt2.f dtpmlqt.f
    dsytrd_2stage.f dsytrd_sy2sb.f dsytrd_sb2st.F dsb2st_kernels.f
    dsyevd_2stage.f dsyev_2stage.f dsyevx_2stage.f dsyevr_2stage.f
-   dsbev_2stage.f dsbevx_2stage.f dsbevd_2stage.f dsygv_2stage.f)
+   dsbev_2stage.f dsbevx_2stage.f dsbevd_2stage.f dsygv_2stage.f
+   dcombssq.f)
 
 set(DXLASRC dgesvxx.f dgerfsx.f dla_gerfsx_extended.f dla_geamv.f
    dla_gercond.f dla_gerpvgrw.f dsysvxx.f dsyrfsx.f
--- a/SRC/Makefile
+++ b/SRC/Makefile
@@ -174,7 +174,7 @@ SLASRC = \
    ssytrd_2stage.o ssytrd_sy2sb.o ssytrd_sb2st.o ssb2st_kernels.o \
    ssyevd_2stage.o ssyev_2stage.o ssyevx_2stage.o ssyevr_2stage.o \
    ssbev_2stage.o ssbevx_2stage.o ssbevd_2stage.o ssygv_2stage.o \
-   sgesvdq.o
+   sgesvdq.o scombssq.o
 
 DSLASRC = spotrs.o sgetrs.o spotrf.o sgetrf.o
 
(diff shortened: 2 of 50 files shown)


mgates3:
This seems good for the moment — gives a pointer, without raising expectations too high.

CMakeLists.txt
```diff
@@ -315,6 +315,31 @@ if(LAPACKE)
   add_subdirectory(LAPACKE)
 endif()
 
+# --------------------------------------------------
+# BLAS++ / LAPACK++
+option(BLAS++ "Build BLAS++" OFF)
+option(LAPACK++ "Build LAPACK++" OFF)
```
jcfr: To be consistent with the name of the projects, I suggest to use `BLASPP` and `LAPACKPP`

* https://bitbucket.org/icl/blaspp
* https://bitbucket.org/icl/lapackpp

CMakeLists.txt
```diff
@@ -315,6 +315,31 @@ if(LAPACKE)
...
+  message(STATUS "For support BLAS++ related question, please email: slate-user@icl.utk.edu")
+  message(STATUS "----------------")
+endif()
+
+if (LAPACK++ )
+  message(STATUS "LAPACK++ enable")
+  message(STATUS "----------------")
+  message(STATUS "Thank you for your interest in LAPACK++ , a newly developed C++ API for LAPACK library")
+  message(STATUS "The objective of LAPACK++ is to provide a convenient, performance oriented API for development in the C++ language, that, for the most part, preserves established conventions, while, at the same time, takes advantages of modern C++ features, such as: namespaces, templates, exceptions, etc.")
+  message(STATUS "We are still working on integrating LAPACK++ in our library. For the moment, you can download directly blas++ from https://bitbucket.org/icl/lapackpp")
```
jcfr: `blas++` -> `lapack++`

CMakeLists.txt
```diff
@@ -315,6 +315,31 @@ if(LAPACKE)
...
+  message(STATUS "For support LAPACK++ related question, please email: slate-user@icl.utk.edu")
+  message(STATUS "----------------")
+endif()
```
jcfr: To ensure consistency in the message, I instead suggest something like this:


```
function(_display_cpp_implementation_msg name)
  string(TOLOWER ${name} name_lc)
  message(STATUS "${name}++ enable")
  message(STATUS "----------------")
  message(STATUS "Thank you for your interest in ${name}++, a newly developed C++ API for ${name} library")
  message(STATUS "The objective of ${name}++ is to provide a convenient, performance oriented API for development in the C++ language, that, for the most part, preserves established conventions, while, at the same time, takes advantages of modern C++ features, such as: namespaces, templates, exceptions, etc.")
  message(STATUS "We are still working on integrating ${name}++ in our library. For the moment, you can download directly ${name_lc}++ from https://bitbucket.org/icl/${name_lc}pp")
  message(STATUS "For support ${name}++ related question, please email: slate-user@icl.utk.edu")
  message(STATUS "----------------")
endif()
if(BLASPP)
  _display_cpp_implementation_msg("BLAS")
endif()
if(LAPACKPP)
  _display_cpp_implementation_msg("LAPACK")
endif()

```

Makefile
```diff
@@ -23,6 +23,18 @@ lapacklib:
...
+	@echo "The objective of BLAS++ is to provide a convenient, performance oriented API for development in the C++ language, that, for the most part, preserves established conventions, while, at the same time, takes advantages of modern C++ features, such as: namespaces, templates, exceptions, etc."
+	@echo "We are still working on integrating BLAS++ in our library. For the moment, you can download directly blas++ from https://bitbucket.org/icl/blaspp"
+	@echo "For support BLAS++ related question, please email: slate-user@icl.utk.edu"
+
+lapackpplib:
+	@echo "Thank you for your interest in LAPACK++, a newly developed C++ API for LAPACK library"
+	@echo "The objective of LAPACK++ is to provide a convenient, performance oriented API for development in the C++ language, that, for the most part, preserves established conventions, while, at the same time, takes advantages of modern C++ features, such as: namespaces, templates, exceptions, etc."
+	@echo "We are still working on integrating LAPACK++ in our library. For the moment, you can download directly blas++ from https://bitbucket.org/icl/lapackpp"
```
jcfr: `ou can download directly blas++` -> `ou can download directly lapack++`

jcfr:
It is also strange that there is a commit like `Merge remote-tracking branch 'upstream/master'` in this branch, I suggest you rebase your changes on top of an update to date master.

CMakeLists.txt
mgates3: The projects are named BLAS++ and LAPACK++. The "blaspp" and "lapackpp" names are just because + doesn't play well as a filename. Bitbucket doesn't seem to allow + in a repo name.

jcfr: I still think the option should be consistent with the name of the cmake project and associated targets, see https://bitbucket.org/icl/blaspp/src/da11db2a6f85b402591005b05f233bda19aa4afe/CMakeLists.txt?at=default&fileviewer=file-view-default#CMakeLists.txt-7

CMakeLists.txt
jcfr: @mgates3  As you can see in the text above, the actual project is still describe as "${name}++" where `name` could be either `LAPACK` or `BLAS`

julielangou:
will do a clean pull request

PR #297, opened 2018-11-10
BLAS++ and LAPACK++ info

julielangou:
Just to let know our users about those two projects and their possibly future external integration into the LAPACK library


julielangou:
redoing #295 with @jcfr suggestions

PR #310, opened 2019-01-13
Correct/clarify CI config, document CMake build, update obsolete CMake syntax

scivision:


langou:
Thanks @scivision for the pull request. Thanks @jcfr for the review. !!!!

CMakeLists.txt
```diff
@@ -72,18 +72,18 @@ include(PreventInSourceBuilds)
 include(PreventInBuildInstalls)
 
 if(UNIX)
-  if("${CMAKE_Fortran_COMPILER}" MATCHES "ifort")
-    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fp-model strict")
+  if(CMAKE_Fortran_COMPILER_ID STREQUAL Intel)
```
jcfr: @langou For reference, here is the correspondence between compiler name and compiler id. See https://github.com/Kitware/CMake/blob/40628b2519ae1553a33f701cf46ee8fbbd013b85/Modules/CMakeDetermineFortranCompiler.cmake#L84-L93

scivision: Perhaps reference to the compiler vendor would be more abstract and generally applicable: https://cmake.org/cmake/help/v3.13/variable/CMAKE_LANG_COMPILER_ID.html

PR #319, opened 2019-01-30
For symmetric eigensolvers  only upper or lower triangle of a matrix …

sergey-v-kuznetsov:
…should be set

and  remaining  part of matrix can contain any values including NaNs but  these LAPACKE
symmetric eigensolver drivers checked the whole matrix for NaNs and they used
transposition  of the whole matrix. The new versions check and transpose upper or lower
triangle of matrix.


langou:
Thanks @sergey-v-kuznetsov !

PR #340, opened 2019-05-10
small fixes in doc of xlatmr

SimonDeRidder:
The `SPARSE` parameter was in the wrong position in the docs, and the setting of `KLL` (as it is written in the code), is maximally `M-1`, not `N-1`.


julielangou:
Thank you for your contribution.

PR #347, opened 2019-07-28
Makefile cleanup

turboencabulator:
Here are a few changes to make the makefiles a little more standardized/portable.

There are a few things still using GNU make syntax, but the files can now easily accommodate FreeBSD's make by doing a couple find & replaces:
* Replace $^ with $>
* Prefix all ifdef/else/endif keywords with a leading '.'


langou:
Hi @turboencabulator . Thanks. Looks great to me. Cheers, @langou .

PR #356, opened 2019-10-15
Changes to TESTING/LIN to make it compile with recent gfortran

tkoenig1:
After https://gcc.gnu.org/viewcvs/gcc?view=revision&revision=276972 , gfortran rejects passing a scalar and an array as the same actual argument to a procedure (except for the case of an array element, which is allowed).

This has led to several compilation errors in TESTING/LIN, which I have tried to fix with this pull request.

It is also possible to use -fallow-argument-mismatch to downgrade the error to a warning.


langou:
Thanks @tkoenig1 !

PR #363, opened 2019-11-15
Fixes CMake builds and documentation in [sz]laorhr_getrfnp2

zerothi:
A recent commit 8189dae5acf59ba4da35f06c4c2c5e88ee0f7147 forgot
to remove the deleted file from SRC/CMakeLists.txt.

NOTE: The commit referenced above changed alot in cmake builds. Are all these things tested? It seems the CI was broken by the changes.


zerothi:
This fix doesn't really fix the cmake build, sorry. It was just to show that a filename was missed deleted in the CMakeLists.txt.

The entire cmake system is broken since the referenced commit... :(

PR #370, opened 2019-11-25
Fix MinGW xgesvd build error

jschueller:
With MinGW the build fails for some LAPACKE xgesvd routines with:
lapacke_cgesvdq.c:74:5: error: aggregate value used where an integer was expected
Seems the C2INT macro is useful there.


langou:
Thanks @jschueller !

moe123:
@langou would be very nice to kill 3.9.0-release as is; waiting a bit until master reaches a stable state; for now, that should be re-qualified under an alpha label.

jschueller:
@moe123 they can still release a 3.9.1 some day

PR #378, opened 2019-12-20
LAPACK: avoid out-of-bound write in ?LANTR

vladimir-ch:
The out-of-bound write will happen when computing the inf-norm of wide, lower triangular matrices with "tight" work. It is documented that this case is not supported by ?LANTR but since the change is small and easy, it is worth doing it.

I think the documentation unfortunately cannot be updated to lift the restriction on matrix sizes.

### DIFF
--- a/SRC/clantr.f
+++ b/SRC/clantr.f
@@ -287,7 +287,7 @@
             END IF
          ELSE
             IF( LSAME( DIAG, 'U' ) ) THEN
-               DO 210 I = 1, N
+               DO 210 I = 1, MIN( M, N )
                   WORK( I ) = ONE
   210          CONTINUE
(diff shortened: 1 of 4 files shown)


langou:
Hi @vladimir-ch , ``I think the documentation unfortunately cannot be updated to lift the restriction on matrix sizes.`` => What do you mean?, (Thanks for the PR.) Cheers,  @langou

vladimir-ch:
I mean that until other lapack implementation providers include this change, the documentation would be misleading if someone wanted to use the reference lapack's documentation as the reference documentation. But maybe it's a pessimization on my side and I'd be happy if the restriction on matrix shape were removed.

PR #382, opened 2020-01-10
LAPACKE: don't allocate transposed matrix in ?lantr_work

vladimir-ch:
This PR modifies `LAPACKE_?lantr_work` functions to avoid allocation (and filling) of the transposed triangular matrix `a_t`. It uses the standard trick that a lower triangular matrix in row-major storage can be viewed as an upper triangular matrix in column-major (and vice versa) if the parameters are properly adjusted. In fact, this is the same approach that `LAPACKE_?lange_work` functions use. With this approach a new `work` array must be allocated because the rows and columns are "swapped" between row/column-major (and not because LAPACKE is taking into account that it would make sense to have `work` sized differently in row major as I originally misunderstood).

I feel a little uneasy because of the lack of any tests but at least for double precision our Gonum test passes.


langou (approved):
Excellent! Very elegant. Thanks!

weslleyspereira (approved):

PR #383, opened 2020-01-17
?gesvdq.f: fix typo in documentation

tenomoto:
The estimated length of `IWORK` seems to be given in `IWORK(1)` rather than `LIWORK(1)`.


langou:
Thanks @tenomoto for the fix and thanks @zerothi for the review. Cheers, Julien.

PR #386, opened 2020-01-29
doc: fixed documentation in [cz]dot[uc]

zerothi:
The documentation specified arguments as REAL arrays, however
they should be COMPLEX.


langou:
Thanks Nick. This was quite a bad typo. Thank!

Issue #336, opened 2019-04-23
CMake build with Intel compiler fails


corbett5:
When using the Intel compiler `CMAKE_Fortran_FLAGS` is being set incorrectly. The culprit is
https://github.com/Reference-LAPACK/lapack/blob/master/CMakeLists.txt#L75-L80
which yields `-fp-model;strict` instead of `-fp-model strict`. I believe these lines should be changed to
```cmake
if(CMAKE_Fortran_COMPILER_ID STREQUAL Intel)
    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fp-model strict")
endif()
if(CMAKE_Fortran_COMPILER_ID STREQUAL XL)
    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -qnosave -qstrict=none")
endif()
```

Note that this was the original implementation before https://github.com/Reference-LAPACK/lapack/commit/7c984113968c54fa52e081aa5c9a3f226b494c1b.

raovgarimella:
Yes, CMake sees ```"-fp-model strict"``` as two list items to be added to ```CMake_Fortran_FLAGS```

raovgarimella:
@julielangou This was a pretty serious bug in the CMake build system and makes it impossible to build 3.9.0 with the intel compiler without patching the CMakeLists.txt file. Will there be a patch release soon?

julielangou:
We will release in June.

PR #388, opened 2020-02-20
reverting incorrect changes made to syntax for CMAKE_Fortran_FLAGS with Intel/XL compilers

raovgarimella:
commits 7c98411 and ac26c7cc introduced an incorrect syntax for setting CMAKE_Fortran_FLAGS. This variable was treated as a list instead of a string and was generating incorrect syntax. Verified fix with a successful Intel 18 configuration on linux.

References issue #336


julielangou:
Thanks!

Issue #247, opened 2018-04-21
TESTING/dbal.in had extra semi-colon on second to last line causing I/O error


gjparker9:
$ tail -n 3 ./TESTNG/dbal.in
  2.494800386918399765D+291  1.582914569427869018D+175  1.004336277661868922D+59  3.186183822264904554D-58  5.053968264940243633D-175  0.40083367200179455560D-291;

0
$

the ';' caused ifort (18.0.1) with OPTS="-O3 -fp-model strict -assume protect_parens -axAVX,CORE-AVX2,CORE-AVX512,MIC-AVX512,COMMON-AVX512" otherwise INSTALL/make.inc.ifort to fail with an I/O error:

$ cat TESTING/dbal.out
forrtl: severe (59): list-directed I/O syntax error, unit 5, file /proc/25425/fd/0
Image              PC                Routine            Line        Source             
xeigtstd           0000000000E9FE6E  Unknown               Unknown  Unknown
xeigtstd           0000000000E77134  Unknown               Unknown  Unknown
xeigtstd           0000000000E754A2  Unknown               Unknown  Unknown
xeigtstd           000000000046A5A1  Unknown               Unknown  Unknown
xeigtstd           000000000042503C  Unknown               Unknown  Unknown
xeigtstd           0000000000402FDE  Unknown               Unknown  Unknown
libc-2.17.so       00007F71BB27AC05  __libc_start_main     Unknown  Unknown
xeigtstd           0000000000402EE9  Unknown               Unknown  Unknown

colinbann:
I've hit this error, using Intel 17.  I wonder if this is likely to be fixed ?

julielangou:
Fixed in #391

PR #391, opened 2020-02-23
remove extraneous semicolon

plunky:
this semicolon causes a test to fail with read error, using g95

this was already reported as issue #247


julielangou:
Thanks Iain

Issue #394, opened 2020-03-10
Intel "undefined reference to `xerbl_'" link errors in v3.9.0


sethrj:
Starting with Lapack 3.9.0, which is included by OpenBLAS 0.3.8+, the file `sorhr_col.f` includes an "external" declaration that extends just beyond the 72-column limit:
```fortran
      EXTERNAL           SCOPY, SLAORHR_COL_GETRFNP, SSCAL, STRSM, XERBLA
```
The final `A` is truncated by the Intel compiler, creating an invalid undefined symbol:
```console
$ nm SRC/sorhr_col.o | grep xerbl
                 U xerbl_
                 U xerbla_
```
leading to a link error
```
undefined reference to `xerbl_'
```

See https://github.com/spack/spack/issues/15385

PR #396, opened 2020-03-10
Obey Fortran fixed-form limit to fix missing 'xerbl'

sethrj:
There isn't enough room on the punch card to fit the trailing 'A' in
'XERBLA', leading to an undefined symbol reference to 'XERBL_' in the
stricter Intel compilers. Fixes https://github.com/Reference-LAPACK/lapack/issues/394, will fix an issue on https://github.com/xianyi/OpenBLAS/issues/2386#issuecomment-582096592 and will fix https://github.com/spack/spack/issues/15385 when it's merged upstream into OpenBLAS.

### DIFF
--- a/SRC/sorhr_col.f
+++ b/SRC/sorhr_col.f
@@ -282,7 +282,8 @@
      $                   NPLUSONE
 *     ..
 *     .. External Subroutines ..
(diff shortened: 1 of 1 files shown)


langou:
Hi @sethrj, Thanks! I just closed this request. Let me know if problems. Julien.

PR #398, opened 2020-03-18
Add more tests for edge cases in level-1 BLAS.

matcross:
At NAG we've observed issues with the BLIS version of IDAMAX and with the Apple vecLib version of ZCOPY in some edge cases.

We thought it would be useful to have additional test cases in place upstream here:

added test of I*AMAX when the input vector is constant;

added test of *COPY when the source and target have increment 0.


langou:
Excellent! Thanks!

PR #399, opened 2020-03-23
cmake: Tell CMake that our Fortran sources are in fixed format

bradking:
CMake will automatically generate the proper flag to tell each compiler
that our sources use fixed format.  Drop our manual use of `-qfixed` for
the XL Fortran compiler because it breaks CMake's FortranCInterface
detection (which uses free-format sources).


julielangou:
Thank you @bradking

PR #400, opened 2020-03-25
Added test of I?AMAX when there are multiple maxima.

matcross:
Sorry for the extra noise, this is an additional set of tests related to #398 following problems observed at NAG with the BLIS version of IDAMAX. It's the final set of additions we know of for the time being.


langou:
Thanks @matcross !

PR #402, opened 2020-03-30
Incorrect behaviour in real ?LACN2 under Fortran 95 SIGN semantics. Improved some ?COPY testing.

matcross:
Hi,

My colleague Edvin Hopkins at NAG noticed that SLACN2 and DLACN2 suffers from the change in meaning of SIGN at Fortran 95, with respect to negative zero. With DLACN2 this was manifesting as incorrect results in the DGEEVX example program we distribute with our numerical library.

The intended semantics can be seen in the original paper e.g. at http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.117.4839&rep=rep1&type=pdf (page 384), so we suggest inlining the expression as an IF block in order to uniformly force the correct behaviour.

I have also taken the opportunity to improve the new INCX==INCY=0 testing in ?blat1 in order to run the new tests only once and to display the correct INC values on failure.

### DIFF
--- a/BLAS/TESTING/cblat1.f
+++ b/BLAS/TESTING/cblat1.f
@@ -344,7 +344,8 @@
       LOGICAL           PASS
 *     .. Local Scalars ..
       COMPLEX           CA
-      INTEGER           I, J, KI, KN, KSIZE, LENX, LENY, MX, MY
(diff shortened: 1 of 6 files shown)


langou:
Thanks Mat!

Issue #401, opened 2020-03-28
DSYGVD thread-safety


MaximeVdB:
First of all: thanks a lot for developing Reference-LAPACK!

I've been trying to use both Netlib's DSYGVD routine (LAPACK v.3.9.0) inside a simple do-loop parallelized with OpenMP, and I kept encountering diagonalization errors (INFO>0) or segmentation faults when using more than 1 OpenMP thread.

This occurred when using LAPACK compiled using ifort 18.0.5 using the regular `INSTALL/make.inc.ifort` configuration file. This problem did not arise when using either
* a gfortran-compiled LAPACK using the included `INSTALL/make.inc.gfortran` configuration file (which includes the "-frecursive" FC flag),
* an ifort-compiled LAPACK using the included `INSTALL/make.inc.ifort` configuration file after adding an extra "-recursive" FC flag,
* the DSYGV routine instead of DSYGVD (regardless of the "-recursive"/"-frecursive" FC flag).

The addition of the "-frecursive" flag in the gfortran configuration file was motivated by bug 112, where the DSYEVR routine was found to be thread-unsafe:

[http://performance.netlib.org/lapack/errata_from_342_to_350.html](http://performance.netlib.org/lapack/errata_from_342_to_350.html)

So, does it indeed look like `INSTALL/make.inc.ifort` would benefit from having "-recursive" flags by default? And probably this also applies to other compilers? 

If you want I can also provide a (semi-)minimalistic example.

langou:
Hi Maxime, sounds good, can you provide us with a pull request? It sounds like a good idea to add the flag given your experience. Cheers, Julien.

PR #403, opened 2020-03-30
Recursive option for ifort configuration

MaximeVdB:
(see issue #401)

I don't about similar options for other compilers though, like  the Portland group and Sun compilers. But at least Intel Fortran is covered with this small change (the corresponding "-frecursive" option was already added to the GNU Fortran configuration file).

### DIFF
--- a/INSTALL/make.inc.ifort
+++ b/INSTALL/make.inc.ifort
@@ -16,9 +16,9 @@ CFLAGS = -O3
 #  the compiler options desired when NO OPTIMIZATION is selected.
 #
 FC = ifort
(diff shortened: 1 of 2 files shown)


martin-frbg:
-Mrecursive on PGI Fortran apparently (wouldn't it be nice if they at least honored each others' option syntax...)

MaximeVdB:
Thanks for the tip -- after getting the `pgf95` compiler, I ran the same (OpenMP-parallelized) DSYGVD test as for `gfortran` and `ifort` and I found the same behaviour (i.e. diagonalization errors when LAPACK was compiled without the "-Mrecursive" flag, and no errors after recompiling LAPACK with this flag).

PR #408, opened 2020-04-27
Fix the loop index scalar in ZHEEQUB.f

ilayn:
In #61 one of the files has been missed in terms of the loop index modifications and that in turn causes wildly different results depending on the versions due to the data being not scaled appropriately.

This PR completes the suite.

### DIFF
--- a/SRC/zheequb.f
+++ b/SRC/zheequb.f
@@ -271,7 +271,7 @@
          AVG = AVG / N
 
          STD = 0.0D0
-         DO I = N+1, N
+         DO I = N+1, 2*N
             WORK( I ) = S( I-N ) * WORK( I-N ) - AVG
          END DO
          CALL ZLASSQ( N, WORK( N+1 ), 1, SCALE, SUMSQ )


ilayn:
pinging @christoph-conrads as the previous author of the PR mentioned

christoph-conrads:
> pinging @christoph-conrads as the previous author of the PR mentioned

I don't remember this #61 but the issue description and the proposed change seem sound.

ilayn:
A gentle nudge for this

zerothi:
Could you please also fix this in `cheequb.f`?

zerothi:
I can confirm that in `[sd]syequb.f` it was already `do i = N+1, 2*N` and I think the fix seems correct. I don't however, have any other basis for this fix. :)

ilayn:
> Could you please also fix this in cheequb.f?

`cheequb.f` is not affected.

https://github.com/Reference-LAPACK/lapack/blob/master/SRC/cheequb.f#L274

zerothi:
> > Could you please also fix this in cheequb.f?
> 
> `cheequb.f` is not affected. I caught these by looking at the diff with #61 via git blame while debugging for something else.
> 
> https://github.com/Reference-LAPACK/lapack/blob/master/SRC/cheequb.f#L274

I also just looked, it isn't affected. Thanks for double checking!

ilayn:
I noticed that MKL is still missing this fix still in the newer versions though other vendors seem to have picked it up, possibly due to the missing milestone. I don't know anyone from MKL team and Intel forums are a different kind of fun to navigate. In case any folks reading this, maybe ping someone from that team, it would be highly appreciated.

langou:
Thanks Ilhan. I sent an email to Sarah Knepper and she acknowledged reception. We'll see. It is a good idea to contact them, thanks for suggesting, and I am sure they appreciate!

sknepper:
Thank you very much @ilayn and @langou ! This fix will be part of the upcoming oneMKL 2026.2 release.

Issue #410, opened 2020-04-28
CBLAS install step with USE_OPTIMIZED_BLAS=ON


derpda:
### Problem
I configure CBLAS (and LAPACKE) out of source with

```
cmake -DUSE_OPTIMIZED_BLAS=ON -DCBLAS=ON -DUSE_OPTIMIZED_LAPACK=ON -DLAPACKE=ON ..
```
Building (`make`) works without problems.
When I then use `make install`, `cblas-targets.cmake` does not get installed. 
However, the installed `cblas-config.cmake` directly refers to it, causing errors when it is used by `find_package`.

I have tracked this down to the fact that `ALL_TARGETS` will be empty at the point when it is checked in 
https://github.com/Reference-LAPACK/lapack/blob/8e1e16c6300861f096744a5ab75c3572c4b9b756/CBLAS/CMakeLists.txt#L51
Since this is my first time dealing with the LAPACK code, I don't quite understand why that variable is checked here in the first place.
In the comparable place of the LAPACKE code
https://github.com/Reference-LAPACK/lapack/blob/8e1e16c6300861f096744a5ab75c3572c4b9b756/LAPACKE/CMakeLists.txt#L129
no such check is performed.

### Solution?
If I comment out above mentioned check, installation works as intended and I can use `find_package(CBLAS)` as usual. 
Normally I would make a pull request, but I don't understand the purpose of `ALL_TARGETS` well enough to suggest proper changes right now.

### Environment
OS: Arch Linux (5.6.6-arch1-1)
CMake: 3.17.1

moe123:
@hello, (not directly related to, but somehow it is) for people writing C/C++ programs, we re-wrote from scratch a full reference implementation (minus complex; work in progress) of fortran-blas in vanilla C (no cblas):

https://github.com/moe123/macadam/tree/master/macadam/details/numa/lapack/blas

it can compile c-native row-major as well as column-major to mockup fortran intrinsics; however, everything is header inlined, hence make sure if binary size matter to you to bootstrap those calls within your program.

derpda:
@moe123 thanks for the info!
We mostly target supercomputers and rely on optimized BLAS/LAPACK installations, which is why we probably need to stick to CBLAS and LAPACKE for now...

derpda:
I'm also currently working on a pull request for this issue, but am hoping to hear from some contributers before I post it

moe123:
[...]

PR #413, opened 2020-05-04
Improvements of CMake (instalL) code for CBLAS, LAPACKE 

derpda:
### Introduction
This pull request is my own attempt at fixing issue #410 (which I have created myself).
The main problem was that the created CMake packages for CBLAS/LAPACKE would not be usable when providing preinstalled optimized BLAS/LAPACK.

This is my first attempt at contributing to this repository, so I would greatly appreciate feedback of any kind.

### Changes
The issue stemmed mainly from the CMake code for CBLAS, which differed significantly from that of LAPACKE. 
I was unable to figure out a reason for this difference, and removing it largely fixed the issues I had with the install instructions.
Specifically, checks for the content `ALL_TARGETS` were made inside `CBLAS/CMakeLists.txt` that I do not think were necessary.

While I believe the above to be just fixed, I have also made one significant change.
This concerns the case outlined in the introduction, where intend to build and install LAPACKE using optimized BLAS/LAPACK libraries. 
If we additionally want to use the TMG library content (`LAPACKE_WITH_TMG=ON`), but `tmg` was not built with the optimized LAPACK, then we need to also build (and install) the `tmg` library.
Before the changes made here, the CMake code would build `tmg` as part of the `lapack-targets` (as opposed to the `lapacke-targets`) and attempt to install an additional `lapack-config.cmake` file. 
In this case however, I think it would be better to instead build and install it as part of the `lapacke-targets`, since modifying an already existing package doesn't seem prudent (or might even not be possible, for example for lack of permissions). 

Making this change allowed the library we are developing to build and install CBLAS/LAPACKE (if they are not found) with optimized BLAS/LAPACK, all without caring whether or not `tmg` is included in the optimized libraries on the system. 
At least for our use case, this is a major portability improvement.

### To be done
In the case mainly of interest in this pull request, no targets associated with `lapack-targets` are thus built. Ideally, installing `lapack-config.cmake` would no longer be necessary, and only the CBLAS and LAPACKE CMake packages should be installed.
I have not made this particular change yet, and a (seemingly useless) `lapack-config.cmake` will also be installed. 
This might cause issues, and I would love to get some feedback on how to alleviate this.

### More ideas
While working on this pull request, I had a few ideas for things that could possibly be done in the future, but certainly require more thought.
I won't go into details here, but the gist is:
 - Possibly get rid of the `ALL_TARGETS` part of code (seems obsolete to me)
 - Make CBLAS and LAPACKE separate repositories (with git submodules for BLAS/LAPACK)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -469,10 +469,6 @@ if(NOT LATESTLAPACK_FOUND)
   set(ALL_TARGETS ${ALL_TARGETS} ${LAPACKLIB})
 endif()
 
-if(BUILD_TESTING)
-  set(ALL_TARGETS ${ALL_TARGETS} ${TMGLIB})
-endif()
-
 # Export lapack targets, not including lapacke, from the
 # install tree, if any.
 set(_lapack_config_install_guard_target "")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -353,9 +353,27 @@ option(LAPACKE_WITH_TMG "Build LAPACKE with tmglib routines" OFF)
 if(LAPACKE_WITH_TMG)
   set(LAPACKE ON)
 endif()
-if(BUILD_TESTING OR LAPACKE_WITH_TMG) #already included, avoid double inclusion
+
+# TMGLIB
+# Cache export target
+set(LAPACK_INSTALL_EXPORT_NAME_CACHE ${LAPACK_INSTALL_EXPORT_NAME})
+if(BUILD_TESTING OR LAPACKE_WITH_TMG)
+  if(LATESTLAPACK_FOUND AND LAPACKE_WITH_TMG)
+    set(CMAKE_REQUIRED_LIBRARIES ${LAPACK_LIBRARIES})
+    # Check if dlatms (part of tmg) is found
+    CHECK_FORTRAN_FUNCTION_EXISTS("dlatms" LAPACK_WITH_TMGLIB_FOUND)
+    unset(CMAKE_REQUIRED_LIBRARIES)
+    if(NOT LAPACK_WITH_TMGLIB_FOUND)
+      # Build and install TMG as part of LAPACKE targets (as opposed to LAPACK
+      # targets)
+      set(LAPACK_INSTALL_EXPORT_NAME ${LAPACKELIB}-targets)
+    endif()
+  endif()
   add_subdirectory(TESTING/MATGEN)
 endif()
+# Reset export target
+set(LAPACK_INSTALL_EXPORT_NAME ${LAPACK_INSTALL_EXPORT_NAME_CACHE})
+unset(LAPACK_INSTALL_EXPORT_NAME_CACHE)
 
 if(LAPACKE)
   add_subdirectory(LAPACKE)
@@ -451,7 +469,7 @@ if(NOT LATESTLAPACK_FOUND)
   set(ALL_TARGETS ${ALL_TARGETS} ${LAPACKLIB})
 endif()
 
-if(BUILD_TESTING OR LAPACKE_WITH_TMG)
+if(BUILD_TESTING)
   set(ALL_TARGETS ${ALL_TARGETS} ${TMGLIB})
 endif()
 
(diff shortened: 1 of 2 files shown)


drhpc:
So you're trying to fix up the CMake build situation. May I point to what we do with pkgsrc to get the lapack stuff built? See [wip/lapack/patches/](https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=tree;f=lapack/patches;hb=HEAD). I just now had to remove the lapacke-targets block at the end of LAPACKE/CMakeLists.txt to make this part work again. I am puzzled that the build worked before (other cmake version?).  I see that you're trying to fix that portion in CBLAS and LAPACKE instead of removing it.

I hope this pull request means we can drop at least that part of our patches in future.

Just general, as someone working on the CMake build of LAPACK, could you have a look at the other patches we carry, too? I'm really mostly cargo-coding with this, so it's perhaps best someone has a look before a pull request is made. Our changes are for supporting both static and dynamic libraries in one build and for separating the sub-packages. See [lapack/Makefile.common](https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=blob;f=lapack/Makefile.common;hb=HEAD) and the respective individual package Makefiles ([lapack/Makefile](https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=blob;f=lapack/Makefile;hb=HEAD), [lapacke/Makefile](https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=blob;f=lapacke/Makefile;hb=HEAD), [blas/Makefile](https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=blob;f=blas/Makefile;hb=HEAD), [cblas/Makefile](https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=blob;f=cblas/Makefile;hb=HEAD))  for the CMAKE_ARGS we use for this.

The split is necessary since we want to be able to combine CBLAS and LAPACKE with optimized BLAS/LAPACK implementations, mostly. Of course, since we got a working setup up, I don't really fancy upstream splitting the packages to cause us additional work;-)

Would be nice if upstream LAPACK would support that.

weslleyspereira:
Hi @derpda. Could you please rebase your branch? I would like to test it again. Thanks!

derpda:
Rebase complete.
A year has passed since I made this PR so I'm not quite sure if everything works as intended still.

derpda:
I'd be happy to help out with a general CMake overhaul, I think there is quite some baggage left.
Sine apparently the minimum required CMake version has been bumped to 3.2, I think many things could be improved.

CMakeLists.txt
```diff
@@ -451,7 +469,7 @@ if(NOT LATESTLAPACK_FOUND)
   set(ALL_TARGETS ${ALL_TARGETS} ${LAPACKLIB})
 endif()
 
-if(BUILD_TESTING OR LAPACKE_WITH_TMG)
+if(BUILD_TESTING)
```
weslleyspereira, line 472: Travis shows an error. See https://travis-ci.org/github/Reference-LAPACK/lapack/builds/766858282
```
CMake Error in CMakeLists.txt:

  given target "tmglib" more than once.
```

I think the issue can be solved by just removing these 3 lines from the `CMakeLists.txt`.

weslleyspereira:
Thanks for rebasing it! As you predicted, we will need some minor modifications to make it work.

> I'd be happy to help out with a general CMake overhaul, I think there is quite some baggage left.
> Sine apparently the minimum required CMake version has been bumped to 3.2, I think many things could be improved.

And I'd be happy to review it :)

CMakeLists.txt
weslleyspereira: May I commit directly to your branch?

derpda: Yes sorry! I was (and still am) very busy these past days so I didn't get around to it.
It's still in my to-do list, but I had to reduce the priority.
Feel free to commit, I have added you as a collaborator.

weslleyspereira: Ah ok. Thanks! I just asked to check if my suggestion makes sense. Normally, I have the permission to modify the PRs directly 😊

derpda: Thanks for making that change! Looks like all tests are passing

weslleyspereira (approved):

drhpc:
I reworked the pkgsrc CMake patches in #556 . Might wanna handle this PR and the other one together.

langou:
Thanks @derpda !

PR #417, opened 2020-05-21
Multishift QZ with AED

thijssteel:
This PR adds an implementation of the multishift QZ algorithm with AED.
It is loosely based on my implementation of the rational QZ algorithm (https://github.com/thijssteel/multishift-multipole-rqz).

It features:
- Agressive early deflation
- Multishift QZ sweeps using optimal packing of the bulges
- A new heuristic to select the number of positions in the sweep windows
It does not feature:
- A windowed deflation of infinite eigenvalues

This PR only contains the double precision code. My previous experience with Fortran is rather limited so I decided to get some feedback before finishing up the other types. If you want me to add the rest before reviewing it, that's fine too, just let me know.


thijssteel:
There are some pretty significant problems with this PR revealed by the tests for DGGEV3. Even though the framework reports that the tests have passed, the tolerances are not reached.

PR #420, opened 2020-06-21
Accuracy improvements to xtgex2

thijssteel:
This PR contains two improvements to xtgex2:

- Daan Camps has proven in his dissertation that replacing the condition in xtgex2 when swapping 1x1 blocks can improve the accuracy. This condition and the respective error tolerances have been changed to reflect it.

- Also by Daan Camps, "SWAPPING 2 × 2 BLOCKS IN THE SCHUR AND GENERALIZED
SCHUR FORM", shows that the accuracy of swapping 2x2 blocks can be improved by iterative refinement and by replacing the QR factorisation with an SVD based method. This PR only implements the first improvement as the paper does not mention how to avoid overflow.


thijssteel:
@julielangou most of the tests seem to avoid testing specific functions as much as possible. Is that by design or is it fine if i add some tests specifically for xtgexc to fix the coverage? The same goes for the xlaqz help functions for my other PR.

julielangou:
@langou , could you please take a look at this contribution.
For the testing no, this is not done on purpose. Better testings and a higher coverage of functions is what we aim for.

thijssteel:
I decided to remove the iterative refinement. I have tested that it works as expected. However, I can't even find any example that fails the threshhold (when 2x2 swaps fail, it fails at the calculation of dtgsy2). I suspect that Daan Camps used a more strict threshhold in his paper. The low diff coverage is also a consequence of the examples not failing the threshhold and should be ok to merge.

Additionally, lowering the threshhold or forcing iterative refinement some other way improves the backward error, but increases the error on the eigenvalues. If the accuracy ever becomes a problem, i'll have a look at implementing the svd swaps instead as those should also improve the error on the eigenvalues.

If you think it could be useful to add the refinement anyway in case there is a case where it actually works let me know, but it will not be covered by the unit tests.

TESTING/EIG/dchkec.f
```diff
@@ -189,9 +189,15 @@ SUBROUTINE DCHKEC( THRESH, TSTERR, NIN, NOUT )
          OK = .FALSE.
          WRITE( NOUT, FMT = 9991 )RLAQTR, LLAQTR, NLAQTR, KLAQTR
       END IF
+*
+      CALL DGET40( RTGEXC, LTGEXC, NTGEXC, KTGEXC, NIN )
+      IF( RLAQTR.GT.THRESH ) THEN
```
weslleyspereira, line 194: Hi @thijssteel. Does my suggestion make sense? Thanks.
```suggestion
      IF( RTGEXC.GT.THRESH ) THEN
```

thijssteel: yes, very good catch. I see i made the same mistake in the single precision version.

weslleyspereira:
Nice job @thijssteel ! I think we can accept it now. Thank you !

PR #434, opened 2020-09-01
LAPACKE/include/lapack.h: fix C++ compatibility and prototypes for xGGSVD/xGGSVP

martin-frbg:
Backports from OpenBLAS : 
1.  need to include <complex> rather than <complex.h> in a C++ context (this was also created as issue #320 here a while ago)
and need to move the block of LAPACK_COMPLEX_CUSTOM declarations outside the `extern "C"`
2. when prototypes for several deprecated LAPACK functions were restored in #367, xGGSVD and xGGSVP erroneously got their LAPACKE prototypes added instead of the LAPACK ones.


julielangou:
Thank you @martin-frbg , closing #320

h-vetinari:
I'll allow myself to note that https://github.com/Reference-LAPACK/lapack/pull/434/commits/14740d3aa31b45f284c4302d348ab8016107f3a5 is essentially https://github.com/Reference-LAPACK/lapack/commit/127f0bfa04ba1f327fcea63f1771fe4fc5537048, which has been lying around for a couple of months.

martin-frbg:
Seems you are right, sorry for not noticing that but I guess our different styles of writing commit messages may be to blame.

PR #438, opened 2020-09-04
Fix some incorrect SLASET/DLASET calls (includes backport of openblas#2778)

### LINKED ISSUE #318: out of the array bound in SRC/slaset.f
The previous issues with some array access in LAPACK 3.8.0 have been fixed, so please cf.  #312. 

Here I'm reporting the other out of array bound issue in ./EIG/xeigtsts. What I did is same as before: LAPACK 3.8.0, GCC (gfortran v8.2.1), appended "-fcheck=all" in "OPTS" of make.inc, then run "make all".

Below is the detail:
...
./EIG/xeigtsts < se2.in > sse2.out 2>&1
make[1]: *** [sse2.out] Error 2
make[1]: Leaving directory `../lapack/TESTING'
make: *** [lapack_testing] Error 2

The content of **sse2.out** is:

At line 169 of file **slaset.f**
Fortran runtime error: Index '2' of dimension 1 of array 'a' above upper bound of 1

Error termination. Backtrace:
#0 0x10000042ddaf in ???
#1 0x10000042f7e7 in ???
#2 0x10000043019b in ???
#3 0x101c9107 in ???
#4 0x1002889b in ???
#5 0x10005a13 in ???
#6 0x10001313 in ???
#7 0x100000764bf7 in ???
#8 0x100000764e03 in ???
#9 0xffffffffffffffff in ???
Tests of the Symmetric Eigenvalue Problem routines

LAPACK VERSION 3.8.0

The following parameter values will be used:
M: 0 1 2 3 5 20
N: 0 1 2 3 5 20
NB: 1 3 3 3 10
NBMIN: 2 2 2 2 2
NX: 1 0 5 9 1

Relative machine underflow is taken to be 0.117549E-37
Relative machine overflow is taken to be 0.340282E+39
Relative machine precision is taken to be 0.596046E-07

Routines pass computational tests if test ratio is less than 50.00
SST routines passed the tests of the error exits (209 tests done)
SE2: NB = 1, NBMIN = 2, NX = 1

Thanks a lot for your time again!
Collin

### LINKED ISSUE #425: Incorrect call to SLASET in schkst2stg.f
This appears to be related to #318 [https://github.com/Reference-LAPACK/lapack/issues/318](url). The problem is not in SLASET but in the calling routine.

TESTING/EIG/schkst2stg.f contains the following code fragment (both in 3.9.0 and in the repo):

1000 *           using the 1-stage.
1001 *
1002             CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
1003             CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
1004             CALL SLACPY( "U", N, N, A, LDA, V, LDU )
1005             LH = MAX(1, 4*N)
1006             LW = LWORK - LH

This code is incorrect, and causes a bounds violation in SLASET. The correct code is:

1002             CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, N )
1003             CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, N )

### LINKED ISSUE #429: Incorrect calls to DLASET in TESTING/EIG/cchkhb2stg.f
cchkhb2stg.f contains a number of calls to DLASET, in the form:

               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )

These calls are wrong, for three reasons:

1. SD and SE are REAL variables of the default kind, while DLASET expects a DOUBLE PRECISION real kind. There is no guarantee that the underlying binary representation of zero is the same for both types, or how DLASET would interpret default real-kind ZERO, when it expects a DOUBLE PRECISION value.

2. The calling routine expects SD and SE to be (N) REAL elements in size. It is quite likely that DLASET will initialize more than the expected size, corrupting something else.

3. The LDA argument to DLASET (1) is potentially less than N, and can trigger a bounds violation in DLASET.

One might be tempted to replace the calls with:

               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, N )
               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, N )

... however, I have no clue whether this is what the author of the code really intended.

h-vetinari:
@serguei-patchkovskii observed some errors and suggested some fixes in #425 & #429 that were picked up by @martin-frbg in xianyi/OpenBLAS#2778. Serguei [said](https://github.com/Reference-LAPACK/lapack/issues/429#issuecomment-674513208) he won't have time to prepare a backport-PR, and it seems (after ~3 weeks) neither did Martin.

I have left the commits from xianyi/OpenBLAS#2778 intact; just added `Suggested-By: Serguei Patchkovskii <Serguei.Patchkovskii@gmail.com>` to reflect Martin's [comment](https://github.com/Reference-LAPACK/lapack/issues/429#issuecomment-674447761

Relates to #318
Closes #425 
Closes #429

Edit: after some more explanation by @serguei-patchkovskii & tips by @martin-frbg, I investigated a bit more and I believe this PR now additionally:
Closes #318

### DIFF
--- a/TESTING/EIG/cchkhb2stg.f
+++ b/TESTING/EIG/cchkhb2stg.f
@@ -680,8 +680,8 @@
 *              the one from above. Compare it with D1 computed 
 *              using the DSBTRD.
 *            
-               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
-               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, N )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, N )
                CALL CLACPY( ' ', K+1, N, A, LDA, U, LDU )
                LH = MAX(1, 4*N)
                LW = LWORK - LH
@@ -753,8 +753,8 @@
 *              the one from above. Compare it with D1 computed 
 *              using the DSBTRD. 
 *           
-               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
-               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, N )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, N )
                CALL CLACPY( ' ', K+1, N, A, LDA, U, LDU )
                LH = MAX(1, 4*N)
                LW = LWORK - LH
--- a/TESTING/EIG/cchkhb2stg.f
+++ b/TESTING/EIG/cchkhb2stg.f
@@ -680,8 +680,8 @@
 *              the one from above. Compare it with D1 computed 
 *              using the DSBTRD.
 *            
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
                CALL CLACPY( ' ', K+1, N, A, LDA, U, LDU )
                LH = MAX(1, 4*N)
                LW = LWORK - LH
@@ -753,8 +753,8 @@
 *              the one from above. Compare it with D1 computed 
 *              using the DSBTRD. 
 *           
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
                CALL CLACPY( ' ', K+1, N, A, LDA, U, LDU )
                LH = MAX(1, 4*N)
                LW = LWORK - LH
--- a/TESTING/EIG/cchkst2stg.f
+++ b/TESTING/EIG/cchkst2stg.f
@@ -1014,8 +1014,8 @@
 *           the one from above. Compare it with D1 computed 
 *           using the 1-stage.
 *
-            CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
-            CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
+            CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, N )
+            CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, N )
             CALL CLACPY( 'U', N, N, A, LDA, V, LDU )
             LH = MAX(1, 4*N)
             LW = LWORK - LH
@@ -1048,8 +1048,8 @@
 *           the one from above. Compare it with D1 computed 
 *           using the 1-stage. 
 *
-            CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
-            CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
+            CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, N )
+            CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, N )
             CALL CLACPY( 'L', N, N, A, LDA, V, LDU )
             CALL CHETRD_2STAGE( 'N', "L", N, V, LDU, SD, SE, TAU, 
      $                   WORK, LH, WORK( LH+1 ), LW, IINFO )
--- a/TESTING/EIG/dchksb2stg.f
+++ b/TESTING/EIG/dchksb2stg.f
@@ -670,8 +670,8 @@
 *              the one from above. Compare it with D1 computed 
 *              using the DSBTRD.
 *            
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
+               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, N )
+               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, N )
                CALL DLACPY( ' ', K+1, N, A, LDA, U, LDU )
                LH = MAX(1, 4*N)
                LW = LWORK - LH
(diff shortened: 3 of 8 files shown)


TESTING/EIG/cchkhb2stg.f
```diff
@@ -680,8 +680,8 @@ SUBROUTINE CCHKHB2STG( NSIZES, NN, NWDTHS, KK, NTYPES, DOTYPE,
 *              the one from above. Compare it with D1 computed 
 *              using the DSBTRD.
 *            
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
```
h-vetinari, line 684: Now that I compare the diff with the OP from #429, @serguei-patchkovskii actually suggested the following

```suggestion
               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, N )
               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, N )
```

Was it intentional to implement this differently @martin-frbg?

TESTING/EIG/cchkhb2stg.f
```diff
@@ -753,8 +753,8 @@ SUBROUTINE CCHKHB2STG( NSIZES, NN, NWDTHS, KK, NTYPES, DOTYPE,
 *              the one from above. Compare it with D1 computed 
 *              using the DSBTRD. 
 *           
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
-               CALL DLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, 1 )
+               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 )
```
h-vetinari, line 757:
```suggestion
               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SD, N )
               CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, N )
```

Same here

h-vetinari (commented):
Stumbled over a difference in the suggestion from @serguei-patchkovskii and the implementation from @martin-frbg.

PTAL

martin-frbg:
Intentional, and also the reason why I did not commit the PR immediately, and/or copy it here - as far as I can tell, specifying "1" instead of "N" is consistent with current usage in the test code, while for a general case serguei-patchkovskii's assessment seems correct.

serguei-patchkovskii:
> Intentional, and also the reason why I did not commit the PR immediately, and/or copy it here - as far as I (and valgrind) can tell, specifying "1" instead of "N" is consistent with current usage in the test code, while for a general case serguei-patchkovskii's assessment seems correct.

@martin-frbg is absolutely correct: this is the current usage in the code. This usage is contrary to the letter of the standard, and should be corrected. Let me explain in detail.

The relevant fragments of lapack/SRC/slaset.f are:

      SUBROUTINE SLASET( UPLO, M, N, ALPHA, BETA, A, LDA )

      REAL               A( LDA, * )

*
*        Set the leading m-by-n submatrix to ALPHA.
*
         DO 60 J = 1, N
            DO 50 I = 1, M
               A( I, J ) = ALPHA
   50       CONTINUE
   60    CONTINUE


In the calling routine, we have the call: CALL SLASET( 'Full', N, 1, ZERO, ZERO, SE, 1 ), where SE is declared as a 1-dimensional array, with the at least SE(1:N) index range. In SLASET, we are trying to set A(1:N,1:1) = ZERO. This is not pretty, but as far as I can tell is allowed under the Fortran sequence association rules for Fortran77-style calls.

Unfortunately, the matrix A in this call is declared as A(1:1,*) - so accessing the first index beyond 1 is a bounds violation, and explicitly invokes undefined behaviour. Therefore, I would claim that all calls to SLASET/DLASET with the argument LDA<M are incorrect - even if they are the current usage. While it might seem to work, it is not hard to imagine a standard-conforming implementation, where it will generate incorrect code (e.g. an implementation which relies on the callee performing copy-in/copy-out for array arguments will produce unexpected results).

If I had to guess why it wasn't spotted and corrected so far, I would speculate that this is an archaism in many compilers, which goes back to FORTRAN-IV. Before Fortran77 (which as far as I recall introduced the A(*) array bound declaration), the commonly-used extension was to declare arrays with an unknown upper bound was to give them with the upper bound of 1. For this reason, many Fortran compilers still treat the upper bound of "1" same as "*" in bounds checking, and do not flag a bounds violation.

So this code might have been semi-OK under the 60-year-old language definition - and has been grandfathered by compilers ever since. Quite possibly because every self-respecting Fortran compiler must be able to compile LAPACK :P

h-vetinari:
@serguei-patchkovskii Thanks for the explanation!

h-vetinari:
@serguei-patchkovskii 
The requirement that you bring up is also explicitly noted in the docstrings of [SLASET](https://github.com/Reference-LAPACK/lapack/blob/v3.9.0/SRC/slaset.f#L95) & [DLASET](https://github.com/Reference-LAPACK/lapack/blob/v3.9.0/SRC/dlaset.f#L95).

I grepped around a bit and found some more occurrences, which I fixed in the last two commits. However, I make no claims of exhaustiveness (in particular, I didn't check all cases where the `LDA` argument didn't show up due to a linebreak)

h-vetinari:
Looking also at #318 in particular, one would be tempted to actually implement the check for `LDA >= max(1,M)` in SLASET/DLASET (and throw with XERBLA if violated). Would that make sense?

Also CC @langou, since #318 is marked "Priority: High"

martin-frbg:
@h-vetinari From the original report in #318 it seems running a testsuite build with `-fcheck=all` would be sufficient (and more effective than `grep`) for catching all current offenders.

h-vetinari:
@martin-frbg, thanks for the tip (although I needed to find https://github.com/Reference-LAPACK/lapack/commit/33258cb4b25d50c721646ed97bf22a10724136f3 to understand where exactly to add `-fcheck=all`).

In any case, I ran a [build](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=205734&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d) in https://github.com/conda-forge/lapack-feedstock/pull/32 with the current patch & `-fcheck=all`, and while there were some errors, none of them come from SLASET/DLASET anymore (I had first run `-fcheck=all` without the patch, and the bounds violations for SLASET/DLASET were raising errors).

For the other errors that were uncovered, I opened #440.

julielangou:
Thanks again Martin!

martin-frbg:
all glory to h-vetinary and serguei-patchkovskii :smile:

h-vetinari:
It was a team effort! 💪 🙃

PR #444, opened 2020-09-17
Fix OSX builds on Travis CI

martin-frbg:
gcc9 comes already installed now, upgrade it to gcc10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
     env: CMAKE_BUILD_TYPE=Coverage
     before_install:
     - brew update > /dev/null
-    - brew install gcc > /dev/null
+    - brew upgrade gcc > /dev/null
 
 script:
   - export PR=https://api.github.com/repos/$TRAVIS_REPO_SLUG/pulls/$TRAVIS_PULL_REQUEST
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
     env: CMAKE_BUILD_TYPE=Release
     before_install:
     - brew update > /dev/null
-    - brew install gcc > /dev/null
+    - brew upgrade gcc > /dev/null
   - os: osx
     env: CMAKE_BUILD_TYPE=Coverage
     before_install:


martin-frbg:
Not sure I understand the codecov failure here, in code not touched by the patch, or is it simply that the Travis jobs have not run to completion for years ?

langou:
I am confused by the codecov report. I think it ran quite recently. Not sure why we are down at 81.86%. I would not worry about it then. Maybe they changed their algorithm.

martin-frbg:
I must admit I know too little about Codecov, at least the Codecov-patch seems to be satisfied that the PR itself does not change coverage.

ilayn:
@thomasrockhu Care to help us a bit here?

thomasrockhu:
Sure @ilayn. **TL;DR The HEAD commit has coverage two reports including an OSX build which contains more coverage information, while the base commit only has the Linux report**

So you are right, Codecov received coverage reports quite recently. You can view the commits [here](https://codecov.io/gh/Reference-LAPACK/lapack/pull/444/commits).

To investigate the coverage drop, we can view the coverage reports uploaded to Codecov between the [base](https://codecov.io/gh/Reference-LAPACK/lapack/commit/0148a3625cf4f738c067b76f30822b31b19a6893/build) and the [HEAD](https://codecov.io/gh/Reference-LAPACK/lapack/commit/9861f27c0c8a0d3dead73b77bd1e6e719443b014/build)

In order to help show the coverage changes are coming from the coverage reports, I'm just going to focus on `lsamen.f`. In the [coverage diff](https://codecov.io/gh/Reference-LAPACK/lapack/pull/444/changes), we see:

image

line 103 without coverage and line 114 with coverage. So let's take a look at the base commit to see what is happening first. 

Looking at the [file diff](https://codecov.io/gh/Reference-LAPACK/lapack/src/0148a3625cf4f738c067b76f30822b31b19a6893/SRC/lsamen.f) I notice that lines 103 and 114 don't have coverage supplied:

image

I can confirm this is true by looking at the [uploaded coverage report](https://codecov.io/api/gh/Reference-LAPACK/lapack/download/build?path=v4/raw/2020-09-16/86338083E8155A7D64AEB744FE5D75B1/0148a3625cf4f738c067b76f30822b31b19a6893/a883a416-3265-46c1-a46e-640bf94c1542.txt) for this commit:

image
(note that there is no line 103 and 114)

Going to the HEAD commit, I can look at its [file diff](https://codecov.io/gh/Reference-LAPACK/lapack/src/9861f27c0c8a0d3dead73b77bd1e6e719443b014/SRC/lsamen.f), I see that 

image

line 103 is uncovered and line 114 is covered. This matches up with the [uploaded coverage reports](https://codecov.io/gh/Reference-LAPACK/lapack/commit/9861f27c0c8a0d3dead73b77bd1e6e719443b014/build) (note there are two coverage reports)
image

image

The most common (and 9 times out of 10) reason for changes like this are idempotent tests, coverage reports failing to upload, or CI's failing (which often results in the Codecov step not running). In this case, we have another build running which does not appear in the base commit. This is the [OSX build](https://travis-ci.org/github/Reference-LAPACK/lapack/jobs/727965111). I would investigate there first.

The Codecov-patch is the coverage for new code added or changed. So it tells you of the code introduced, what is the coverage on it? This is useful for repositories that are targeting a general increase in coverage for a project. The Codecov-project is the overall coverage for the entire repository.

martin-frbg:
So in short this is a consequence of the CI job failing before my fix for it (that triggered this question) ?

thomasrockhu:
@martin-frbg seems like it

martin-frbg:
Thanks, seems I need to try harder to really break something here :)

Issue #263, opened 2018-07-05
dhseqr gives wrong result


andreasnoack:
This program
```f95
program bug
    implicit none

    double precision :: H1(4, 4), H2(4, 4), wr1(4), wi1(4), wr2(4), wi2(4), Z, work(10), dble, epsilon
    integer*8 :: info

    H1(:,:) = 0

    H1(2, 1) = dble(z'bfdf916d32df0e1d')
    H1(3, 2) = dble(z'bf782807624514d9')
    H1(4, 3) = dble(z'bf80d94d89578784')
    H1(1, 2) = dble(z'3fdf916d32df0e1d')
    H1(2, 3) = dble(z'3f782807624514da')
    H1(3, 4) = dble(z'3f80d94d89578784')

    H2(:,:) = H1
    H2(4,4) = epsilon(H2(4,4))

    call dhseqr('E', 'N', 4_8, 1_8, 4_8, H1, 4_8, wr1, wi1, Z, 1_8, work, 10, info)
    call dhseqr('E', 'N', 4_8, 1_8, 4_8, H2, 4_8, wr2, wi2, Z, 1_8, work, 10, info)

    write(*,*) 'wr1: ', wr1
    write(*,*) 'wi1: ', wi1
    write(*,*) 'wr2: ', wr2
    write(*,*) 'wi2: ', wi2

end program bug
```

produces 

```
 wr1:    0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000
 wi1:   0.73992959728054897      -0.73992959728054897        8.2263841908860064E-003  -8.2263841908860064E-003
 wr2:    0.0000000000000000        0.0000000000000000        1.1102230246251565E-016   1.1102230246251565E-016
 wi2:   0.49328639818703252      -0.49328639818703252        8.2263841908860116E-003  -8.2263841908860116E-003
```

on my machine with LAPACK 3.8.

I don't think there should be a conditioning issue here. I guess the convergence criterion might be confused by the zero diagonal.

thijssteel:
i've had a crack at this bug. I'm not sure but it looks like it's caused by underflow/rounding errors in dlanv2.

The following happens during the iterations:
- If the shift is purely imaginary, the structure will be preserved
- because we compare with the diagonal elements, we would need H(3,2) to be under sfmin to get a deflation
- after 10 iterations, we use the ad hoc shift (which is not purely imaginary) and get a deflation

[...]

PR #445, opened 2020-09-18
fix dlanv2 and slanv2

thijssteel:
My attempt at fixing #263

The problem was underflow (and it's probably also weak to overflow), so the fix is simply to scale the relevant parts in a similar way to XLARTG.

I've also gone ahead and included the changes suggested by Martin so we don't get an infinite loop in case of Inf or Nan.

### DIFF
--- a/SRC/dlanv2.f
+++ b/SRC/dlanv2.f
@@ -140,13 +140,16 @@
 *
 *     .. Parameters ..
       DOUBLE PRECISION   ZERO, HALF, ONE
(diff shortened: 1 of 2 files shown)


thijssteel:
this is going to need some more testing before merging, those coverage changes seem pretty suspicious

martin-frbg:
Still unsure what to make of codecov reports even after thomasrockhu's explanation in #444 - maybe this is still an artefact of just having restored the travis runs (which contain coverage builds as well) yesterday ?

thijssteel:
oh, #444 does explain a lot, I've been going crazy trying to find the cause.

martin-frbg:
LGTM, thanks

PR #449, opened 2020-09-26
[BLAS] Remove unused NCOLA variable

timleslie:
The variable `NCOLA` was computed but not used in the files `*gemm.f`, This PR removes this variable and updates the appropriate comments.

### DIFF
--- a/BLAS/SRC/cgemm.f
+++ b/BLAS/SRC/cgemm.f
@@ -215,7 +215,7 @@
 *     ..
 *     .. Local Scalars ..
       COMPLEX TEMP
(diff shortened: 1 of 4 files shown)


martin-frbg:
Bet the management of Nitro-Cola (or was it Nuclear Cola ?) will hate you for this - the unused variable must have been there for something like 12 years at least.(I do note that not even gfortran10 -Wunused-but-set-variable
would flag this - was this from visual inspection or did some other compiler complain ?)

timleslie:
This was from visual inspection. I'm amazed this wasn't flagged by compiler warnings too! 🤷

JHenneberg:
nice catch

PR #452, opened 2020-10-14
Use ZX, ZY rather than CX, CY in ZDROT

timleslie:
The canonical names for double precision complex vectors are `ZX` and `ZY`. This fixes `zdrot`, which was using `CX` and `CY`, which imply single precision complex vectors.


langou:
OKey doc. The change sounds good to me. Thanks @timleslie

PR #453, opened 2020-10-14
Mark xrotg() parameters as [in,out]

timleslie:
The `SA`, `SB` and `DA`, `DB` parameters of `SROTG` and `DROTG`, and `CA` parameters of `CROTG` and `ZROTG` are used to return the `r` and `z` value of the Givens rotation calculation. This PR updates the inline docs to represent this.


langou:
Good catch. Good fix. Thanks @timleslie

PR #454, opened 2020-10-16
[BLAS] Skip calculation of ROE when SCALE is zero

timleslie:
These changes in `SROTG` and `DROTG` will avoid an unnecessary comparison when `SCALE` is zero.


langou:
Yes, that's a good idea. Thanks @timleslie for the PR

PR #455, opened 2020-10-17
[BLAS] Fix missing logic path in xrotmg.f

timleslie:
This fixes a bug in `SROTMG` and `DROTMG` where there is a missing code path.

I traced the bug back to the following commit: https://github.com/Reference-LAPACK/lapack/commit/1767365dbf4c05eb481a920a5c59beb20cbe9ba5#diff-17bc8cba27cfa755d1503dee5651235a35de951b677ecaa5f3adb596137005a2

In particular, the original code had:

```fortran
  IF (.NOT.DU.LE.ZERO) GO TO 30
*         GO ZERO-H-D-AND-DX1..
  GO TO 60
```

The code at `60` provided the `ZERO-H-D-AND-DX1` operations, which is what was lost in the referenced commit. This PR puts this code path back in.

It's worth noting that according to file:///Users/timl/Downloads/Restructuring_the_BLAS_Level_1_Routine_for_Computi.pdf (bottom of page 7) this code path may not actually be reachable, so this change is possibly just for completeness, it may not fix any real world cases.

### DIFF
--- a/BLAS/SRC/drotmg.f
+++ b/BLAS/SRC/drotmg.f
@@ -153,6 +153,9 @@
              DD2 = DD2/DU
              DX1 = DX1*DU
            ELSE
+*            This code path if here for safety. We do not expect this
+*            condition to ever hold except in edge cases with rounding
+*            errors. See DOI: 10.1145/355841.355847
              DFLAG = -ONE
              DH11 = ZERO
              DH12 = ZERO
--- a/BLAS/SRC/drotmg.f
+++ b/BLAS/SRC/drotmg.f
@@ -152,6 +152,16 @@
              DD1 = DD1/DU
              DD2 = DD2/DU
              DX1 = DX1*DU
+           ELSE
+             DFLAG = -ONE
+             DH11 = ZERO
+             DH12 = ZERO
+             DH21 = ZERO
+             DH22 = ZERO
+*
+             DD1 = ZERO
+             DD2 = ZERO
+             DX1 = ZERO
            END IF
          ELSE
 
--- a/BLAS/SRC/srotmg.f
+++ b/BLAS/SRC/srotmg.f
@@ -153,6 +153,9 @@
              SD2 = SD2/SU
              SX1 = SX1*SU
            ELSE
+*            This code path if here for safety. We do not expect this
+*            condition to ever hold except in edge cases with rounding
+*            errors. See DOI: 10.1145/355841.355847
              SFLAG = -ONE
              SH11 = ZERO
              SH12 = ZERO
--- a/BLAS/SRC/srotmg.f
+++ b/BLAS/SRC/srotmg.f
@@ -152,6 +152,16 @@
              SD1 = SD1/SU
              SD2 = SD2/SU
              SX1 = SX1*SU
+           ELSE
+             SFLAG = -ONE
+             SH11 = ZERO
+             SH12 = ZERO
+             SH21 = ZERO
+             SH22 = ZERO
+*
+             SD1 = ZERO
+             SD2 = ZERO
+             SX1 = ZERO
            END IF
          ELSE
 


martin-frbg:
Hmm - if a code path is unreachable and was removed by a cleanup, I am not sure it should be put back "for completeness"  unless there is evidence that it _is_ actually reachable ? (At the very least, nobody appears to have missed it since its removal, 9 years ago ?)

langou:
Hi @timleslie, you write: ``This fixes a bug in SROTMG and DROTMG``. Can you explain the bug that is fixed? How you discovered the bug, etc.? It would be good for many reasons that we understand what these few lines of codes. I think we must have SU >=0 and so it is possible that SU=0 in some cases. And actually we might have SU < 0. Is that what is going on? And so this is a piece of code that handles the rare case when SU <=0. (So we are essentially assuming that SU = 0 in this case. Which sounds good to me.) @timleslie: can you explain a tat? Cheers, Julien.

timleslie:
Here's the correct link to the paper I referenced rather than a local copy 😅  https://www.researchgate.net/publication/2388100_Restructuring_the_BLAS_Level_1_Routine_for_Computing_the_Modified_Givens_Transformation

> Hmm - if a code path is unreachable and was removed by a cleanup, I am not sure it should be put back "for completeness" unless there is evidence that it _is_ actually reachable ? (At the very least, nobody appears to have missed it since its removal, 9 years ago ?)

Yep, I think you're probably right. if the code path isn't reachable then the alternate change here would be to simply remove the check of `IF (DU.GT.ZERO) THEN...`. My concern is that either a) the check is necessary, in which case we have a missing code path or b) the check isn't necessary, in which case we have an extra check that can be removed.

I don't have any evidence that the check is necessary, but I also haven't run any experiments to try to find a set of values which can trigger the code path.

> Hi @timleslie, you write: `This fixes a bug in SROTMG and DROTMG`. Can you explain the bug that is fixed? How you discovered the bug, etc.? It would be good for many reasons that we understand what these few lines of codes. I think we must have SU >=0 and so it is possible that SU=0 in some cases. And actually we might have SU < 0. Is that what is going on? And so this is a piece of code that handles the rare case when SU <=0. (So we are essentially assuming that SU = 0 in this case. Which sounds good to me.) @timleslie: can you explain a tat? Cheers, Julien.

Calling it a bug is perhaps overstating the situation. The potential bug in this case is that if we have `SU <= 0` then the value of SFLAG and the other SPARAMS are never set. I'm not sure how different compilers will handle this but I suspect it would result in undefined behaviour.

As noted above, this is just a potential bug, I don't have any evidence yet that indicates it can possibly be triggered. This extract from linked article notes that it would require some very specific rounding conditions to exist.

Screen Shot 2020-10-18 at 12 23 18 pm

That said, if we think there is an outside chance that the condition could be met then the code should account for it explicitly.

I found this issue by inspection of the code. I'm trying to gain a deep understanding of the BLAS library for my own interest and I noticed this missing logic path.

The algorithm involved is best described here (see Appendix): https://www.researchgate.net/publication/220492248_Basic_linear_algebra_subprograms_for_FORTRAN_usage

It notes that `0 < u <= 2` should hold true, which implies that the check is actually unnecessary.

langou:
Thanks @timleslie for explaining. This helps a lot. So we have a condition that we do not think will ever happen. (Per the paper that you quote.) And the current code has a conditional `IF ( something that we think we always be true ) . . . ENDIF`. And you are arguing that we should either remove the conditional or have an `ELSE` statement just in case. Thanks for explaining. And refreshing my memory with the paper of Tim Hopkins. Okey doc, so my opinion is that we should accept your PR. I would be in favor to have a quick comment just after the ELSE to explain that `this ELSE is protective coding and we do not think we should enter the ELSE statement`. 

Note: I was surprised to see that the CODECOV report did not report a decrease in code coverage. This PR should decrease the code coverage since we now have some line of codes that our test suite is not entering. Maybe `SROTMG` and `DROTMG` are not tested in our test suite.

timleslie:
> Okey doc, so my opinion is that we should accept your PR. I would be in favor to have a quick comment just after the ELSE to explain that `this ELSE is protective coding and we do not think we should enter the ELSE statement`.
> 

Yep, I think that sounds good, I'll add that comment and add a reference to the Hopkins paper for the next person who encounters this code.

> Note: I was surprised to see that the CODECOV report did not report a decrease in code coverage. This PR should decrease the code coverage since we now have some line of codes that our test suite is not entering. Maybe `SROTMG` and `DROTMG` are not tested in our test suite.

I'm not yet familiar enough with the test suite to say for sure, but I agree that it makes sense that the coverage would decrease with this change.

PR #458, opened 2020-10-22
Fix macro used in LAPACKE_zgesvdq

nforro:
This looks like a copy-paste error. gcc emits a warning about it:

```
lapacke_zgesvdq.c: In function 'LAPACKE_zgesvdq':
../include/lapacke.h:50:42: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
```


langou:
Thanks!

Issue #459, opened 2020-10-27
CBLAS 64bit index build is not possible


epsilon-0:
It seems like CBLAS does not have a index-64 API possible due to not having a variable used for function definition similar to LAPACKE, like `lapack_int`.

This creates a tiny problem as it seems like OpenBLAS and BLIS seem to have a index-64 header defined, but it becomes hard to
check if they are compatible as there is no reference implementation.

Is there any interest in creating such an API?
I would be willing to do a small PR which is going to try and extend it but I don't want to do that if there are other efforts already in progress, which might conflict with mine.

Thanks a lot,
Aisha

epsilon-0:
Seems like there is an old issue also referencing this issue - https://github.com/Reference-LAPACK/lapack/issues/184

epsilon-0:
Fixed :smile_cat:

Issue #184, opened 2017-09-07
cblas cannot be compiled with int64


chs040701:
As long as BLAS, Lapack and Lapacke can use 64-bit integer, cblas is fixed with "int" type……

cdluminate:
Any update on this? May I work on patching CBLAS for ILP64/64-bit-indexing support and submit a PR?

weslleyspereira:
Closed by #460.

PR #460, opened 2020-10-27
enable building the cblas64 library

epsilon-0:
changes all function signatures to use CBLAS_INDEX
also fixes internal variables to be CBLAS_INDEX
all tests are passing

---

A lot of the optimized implementations are providing an implementation for CBLAS64
It is good to have the reference implementation also providing this, as we already have the 64 bit libraries for others.

I've followed your standard syntax rules from other files.
The variable name used is the standard CBLAS_INDEX which should keep it compatible with all other files.

Would be **really** awesome to finally have this reference implementation :heart: 

Thanks a lot,
Aisha

### DIFF
--- a/CBLAS/include/cblas.h
+++ b/CBLAS/include/cblas.h
@@ -13,7 +13,7 @@ extern "C" {            /* Assume C declarations for C++ */
 #ifdef WeirdNEC
    #define CBLAS_INDEX long
 #else
-    #define CBLAS_INDEX int
+   #define CBLAS_INDEX int
 #endif
 
 typedef enum {CblasRowMajor=101, CblasColMajor=102} CBLAS_LAYOUT;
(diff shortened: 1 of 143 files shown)


langou:
I am in favor of this PR.

martin-frbg:
I'm certainly not against it either, even if it means losing one competitive advantage of OpenBLAS :smiley: 
Just for cross-referencing: this is expected to fix both #459 and #184

epsilon-0:
> I'm certainly not against it either, even if it means losing one competitive advantage of OpenBLAS smiley
> Just for cross-referencing: this is expected to fix both #459 and #184

XD
I'd say this is not going to make OpenBLAS lose an advantage.
In fact, I'd say this should actually be a positive note for OpenBLAS (and any other provider, like blis), because now it can even say that it provides an optimized **reference** CBLAS64 API, as opposed to providing **a** CBLAS64 API, as previously we did not have a standardized CBLAS64 API.

epsilon-0:
@langou 
thanks a lot for the quick merge!

Issue #446, opened 2020-09-22
GGGLM leaves Y uninitialized for N=0


amigalemming:
In case of N=0, the correct result would be Y full of zeros.
Instead, the function returns early and leaves Y uninitialized.
The function documentation does not mention that this case is excluded or that Y must be initialized by the user.

thijssteel:
N=0, means we have an unconstrained problem. This seems like a reasonable use case.
I'll make a PR with a fix.

jschueller:
@langou looks like this is fixed now

langou:
Yes, indeed. Need to close the issue! Thanks for the reminder.

PR #463, opened 2020-11-06
initialize x and y to zero in case N = 0 in xggglm

thijssteel:
This PR solves #446.

When N = 0, the problem is unconstrained. The correct solution is then y = 0 (and x = 0).

amigelemming suggested initializing y to zero (instead of just returning) when N = 0. I've also included an initialization of x.

### DIFF
--- a/SRC/cggglm.f
+++ b/SRC/cggglm.f
@@ -271,8 +271,14 @@
 *
 *     Quick return if possible
 *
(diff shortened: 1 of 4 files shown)


langou:
Thanks @thijssteel and @amigalemming 
I think we lost some code coverage because our testing do not check for this (N=0). Sounds a good change to me nevertheless.

Issue #450, opened 2020-09-29
Reference-LAPACK/lapack vs Reference-LAPACK/lapack-release


JHenneberg:
Hey,

why are there two *.gits existing representing more or less the same? Do I miss something?

`Reference-LAPACK/lapack-release` is missing v.3.9.0 and all recent commits of course and `Reference-LAPACK/lapack` is missing the tags down from 3.6.0.
In the release version different branches are representing the tags.

If all tags down to 3.1.0 could be implemented as tags in `Reference-LAPACK/lapack` `Reference-LAPACK/lapack` could be deleted.

martin-frbg:
I suspect this is simply a result of converting from whatever CVS system was used prior to 3.7 and the move to github. What would be gained by deleting one of them and retroactively tagging releases in the other (which may be tedious or even technically impossible, depending on how source trees where handled in earlier days, and in any case would remove the "archival" quality of having the old releases exactly as they were created) ?

JHenneberg:
The benefit would be less confusion because of consistency. Now there is overlapping data. 

For example someone is searching for the newest release version of lapack and he finds the repoistory `Reference-LAPACK/lapack-release` first. Then he is taking 3.8.0 falsfully for the newest one. 
Also providing feedback by creating Issues is right now done in both repositories and additionaly on netlib.

If it should be kept for archival quality it should be marked as such with reference to the current SCM and issues should be deactivated to avoid confusion.

martin-frbg:
The README.me in the "lapack-release" repo already points here, and I do note that this setup appears to have worked for the past couple of years, however unusual it might look. (If anything a 3.9.0 branch might need to be added to the lapack-release repository, I suspect both maintainers had to move their attention to novel teaching conditions created by the covid pandemic soon after the 3.9.0 release).

ilayn:
They just forgot to add 3.9.0 to `lapack-release` probably. Seems like an SVN habit to have this as trunk and the other as the release copy.

But hardly ever, people use GitHub for LAPACK distribution anyways since most either goes to NetLib for download or use some other specialized implementation anyways.

JHenneberg:
>  I do note that this setup appears to have worked for the past couple of years, however unusual it might look. 

This does not mean it is the perfect setup or there can not be adjustments.

> (If anything a 3.9.0 branch might need to be added to the lapack-release repository, I suspect both maintainers had to move their attention to...

This could be avoided for example as simple it might look. 

[...]

PR #464, opened 2020-11-09
add pr and issue templates

thijssteel:
Inspired by issue #450 I thought it might be a good idea to start using a pretty nice feature: issue templates:

This PR adds a few templates that will be used when creating a new PR or issue on Github.
I think this might improve the quality of the reported issues and help us sort through them (by adding labels automatically).
When creating a new issue, users should now be offered a selection of templates.

Disclaimer, i'm not an expert with github, but have used this feature on other repositories. I think adding the templates to a .github folder does the trick, but some additional configuration might be required.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,20 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: 'Type: Bug'
+assignees: ''
+
+---
+
+# 1. Bug description
+
+A description of the bug: what is the expected behaviour? What is the actual behaviour?
+
+# 2. Reproduction
+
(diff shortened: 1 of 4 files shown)


martin-frbg:
IMHO the main effect is that you get a ton of redundant text and the "it does not work, fix it" types will not fill in the blanks anyway.

thijssteel:
Yeah, that does happen.

You have a lot more experience with open source projects so i'll leave it up to you. This took me about 10 minutes anyway.

The line i liked the most is "Are you willing to make a PR to fix this issue" as i think it could motivate newbies to contribute some small stuff. If you're worried about redundant text I could shorten the templates down and just have the template add some checkboxes.

julielangou:
I did not even know this was existing.
The "Are you willing to make a PR to fix this issue" question is indeed nice.. 
I have no strong opinion on this - @langou  what do you think?

langou (approved):

Issue #145, opened 2017-05-04
Out of range subscript calling dgemqr from TESTING/LIN/derrtsqr.f


mickpont:
There is a minor problem with the test routine DERRTSQR (file derrtsqr.f)
when it calls subroutine dgemqr(). DERRTSQR declares array TAU to
be of size NMAX (= 2), but DGEMQR examines elements TAU(2) and TAU(3)
 - the array TAU is named T in dgemwqr.f - before it decides whether it
needs to do a quick error return because of an incorrect input argument.

At the moment, because TAU(3) is undefined, in dgemqr.f variable NB
ends up as something random, which might affect the way the rest of the
error checking works.

The test routine DERRTSQR can be fixed easily by increasing the declared
size of array TAU from ( NMAX ) to ( NMAX+1 ), and adding a line
TAU(3) = 1 just after the initialization of TAU(1) and TAU(2).

However, DGEMQR itself oughtn't to examine elements T(2) and T(3) before
it knows that TSIZE is greater than 2. Therefore, dgemgq.f also ought
to be changed a bit. I think that one way to fix it is to replace the lines

      MB = INT( T( 2 ) )
      NB = INT( T( 3 ) )

by something like

      IF (TSIZE.GE.3) THEN
        MB = INT( T( 2 ) )
        NB = INT( T( 3 ) )
      ELSE
        MB = 1
        NB = 1
      END IF

but perhaps there are better ways.

The same issue affects the other precision variants of dgemqr.f
i.e. cgemqr.f, sgemqr.f and zgemqr.f, along with the corresponding
test routines in cerrtsqr.f, serrtsqr.f and zerrtsqr.f.

Mick Pont

langou:
Hi Mike. 

Thanks for the feedback. 

Here is what I propose. 

[...]

PR #468, opened 2020-12-26
fix #467

### LINKED ISSUE #467: Wrong declaration of CDUM in c/zgesdd
In c/zgesdd.f line 276 CDUM(1) is declared.
This is wrong because CDUM(4) is referenced in c/zungbr.f line 234 as A(2,2).

I suggest declaring CDUM(LDA*N) to avoid undefined behaviour.

langou:
Fix issue "Wrong declaration of CDUM in c/zgesdd" #467 from @vzecca by
passing A(1,1) instead of A(2,2) on lines 225 and 234 of SRC/[c/z]ungbr.f

### DIFF
--- a/SRC/cungbr.f
+++ b/SRC/cungbr.f
@@ -222,8 +222,8 @@
                CALL CUNGQR( M, N, K, A, LDA, TAU, WORK, -1, IINFO )
             ELSE
                IF( M.GT.1 ) THEN
-                  CALL CUNGQR( M-1, M-1, M-1, A( 1, 1 ), LDA, TAU, WORK,
-     $                         -1, IINFO )
+                  CALL CUNGQR( M-1, M-1, M-1, A, LDA, TAU, WORK, -1,
+     $                         IINFO )
                END IF
             END IF
          ELSE
@@ -231,8 +231,8 @@
                CALL CUNGLQ( M, N, K, A, LDA, TAU, WORK, -1, IINFO )
             ELSE
                IF( N.GT.1 ) THEN
-                  CALL CUNGLQ( N-1, N-1, N-1, A( 1, 1 ), LDA, TAU, WORK,
-     $                         -1, IINFO )
+                  CALL CUNGLQ( N-1, N-1, N-1, A, LDA, TAU, WORK, -1,
+     $                         IINFO )
                END IF
             END IF
          END IF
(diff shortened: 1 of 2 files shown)


zerothi:
While this fixes the issue, I think the intent of the original implementation was that the array passed to those calls (`[cz]ungbr`) are sub-matrices.  
To retain that I would argue they shouldn't be changed, but I don't know if this is a violation?

The calls are still doing work-size queries, so I think we should pass `A` instead of `CDUM`?

langou:
Hi Nick. I think passing A in [C/Z]GESDD (when calling [C/Z]UNGBR  during the workspace query at line 356) would result in the same problem as passing CDUM. Indeed when the user calls [C/Z]GESDD for a workspace query, even if M>1, and N>1, I do not think we can expect the array A to have A(2,2) correctly allocated. Most users probably call with a "correct" A. But, for example, in this case, the user "C/Z]GESDD" calls [C/Z]UNGBR with CDUM(2,2) not correctly allocated, and I think, this is a correct usage during workspace query. So, all this to say, that I do not think passing A in [C/Z]GESDD would be a better fix. Cheers, Julien.

martin-frbg:
So what is/was wrong with my suggestion to correct the dimensioning of CDUM in the test driver and leave everything unchanged ?
A mere workspace query will not access the elements in any case.

langou:
to @martin-frbg: (1) I did not understand you made this suggestion, sorry, I did not understand anyone made this suggestion, (2)  yes, declaring CDUM(2,2) in [C/Z]GESDD would also do the trick. I agree. (3) I prefer to use CDUM(1) in the workspace query of [C/Z]UNGBR (which is what the PR does). 

This is a matter of preference. I do not have a strong opinion on this. So if someone has a strong opinion one way or another, I am happy to go either way.

I think we should not have [C/Z]UNGBR make a call using A(2,2) or CDUM(2,2) if this does not exist, even in a workspace query. So I think we should fix this one way or another. This is not critical. But it would be better, in my opinion, if this is fixed.

martin-frbg:
Ah right, sorry, I confused this with the earlier issue #145 that also got mentioned in #467 (and for which I already committed my supposed fix). I agree that the only "advantage" in keeping the (2,2) argument would be to make the function call look the same for workspace query and the actual computation. (Wonder if the argument should then be passed as simply "A" like in the other branch of the conditional ?)

langou:
You mean, instead of changing `A(2,2)` to `A(1,1)` in this PR, change `A(2,2)` to `A`. I see. Yes. Indeed. This could be nicer.

thijssteel (approved):

weslleyspereira (approved):

weslleyspereira:
Hi! I think all agree the changes @langou proposed are sufficient to solve the bug. So, I've just accepted it. :)

PR #470, opened 2021-01-08
fix osx builds

thijssteel:
Currently, the osx travis builds fail because of the way gcc10 is installed (brew update, upgrade) (see https://github.com/Homebrew/discussions/discussions/226). Instead of using homebrew, we could edit the osx image to `osx_image: xcode10.3`, which should include gcc10, avoiding the issue.

Disclaimer: I'm not an export on CI's, hope this works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,15 +18,11 @@ matrix:
   - os: linux
     env: CMAKE_BUILD_TYPE=Coverage
   - os: osx
+    osx_image: xcode10.3
     env: CMAKE_BUILD_TYPE=Release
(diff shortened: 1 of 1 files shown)


thijssteel:
If i'm right, the failure of the coverage is due to differences in the macos and linux compilation and coverage checks. Because the macos builds take longer, they are uploaded later and overwrite the linux ones.

Maybe we can just disable the macos builds?

jschueller (approved):

jschueller:
@langou good to go ?

PR #471, opened 2021-01-08
dgesdd: Handle norm nan value

### LINKED ISSUE #469: DGESDD with nan exits with: On entry to DLASCL parameter number  4 had an illegal value
in version 3.9.0 and master, dgesdd with a nan value calls exit(0) instead of returning an error code on a 3x3 diagonal matrix plus a nan value and with jobz=N
 (xref http://www.netlib.org/lapack/explore-html/d1/d7e/group__double_g_esing_gad8e0f1c83a78d3d4858eaaa88a1c5ab1.html)
it seems to happen with gcc>=9 (confirmed with 9.3 and 10.2, but not 7.5)

```
M=3
N=3
A = {1,NAN,0, 0,1,0, 0,0,1}
```

full C code to reproduce:
```
#include <stdlib.h>
#include <stdio.h>
#include <math.h>
/* DGESDD prototype */
extern void dgesdd( char* jobz, int* m, int* n, double* a,
                int* lda, double* s, double* u, int* ldu, double* vt, int* ldvt,
                double* work, int* lwork, int* iwork, int* info );
/* Auxiliary routines prototypes */
extern void print_matrix( char* desc, int m, int n, double* a, int lda );

/* Parameters */
#define M 3
#define N 3
#define LDA M
#define LDU M
#define LDVT N

/* Main program */
int main() {
        /* Locals */
        int m = M, n = N, lda = LDA, ldu = LDU, ldvt = LDVT, info, lwork;
        double wkopt;
        double* work;
        /* Local arrays */
   /* iwork dimension should be at least 8*min(m,n) */
   int iwork[8*N];
    double s[N], u[LDU*M], vt[LDVT*N];
    double a[LDA*N] = {1,NAN,0, 0,1,0, 0,0,1};
    /* Query and allocate the optimal workspace */
    lwork = -1;
    char jobz= 'N';// dont ask for vectors, only singular values
    dgesdd_( &jobz, &m, &n, a, &lda, s, u, &ldu, vt, &ldvt, &wkopt,
      &lwork, iwork, &info );
    lwork = (int)wkopt;
    printf( " lwork=%d\n",  lwork);
    work = (double*)malloc( lwork*sizeof(double) );
    /* Compute SVD */
    dgesdd_( &jobz, &m, &n, a, &lda, s, u, &ldu, vt, &ldvt, work,
      &lwork, iwork, &info );
    printf( " info=%d\n",  info);
    /* Check for convergence */
    if( info > 0 ) {
      printf( "The algorithm computing SVD failed to converge.\n" );
      exit( 1 );
    }
    /* Print singular values */
    print_matrix( "Singular values", 1, n, s, 1 );

    /* Free workspace */
    free( (void*)work );
    return 0;
} /* End of DGESDD Example */

/* Auxiliary routine: printing a matrix */
void print_matrix( char* desc, int m, int n, double* a, int lda ) {
        int i, j;
        printf( "\n %s\n", desc );
        for( i = 0; i < m; i++ ) {
                for( j = 0; j < n; j++ ) printf( " %6.2f", a[i+j*lda] );
                printf( "\n" );
        }
}

```

possible xrefs:
https://github.com/statsmodels/statsmodels/issues/3596
https://github.com/statsmodels/statsmodels/issues/5396
https://pystatsmodels.narkive.com/2IKxOFVo/valueerror-on-entry-to-dlascl-parameter-number-4-had-an-illegal-value
https://forge.scilab.org/index.php/p/arpack-ng/issues/641/
https://www.ljll.math.upmc.fr/pipermail/freefempp/2015-November/004288.html

jschueller:
Closes #469

### DIFF
--- a/SRC/cgesdd.f
+++ b/SRC/cgesdd.f
@@ -281,9 +281,9 @@
      $                   CUNGQR, CUNMBR, SBDSDC, SLASCL, XERBLA
 *     ..
 *     .. External Functions ..
-      LOGICAL            LSAME
+      LOGICAL            LSAME, SISNAN
       REAL               SLAMCH, CLANGE
-      EXTERNAL           LSAME, SLAMCH, CLANGE
+      EXTERNAL           LSAME, SLAMCH, CLANGE, SISNAN
 *     ..
 *     .. Intrinsic Functions ..
       INTRINSIC          INT, MAX, MIN, SQRT
@@ -647,6 +647,10 @@
 *     Scale A if max element outside range [SMLNUM,BIGNUM]
 *
       ANRM = CLANGE( 'M', M, N, A, LDA, DUM )
+      IF( SISNAN ( ANRM ) ) THEN
+          INFO = -4
+          RETURN
+      END IF
       ISCL = 0
       IF( ANRM.GT.ZERO .AND. ANRM.LT.SMLNUM ) THEN
          ISCL = 1
--- a/SRC/dgesdd.f
+++ b/SRC/dgesdd.f
@@ -267,9 +267,9 @@
      $                   XERBLA
 *     ..
 *     .. External Functions ..
-      LOGICAL            LSAME
+      LOGICAL            LSAME, DISNAN
       DOUBLE PRECISION   DLAMCH, DLANGE
-      EXTERNAL           DLAMCH, DLANGE, LSAME
+      EXTERNAL           DLAMCH, DLANGE, LSAME, DISNAN
 *     ..
 *     .. Intrinsic Functions ..
       INTRINSIC          INT, MAX, MIN, SQRT
@@ -599,6 +599,10 @@
 *     Scale A if max element outside range [SMLNUM,BIGNUM]
 *
       ANRM = DLANGE( 'M', M, N, A, LDA, DUM )
+      IF( DISNAN( ANRM ) ) THEN
+          INFO = -4
+          RETURN
+      END IF
       ISCL = 0
       IF( ANRM.GT.ZERO .AND. ANRM.LT.SMLNUM ) THEN
          ISCL = 1
--- a/SRC/sgesdd.f
+++ b/SRC/sgesdd.f
@@ -267,9 +267,9 @@
      $                   XERBLA
 *     ..
 *     .. External Functions ..
-      LOGICAL            LSAME
+      LOGICAL            LSAME, SISNAN
       REAL               SLAMCH, SLANGE
-      EXTERNAL           SLAMCH, SLANGE, LSAME
+      EXTERNAL           SLAMCH, SLANGE, LSAME, SISNAN
 *     ..
 *     .. Intrinsic Functions ..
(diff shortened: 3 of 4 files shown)


langou:
Thanks Julien. We probably want to propagate the fix to SGESDD, CGESDD, and ZGESDD then.

jschueller:
ok, done

thijssteel:
minor suggestion, return info = N+1 instead of 1, that lets users distinguish non-convergence from NAN entries. Otherwise, looks good.

langou:
Right, same as @thijssteel. I like the direct return if there is a NaN in the input matrix A. I am not sure about INFO=1 though.
I was planning to write in the comments: 
INFO = 1: NaN present in input matrix A or 
INFO > 0:  DBDSDC did not converge, updating process failed.

But the overlap of INFO > 0 and INFO =1 was not great. (Though I could live with it.)

What about INFO=-4 ? That would be for invalid input A.

That would be consistent with INFO = -2 for example when the input M is invalid. Or INFO = -3 when the input N is invalid, etc.

I do not think we return INFO = -4 (invalid input) for an invalid array yet, but I think we can start on this. 

So: INFO = -4 ?

jschueller:
yes!

langou:
So now, codecov is not happy because we do not have a test to check that INFO=-4 is triggered correctly. Oh well. So, during the tester, we should do a 4x4 matrix (for example), put a NaN in it, send it to xGESDD and make sure xGESDD returns INFO=-4. So, maybe we put this (adding a check for INFO=-4 to xGESDD) on our TODO list and accept this Pull Request from @jschueller. BTW: thanks @jschueller

langou:
Of note, Jim Demmel made a nice presentation a year ago on "Correctness of Floating-Point Programs: Exception Handling and Reproducibility". The slides are https://people.eecs.berkeley.edu/~demmel/ExceptionHandling_Reproducibility_Feb20.pptx
Slides 18-21 are the most relevant to this discussion. And I think @jschueller patch is more or less what is suggested on slide 21

martin-frbg:
This has now caused a regression in NumPy by way of OpenBLAS - https://github.com/numpy/numpy/issues/18914 so there appear to be some codebases who rely on the previous behaviour of silently propagating the NaN

jschueller:
it should probably added to the doc then

langou:
Hi all, 

> it should probably added to the doc then

Yes, when we revised the LUG (LAPACK Users' Guide) we can add that NaN are considered illegal input value and some LAPACK routines will return with "INFO = -i, the i-th argument had an illegal value." when a NaN is present in the matrix/vector.

Right now the INFO description reads as:
```fortran
[out] INFO
          INFO is INTEGER
          = 0:  successful exit.
          < 0:  if INFO = -i, the i-th argument had an illegal value.
          > 0:  DBDSDC did not converge, updating process failed.
```
And I think this is fine. 

If someone has a better suggestion, please send it this way.

PR #472, opened 2021-01-08
CMake: Bump minimum version to 3.2

jschueller:
Newer cmake issues a warning when required version is < 3.
Version 3.2 should be old enough (eg ubuntu 16.04 has version 3.5).


jschueller:
good to go @langou ?

langou:
Thanks for pinging me on this. Yes, this sounds great to me. Thanks.

PR #478, opened 2021-01-27
improve exceptional shift in xlahqr

thijssteel:
While working on the exceptional shift strategy in QZ, i noticed something weird about the exceptional shifts in `xlahqr`.

In that routine, an exceptional shift is used when `ITS` is either 10 or 20. But `ITS` is the iteration counter! So we will be using an exceptional shift whether we get convergence or not. And even worse, if we get stagnation after those first 20 iterations, no further exceptional shifts will be used.

This PR fixes that by counting the number of iterations since the last deflation and basing the exceptional shifts based on that. I also went ahead and lowered the number of required iterations to 6, with an initial requirement of 8 to account for the typical slow startup of the QR algorithm.

### DIFF
--- a/SRC/clahqr.f
+++ b/SRC/clahqr.f
@@ -194,6 +194,7 @@
 *  =====================================================================
       SUBROUTINE CLAHQR( WANTT, WANTZ, N, ILO, IHI, H, LDH, W, ILOZ,
      $                   IHIZ, Z, LDZ, INFO )
+      IMPLICIT NONE
 *
 *  -- LAPACK auxiliary routine (version 3.7.0) --
 *  -- LAPACK is a software package provided by Univ. of Tennessee,    --
@@ -317,9 +318,9 @@
 *
       ITMAX = 30 * MAX( 10, NH )
 *
-*     KDFL counts the number of iterations since a deflation
+*     KDEFL counts the number of iterations since a deflation
 *
-      KDFL = -2
+      KDEFL = -2
 *
 *     The main loop begins here. I is the loop index and decreases from
 *     IHI to ILO in steps of 1. Each iteration of the loop works
@@ -564,7 +565,7 @@
 *
       W( I ) = H( I, I )
 *     reset deflation counter
-      KDFL = 0
+      KDEFL = 0
 *
 *     return to start of the main loop with new value of I.
 *
--- a/SRC/clahqr.f
+++ b/SRC/clahqr.f
@@ -218,6 +218,8 @@
       PARAMETER          ( RZERO = 0.0e0, RONE = 1.0e0, HALF = 0.5e0 )
       REAL               DAT1
       PARAMETER          ( DAT1 = 3.0e0 / 4.0e0 )
+      INTEGER            KEXSH
+      PARAMETER          ( KEXSH = 6 )
 *     ..
 *     .. Local Scalars ..
       COMPLEX            CDUM, H11, H11S, H22, SC, SUM, T, T1, TEMP, U,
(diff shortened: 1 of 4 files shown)


SRC/clahqr.f
```diff
@@ -315,6 +317,10 @@ SUBROUTINE CLAHQR( WANTT, WANTZ, N, ILO, IHI, H, LDH, W, ILOZ,
 *
       ITMAX = 30 * MAX( 10, NH )
 *
+*     KDFL counts the number of iterations since a deflation
+*
+      KDFL = -2
```
weslleyspereira, line 322:
```suggestion
      KDEFL = -2
```

weslleyspereira (changes requested):
Hi @thijssteel. One more typo, I think. Can you please replace `KDFL` by `KDEFL` in the files? Thanks!

weslleyspereira:
Hi @thijssteel. I noticed a strange behavior after this specific merge. If I don't use your new code, all tests pass using both the Makefile and CMake building ways. When using this commit, I observe:

1. CMake runs ok. All tests pass.
2. Several tests fail when I use the Makefile with `make.inc.example` from the master branch.

How to reproduce:
- git checkout 643f6dbc
- cp make.inc.example make.inc
- make
- Watch `TESTING/snep.out` and `TESTING/sed.out`

The errors disappear if I remove this specific PR.

In case you need: Ubuntu 18.04.5 LTS kernel 5.4.0-65-generic

thijssteel:
Damn, i can't seem to reproduce.

I also really can't see any programming errors.
Can you try if setting `KEXSH` to 10 and initializing `KDEFL` to 0 on line 323 fixes it. That should make it functionally equivalent to the old implementation.

thijssteel:
oh, i'm an idiot.
If KDEFL starts at -2, then it'll be in the second iteration. So `MOD(KDEFL,2*KEXSH).EQ.0` will be true. We should probably do the change i suggested above.

weslleyspereira:
Ok! I am testing what you suggested. I will tell you in a minute

weslleyspereira:
Unfortunately no.. I still have the problem:

```
			-->   LAPACK TESTING SUMMARY  <--
		Processing LAPACK Testing output found in the TESTING directory
SUMMARY             	nb test run 	numerical error   	other error  
================   	===========	=================	================  
REAL             	1294343		1508	(0.117%)	251	(0.019%)	
DOUBLE PRECISION	1316967		0	(0.000%)	0	(0.000%)	
COMPLEX          	776316		0	(0.000%)	0	(0.000%)	
COMPLEX16         	777128		0	(0.000%)	0	(0.000%)	

--> ALL PRECISIONS	4164754		1508	(0.036%)	251	(0.006%)
```

Since the CMake build is working, and you could not reproduce the problem, I think we can wait and see if someone else observes the same problem.

thijssteel:
found the mistake: https://github.com/Reference-LAPACK/lapack/pull/478/files#diff-82a702bd73aa9d0998122eda253e4782ba2e38223d4536ccae03305b040e2909L384

accidently removed the `else` on line 384.

thijssteel:
If you confirm that's the issue, its probably fine if you just commit the fix to master.

I think julien's going to have to retract some of his 'great work' statements about me. 😅

weslleyspereira:
Hehehe. This was difficult to see since the change was on the left side of the comparison and we mainly look at the right (new) sidse. Luckly we found it fast :) I will apply these corrections to test.

weslleyspereira:
Now it's perfect! I'll commit the changes. Thanks!!

```
			-->   LAPACK TESTING SUMMARY  <--
		Processing LAPACK Testing output found in the TESTING directory
SUMMARY             	nb test run 	numerical error   	other error  
================   	===========	=================	================  
REAL             	1316145		0	(0.000%)	0	(0.000%)	
DOUBLE PRECISION	1316967		0	(0.000%)	0	(0.000%)	
COMPLEX          	776316		0	(0.000%)	0	(0.000%)	
COMPLEX16         	777128		0	(0.000%)	0	(0.000%)	

--> ALL PRECISIONS	4186556		0	(0.000%)	0	(0.000%)
```

PR #489, opened 2021-02-13
lapack_testing.py: python2 explicit in hashbang

christoph-conrads:
There is no `python` executable in Ubuntu anymore. From Debian package [`python-is-python2`](https://packages.debian.org/bullseye/python-is-python2) description:

>  In Ubuntu, all python packages use explicit python3 or python2 interpreter and do not use unversioned /usr/bin/python at all. Some third-party code may still be python2 based, yet may use /usr/bin/python.
> 
> This is a convenience package which ships a symlink to point /usr/bin/python interpreter at the current default python2. It may improve compatibility with obsolete 3rd-party software, whilst breaking some modern software.
> 
> This package will be installed upon upgrades to Debian 11 (bullseye) and Ubuntu 20.04, if the DEPRECATED python2 was installed.


martin-frbg:
wouldn't this be even more likely to break on systems that have only python3 installed, with a symlink to plain "python" ? AFAIK python2 itself is basically abandoned now.

langou:
Here is what I read on my Mac when I type "python2"

`% python2

WARNING: Python 2.7 is not recommended. 
This version is included in macOS for compatibility with legacy software. 
Future versions of macOS will not include Python 2.7. 
Instead, it is recommended that you transition to using 'python3' from within Terminal.

Python 2.7.16 (default, Jun  5 2020, 22:59:21) 
[GCC 4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.29.20) (-macos10.15-objc- on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 
`

martin-frbg:
https://www.python.org/doc/sunset-python-2/

langou:
We should probably either leave `python` or have `python3`, but not an explicit `python2`. After reading a little bit about this, I am in favor of an explicit `python3`. Comments, opinions welcome.

ilayn:
Python 2 is decommissioned and almost no OS will provide in the near future hence this will become unusable very soon if `python2` is used. I don't think python version is important here anyways since it is a quite straightforward task. The python file can be improved slightly to make it less hardcoded let me know if I can be of use.

weslleyspereira:
What about leaving `/usr/bin/env python` and adding a local symlink to the default Python.
Suggestion:

```sh
env_python="$(which python)"
if [[ "$env_python" == "" ]]; then env_python="$(which python3)"; fi
if [[ "$env_python" == "" ]]; then env_python="$(which python2)"; fi
ln -s "$env_python" python
```

This could avoid problems to old systems that don't use python3. Altough I don't know where to trigger such a script.

christoph-conrads:
> This could avoid problems to old systems that don't use python3.

Python 3.0 was released in 2008 and Python 3.5 already reached its end of life in September 2020, eight months after Python 2. CentOS 7 is distribution with the oldest packages that I am actively using and even this distro has a Python3 interpreter in its stable package repositories.

christoph-conrads:
> Python 2 is decommissioned and almost no OS will provide in the near future hence this will become unusable very soon if `python2` is used.

Changing the hash bang to `python2` is the easiest change but converting to Python3 is the future-proof approach (I did not check if the script is compatible with Python3).

langou:
My opinion: (1) in favor of writing `python3` as opposed to `python`. (2) Quite against `python2`. (3) fine with `python`, preference for `python3`.

Quick survey:
`Thumbs up` for `python3`
`Hooray` for `python` (current state)
`Heart` for `python2`

langou:
Let us move to python3 then. @christoph-conrads : can you modify your PR from python2 to python3? I'll merge right after. Another solution would be to close this PR, and then @weslleyspereira or me can submit a new one, but might be better if we follow up on this one, and merge from this one. Either way is fine though. Let us know @christoph-conrads . Cheers, J.

christoph-conrads:
> @christoph-conrads : can you modify your PR from python2 to python3?

Yes, it will be pushed later today. The changes will include
* the hashbang update,
* removal of the second line with the coding information (Python3 code is always UTF-8 encoded), and
* removal of the `__future__` import.

langou:
Thanks @christoph-conrads. Looks great to me.

PR #491, opened 2021-02-14
xUNHR_COL01: pass required complex-valued argument

christoph-conrads:
This issue was detected by gfortran 10.2 on Fedora 33 (amd64).

### DIFF
--- a/TESTING/LIN/cunhr_col01.f
+++ b/TESTING/LIN/cunhr_col01.f
@@ -313,7 +313,7 @@
 *     TEST 2
 *     Compute |I - (Q**H)*Q| / ( eps * m ) and store in RESULT(2)
 *
-      CALL CLASET( 'Full', M, M, ZERO, CONE, R, M )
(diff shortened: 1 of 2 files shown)


thijssteel:
Out of interest, what compiler flags do you use to get those warnings?

I've made that same mistake quite a few times.

christoph-conrads:
> Out of interest, what compiler flags do you use to get those warnings?

This is a GCC 10 feature (hence, the mention of the compiler version).

[GCC 10 Release Series. Changes, New Features, and Fixes.](https://gcc.gnu.org/gcc-10/changes.html)
> Mismatches between actual and dummy argument lists in a single file are now rejected with an error. Use the new option `-fallow-argument-mismatch` to turn these errors into warnings; this option is implied with `-std=legacy`. `-Wargument-mismatch` has been removed.

weslleyspereira:
Great! Thanks for the fix @christoph-conrads !

PR #492, opened 2021-02-18
Fix segfault xeigtstz 335

### LINKED ISSUE #335: Segfault in ./EIG/xeigtstz
Hi,

this is with Lapack 3.8.0.

A segfault occurs in xeigtstz with lapack 3.8.0 and recent gfortran:

./EIG/xeigtstz < nep.in > znep.out 2>&1
/bin/sh: Zeile 1:  9026 Speicherzugriffsfehler  (Speicherabzug geschrieben) ./EIG/xeigtstz < nep.in > znep.out 2>&1

This goes away when the stack limit is raised from the default of 8192 kb to 81920 kb (the
first value I tried).

weslleyspereira:
Closes #335.

This PR adds recursive flags to the default build systems, and fix segfault in EIG tests.


weslleyspereira:
- I managed to remove the `recursive` flags from `TESTING/EIG/Makefile` using https://github.com/Reference-LAPACK/lapack/issues/335#issuecomment-780863336
- However, I spent some time trying to remove the `recursive` flags from the CMake build but couldn't make it right. This part is still missing.

h-vetinari:
I think there's a small regex-typo in https://github.com/Reference-LAPACK/lapack/issues/335#issuecomment-780863336. Should be
```
string(REGEX REPLACE "-[fM]?recursive" "" FFLAGS ${FFLAGS})
```
to cover all of `-recursive`, `-frecursive`, `-Mrecursive`.

thijssteel:
I see this as a bit of a bandaid and might cause problems later when we inevitably forget why that special flag was there.

I think we should rewrite the eigenvalue tests so that they allocate their workspace explicitly instead of using static allocations. If we don't do that now, perhaps we should add a note about it for later.

weslleyspereira:
> I think there's a small regex-typo in [#335 (comment)](https://github.com/Reference-LAPACK/lapack/issues/335#issuecomment-780863336). Should be
> 
> ```
> string(REGEX REPLACE "-[fM]?recursive" "" FFLAGS ${FFLAGS})
> ```
> 
> to cover all of `-recursive`, `-frecursive`, `-Mrecursive`.

Thanks! Do you know how to do it just for the target `zchkee.o`?

> I see this as a bit of a bandaid and might cause problems later when we inevitably forget why that special flag was there.
> 
> I think we should rewrite the eigenvalue tests so that they allocate their workspace explicitly instead of using static allocations. If we don't do that now, perhaps we should add a note about it for later.
> 
> Looks like @christoph-conrads also mentioned this [#335 (comment)](https://github.com/Reference-LAPACK/lapack/issues/335#issuecomment-780757490)
> 
> ah, but langou mentioned that bandaids are fine for now, nevermind.

Yes, the idea is to avoid spending time here as @langou explained. But I agree we should leave a note about it. I will do it in this PR.

martin-frbg:
I'd just put 
```
string(REGEX REPLACE "-[fM]?recursive" "" CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS}")
```
somewhere near the top of TESTING/EIG/CMakeLists.txt (and as noted elsewhere, this simple fix fails when you try to compile with OpenMP)

weslleyspereira:
> I'd just put
> 
> ```
> string(REGEX REPLACE "-[fM]?recursive" "" CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS}")
> ```
> 
> somewhere near the top of TESTING/EIG/CMakeLists.txt (and as noted elsewhere, this simple fix fails when you try to compile with OpenMP)

Yes... I ran into this problem and couldn't solve it. I am starting to think how much work would be necessary to apply the solution from https://github.com/Reference-LAPACK/lapack/issues/335#issuecomment-780757490

thijssteel:
probably not that hard, it's just 4 files and a couple of static matrices that need to be changed to allocatable (and then also the allocation and deallocation)

weslleyspereira:
This last commit allocates the matrices dynamically.

- Tests with `nep.in` all pass, with both Makefile and CMake build systems;
```
NEP: Testing Nonsymmetric Eigenvalue Problem routines
./EIG/xeigtstz < nep.in > znep.out 2>&1
SEP: Testing Symmetric Eigenvalue Problem routines
./EIG/xeigtstz < sep.in > zsep.out 2>&1
```
- Tests with `sep.in` take forever in my computer, and I couldn't see the results yet. I am putting here to get your feedback, and to see if Travis will have problems with those.

martin-frbg:
My current impression is that `xeigtstz <sep.in` appears to be stuck in ZLARGV, while the other xeigtst? passed reasonably quickly with the same input.

thijssteel:
Looks like the dimensions might be wrong

`A(NMAX,NMAX)` -> `A( NMAX*NMAX, NEED )`

If that couses some out of bounds issue, that could cause some NAN values leading to the infinite scaling issue

weslleyspereira:
That's true! I'm checking this right now... it gets stuck at
https://github.com/Reference-LAPACK/lapack/blob/master/SRC/zlargv.f#218
with `SCALE = 0`

weslleyspereira:
Yes... Sorry! My bad :|

thijssteel:
`SCALE = 0` get stuck ??? I don't see how that's possible, because then `G` should be zero.

martin-frbg:
Cute. So there is another flavor of the infinite scaling issue that is not (yet ?) taken care of by the evil "count to 20" hack. Maybe scale itself is NaN already, so won't compare to anything ?

weslleyspereira:
> `SCALE = 0` get stuck ??? I don't see how that's possible, because then `G` should be zero.

Well... Maybe `SCALE` starts with NaN or Inf at https://github.com/Reference-LAPACK/lapack/blob/master/SRC/zlargv.f#L194
Then line 203 may do the job.

weslleyspereira:
Thanks for the help! I think I abused the Copy&Paste method in the last commit. Now the tests pass. :)

langou:
If someone has an opinion on how ALLOCATE() should be coded in LAPACK (syntax, etc.), now is the good time to let it known.

My opinion: I do not have an opinion. (Yet!) @weslleyspereira introduces ALLOCATE() in TESTING/EIG There already was ALLOCATE() in TESTING/LIN. The style between both is a tat different. I am fine with different styles. We can homogenize, choose, decide later. At this point, I am glad that this 15-year-old bug is gone. Good job @weslleyspereira. I like the fix. Seems what we should have done from the get-go. For some reason, I was reluctant to use ALLOCATE(). (Thanks to @thijssteel to push a little on us using the ALLOCATE().)

thijssteel:
The massive workspace requirements are also present in the other eigenvalue tests, its just that DOUBLE COMPLEX takes up twice the memory as DOUBLE PRECISION / COMPLEX. I suggest changing those too for good measure.

Nice fix

weslleyspereira:
> The massive workspace requirements are also present in the other eigenvalue tests, its just that DOUBLE COMPLEX takes up twice the memory as DOUBLE PRECISION / COMPLEX. I suggest changing those too for good measure.

I agree. I will can that.

weslleyspereira:
I also tested with flag "-fopenmp". In this case, we have:
```
			-->   LAPACK TESTING SUMMARY  <--
		Processing LAPACK Testing output found in the TESTING directory
SUMMARY             	nb test run 	numerical error   	other error  
================   	===========	=================	================  
REAL             	1316145		0	(0.000%)	2	(0.000%)	
DOUBLE PRECISION	1316967		0	(0.000%)	2	(0.000%)	
COMPLEX          	776316		0	(0.000%)	4	(0.001%)	
COMPLEX16         	777128		0	(0.000%)	4	(0.001%)	

--> ALL PRECISIONS	4186556		0	(0.000%)	12	(0.000%)
```
[testing_results(Makefile).txt](https://github.com/Reference-LAPACK/lapack/files/6005869/testing_results.Makefile.txt)

I don't think these issues are the scope of #492, but it is worth registering.

martin-frbg:
Possibly related to machine precision or compiler issues ? I only get the 2 each for REAL and DOUBLE (from error exit tests in ?SYEVD_2STAGE) on i7-7500U

weslleyspereira:
> Possibly related to machine precision or compiler issues ? I only get the 2 each for REAL and DOUBLE (from error exit tests in ?SYEVD_2STAGE) on i7-7500U

I don't know exactly the cause, but the errors are
`*** XERBLA was called from SSYEVD_2STAGE with INFO =      8 instead of 10 ***`
`*** XERBLA was called from DSYEVD_2STAGE with INFO =      8 instead of 10 ***`
`*** XERBLA was called from CHEEVD_2STAGE with INFO =      8 instead of 10 ***`
`*** XERBLA was called from CHBEVD_2STAGE with INFO =     11 instead of 13 ***`
`*** XERBLA was called from CHBEVD_2STAGE with INFO =     11 instead of 15 ***`
`*** XERBLA was called from ZHEEVD_2STAGE with INFO =      8 instead of 10 ***`
`*** XERBLA was called from ZHBEVD_2STAGE with INFO =     11 instead of 13 ***`
`*** XERBLA was called from ZHBEVD_2STAGE with INFO =     11 instead of 15 ***`
and come from https://github.com/Reference-LAPACK/lapack/blob/master/TESTING/xerbla.f#L123 :

```Fortran
 9999 FORMAT( ' *** XERBLA was called from ', A, ' with INFO = ', I6,
     $      ' instead of ', I2, ' ***' )
```

What does it mean? Maybe concurrent threads acess the same INFO and that is why we see the problems only with `-fopenmp`.

martin-frbg:
Seems the test driver fed it a matrix that led to fewer non-converged off-diagonal elements in the output than expected. This might even be a side effect of thijs' recent work on improving convergence elsewhere in the code.

langou:
1. In LAPACK/TESTING/EIG we have a special XERBLA that is used to intercept the calls to XERBLA and instead of stopping the execution (as LAPACK/SRC/XERBLA does), this LAPACK/TESTING/EIG /XERBLA checks that the error INFO value is the one that is expected. See LAPACK/TESTING/EIG/XERBLA.f 

2. This LAPACK/TESTING/EIG/XERBLA.f uses some COMMON blocks. (Global variables.) So yes, this might not be thread-safe.

3. The CHBEVD_2STAGE collection of routines uses OPENMP in the HB 2 ST routines, for example in, chetrd_hb2st.F

So when you compile LAPACK with -fopenmp,  chetrd_hb2st.F will use multithreading. I do not know how this interacts with the COMMON blocks of LAPACK/TESTING/EIG/XERBLA.f

thijssteel:
would forcing the number of openmp threads to 1 help?

martin-frbg:
Guess I was misled by  (in ssyevd_2stage.f)
```
        INFO is INTEGER
*>          = 0:  successful exit
*>          < 0:  if INFO = -i, the i-th argument had an illegal value
*>          > 0:  if INFO = i and JOBZ = 'N', then the algorithm failed
*>                to converge; i off-diagonal elements of an intermediate
*>                tridiagonal form did not converge to zero;
*>                if INFO = i and JOBZ = 'V', then the algorithm failed
*>                to compute an eigenvalue while working on the submatrix
*>                lying in rows and columns INFO/(N+1) through
*>                mod(INFO,N+1).
```
In any case, I guess one of the threads may be raising an error prematurely, for the data that it sees - but I do not see the ?sb2st
access INFO in or after its openmp-parallel region.

martin-frbg:
Yep, running the test with OMP_NUM_THREADS=1 seems to "help".

thijssteel:
How about a nice:

`#if defined(_OPENMP) CALL OMP_SET_NUM_THREADS(1)` in those specific tests so we can still test the multithreading in the actual routine?

maybe reset the value afterwards

weslleyspereira:
> Seems the test driver fed it a matrix that led to fewer non-converged off-diagonal elements in the output than expected. This might even be a side effect of thijs' recent work on improving convergence elsewhere in the code.
> Update: I get the same failure messages after reverting your PR (but still none for the complex cases)

Based on that, I would say we, maybe, open a new issue to deal with thread safety. I think this PR achieved its goals.

TESTING/EIG/cchkee.F
```diff
@@ -1846,8 +1869,16 @@ PROGRAM CCHKEE
          CALL ALAREQ( C3, NTYPES, DOTYPE, MAXTYP, NIN, NOUT )
          CALL XLAENV( 1, 1 )
          CALL XLAENV( 9, 25 )
-         IF( TSTERR )
-     $      CALL CERRST( 'CST', NOUT )
+         IF( TSTERR ) THEN
```
langou, line 1872: Yes, that's it. Something like this and somewhere in the code like this. Good job @weslleyspereira. Thanks @thijssteel

weslleyspereira:
Actually... @langou pushed me forward to try @thijssteel's idea https://github.com/Reference-LAPACK/lapack/pull/492#issuecomment-782135368, and it was easier than I thought.

See 7c74f6ae776bbf07f4b7f089d80e803afc7d0f9f

The non-thread-safe subroutines are `xERRST`

PR #494, opened 2021-02-19
Add safe scaling xLASSQ routines from https://doi.org/10.1145/3061665

weslleyspereira:
Edward Anderson stated in his paper https://doi.org/10.1145/3061665 the following:

_The square root of a sum of squares is well known to be prone to overflow and underflow. Ad hoc scaling
of intermediate results, as has been done in numerical software such as the BLAS and LAPACK, mostly
avoids the problem, but it can still occur at extreme values in the range of representable numbers. More
careful scaling, as has been implemented in recent versions of the standard algorithms, may come at the
expense of performance or clarity. This work reimplements the vector 2-norm and the generation of Givens
rotations from the Level 1 BLAS to improve their performance and design. In addition, support for negative
increments is extended to the Level 1 BLAS operations on a single vector, and a comprehensive test suite
for all the Level 1 BLAS is included._

- This PR replaces the original xLASSQ with Edward's safe scaling xLASSQ codes.

- Since the codes were written in the Fortran90 standard, I removed

```CMake
# Tell CMake that our Fortran sources are written in fixed format.
set(CMAKE_Fortran_FORMAT FIXED)
```

from the `CMakeLists.txt`. The compiler now determines the Fortran layout by the file extension.


weslleyspereira:
I have just added Doxygen preambles to all new files. The `la_constants` module(s) depend(s) on #501.

langou (approved):

Issue #488, opened 2021-02-13
Build systems


christoph-conrads:
As of today, there are three different build systems in LAPACK:
* Makefiles-only,
* CMake, and
* Meson.

The Meson build is incomplete and unmaintained since its introduction in January 2019. The CMake and the Makefile-only build are not identical: The `-frecursive` flag is missing. Having more than one build system has at least three effects:
* It increases the amount of work,
* the issues #276 and #335 do not occur when building with CMake but
* only the CMake-based builds may crash when LAPACK is called concurrently, see the [comment about the need for `-frecursive` in PR #486](https://github.com/Reference-LAPACK/lapack/pull/486#issuecomment-778252995).

Especially the last point is great because even if someone delivers an accurate problem report, it is impossible to retrace the issue when you built with Makefiles.

I strongly suggest to stick with **one** build system.

martin-frbg:
To add to the list, CMake builds currently (appear to) support the option to build only select parts of the library (REAL, DOUBLE, COMPLEX and/or DOUBLE COMPLEX), while Makefile-only builds do not. (I have changed the Makefiles in the copy we distribute with OpenBLAS and could produce a PR if desired - would need to rediff to leave out some changes not relevant here)

christoph-conrads:
> CMake builds currently (appear to) support the option to build only select parts of the library (REAL, DOUBLE, COMPLEX and/or DOUBLE COMPLEX)

The CMake build has several options, I list the most important ones for me here:
* building shared or static libraries
* enabled or disabling tests
* 64- or 32-bit indices
* optimized and debugging builds
* options for matrix generator library (TMG), usage of other BLAS libraries, BLAS++, LAPACK++...

thijssteel:
Good idea. Builds working on my machine with make but not on the ci with CMake has driven me crazy a few times.

Technically, i think the Makefile also supports a lot of those options (you can just edit your make.inc). If you want a build without the tests you can build with `make lib`. But crucially, building a shared library is not supported. Most people use LAPACK as shared library so that makes CMake the clear winner for me.

If we end up going for it, I suggest we add some extra build instructions to the readme. How to build and run the tests, how to add flags, especially the frecursive flag. Perhaps even some small note on how to use the shared library.

martin-frbg:
I suspect there will be many patches floating around already (e.g. in linux distributions that package lapack) to build shared libraries with gmake.

christoph-conrads:
[...]

PR #500, opened 2021-03-01
Try makefile on travis ci

weslleyspereira:
Add tests for the Makefile build system using `make.inc.example` on Travis CI. Tests performed on both Linux and Mac OS


langou:
Thanks @weslleyspereira. That's a really good idea.

weslleyspereira:
This PR should be ready for review. It mitigates the inconsistencies between the Makefile and CMake build systems reported in #488.

In this PR we have:

1. Travis CI working for both CMake and Makefile default configurations.
2. Compatible compilation flags in these two builds systems.

This PR does not modify the Meson build system.

langou:
I think we are ready to retire the Meson build system. Thumbs up to this post if you think we should retire the Meson build system. Thumbs down if you think we should keep the Meson build system.

PR #502, opened 2021-03-02
Fix 411 div by zero in ctgsja

### LINKED ISSUE #411: dead-lock in SLARTG called by CTGSJA

weslleyspereira:
Closes #411

This PR replaces `A1.NE.ZERO` with `GAMMA is finite` in the subroutines xTGSJA.

### DIFF
--- a/SRC/ctgsja.f
+++ b/SRC/ctgsja.f
@@ -409,12 +409,13 @@
       LOGICAL            INITQ, INITU, INITV, UPPER, WANTQ, WANTU, WANTV
       INTEGER            I, J, KCYCLE
       REAL               A1, A3, B1, B3, CSQ, CSU, CSV, ERROR, GAMMA,
-     $                   RWK, SSMIN
+     $                   RWK, SSMIN, SFMIN
       COMPLEX            A2, B2, SNQ, SNU, SNV
 *     ..
 *     .. External Functions ..
       LOGICAL            LSAME
-      EXTERNAL           LSAME
+      REAL               SLAMCH
+      EXTERNAL           LSAME, SLAMCH
 *     ..
 *     .. External Subroutines ..
       EXTERNAL           CCOPY, CLAGS2, CLAPLL, CLASET, CROT, CSSCAL,
@@ -464,6 +465,10 @@
          CALL XERBLA( 'CTGSJA', -INFO )
          RETURN
       END IF
+*
+*     Safe minimum
+*
+      SFMIN = SLAMCH( 'Safe minimum' )
 *
 *     Initialize U, V and Q, if necessary
 *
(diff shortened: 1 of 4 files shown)


langou (commented):
This was a workaround to avoid dividing by a small number so as to avoid infinite loops in xLARTG. However, independently, xLARTG was fixed from infinite loops in PR #493. As a consequence, Issue #411 is fixed (and closed) by #493 without the need for this PR #502. The workaround presented here might have some side effects that we do not understand. (If a number is between ZERO and SFMIN, the code in this PR does nothing, and this might not be an appropriate behavior.) So for all these reason, best is to not merge this PR.

weslleyspereira:
Using PR #493 and the test case from https://github.com/Reference-LAPACK/lapack/issues/411#issue-608776973, we obtain a NaN in the output although INFO=0:
```
 INFO =            0
 K =            0
 L =            6
 ALPHA:    2.73536585E-15   1.48088933E-27   1.57898426E-33   0.00000000       0.00000000       0.00000000    
 BETA:     1.00000000       1.00000000       1.00000000                  NaN   1.00000000       1.00000000
```

3. With the solution from this PR (#502), I obtain:
```
 INFO =            0
 K =            0
 L =            6
 ALPHA:    2.73536606E-15   1.48088913E-27   1.57898444E-33   0.00000000       0.00000000       0.00000000    
 BETA:     1.00000000       1.00000000       1.00000000       1.00000000       1.00000000       1.00000000
```

My last commit tries to improve the readability of my changes.

langou (approved):

PR #505, opened 2021-03-04
improve deflation criterion in xhgeqz

thijssteel:
This PR replaces the normwise deflation in xhgeqz with elementwise deflation. Split from my other PR at the request of @langou 
This is stricter than normwise deflation because the sum of the diagonals is always going to be smaller than the frobenius norm.
This also matches the criterion from my multishift QZ implementation.

Small note: This is less strict than the deflation in the QR algorithm, because there, both the smaller-than-diagonal criterion and the stricter criterion by Françoise Tisseur must be satisfied. This is a step in the right though.

### DIFF
--- a/SRC/chgeqz.f
+++ b/SRC/chgeqz.f
@@ -518,13 +518,17 @@
          IF( ILAST.EQ.ILO ) THEN
             GO TO 60
          ELSE
-            IF( ABS1( H( ILAST, ILAST-1 ) ).LE.ATOL ) THEN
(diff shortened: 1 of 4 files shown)


weslleyspereira (approved):
Thanks @thijssteel ! Looks good to me!

langou:
Sounds good to me. HGEQZ uses |Re|+|Im| while I think LAQZ0 uses sqrt( Re^2 + Im^2 ) for computing the magnitude of complex entries. Not a worry for me, just a comment in passing.

PR #506, opened 2021-03-04
Fix Incorrect INFO value from SLASQ2 and DLASQ2

### LINKED ISSUE #325: Incorrect INFO value from DLASQ2
This is a very minor issue.

The documentation for `DLASQ2` says that 
```
*>  ... if the i-th argument is an
*>             array and the j-entry had an illegal value, then
*>             INFO = -(i*100+j)
```
When the code checks elements of `Z` (the second parameter), `INFO` is set according to this except for the following case at line 187:
```
         IF( Z( 2 ).LT.ZERO .OR. Z( 3 ).LT.ZERO ) THEN
            INFO = -2
            CALL XERBLA( 'DLASQ2', 2 )
            RETURN
```
`INFO` should be `-202` and `-203` respectively.

julielangou:
When the code checks elements of Z (the second parameter), INFO should be set according to the following
"if the i-th argument is an array and the j-entry had an illegal value, then INFO = -(i*100+j)"

Updated 2-by-2 case - adjusted value and added Z(1) check.

Will close #325

### DIFF
--- a/SRC/dlasq2.f
+++ b/SRC/dlasq2.f
@@ -184,10 +184,18 @@
 *
 *        2-by-2 case.
 *
-         IF( Z( 2 ).LT.ZERO .OR. Z( 3 ).LT.ZERO ) THEN
-            INFO = -2
+         IF( Z( 1 ).LT.ZERO ) THEN
+            INFO = -201
             CALL XERBLA( 'DLASQ2', 2 )
             RETURN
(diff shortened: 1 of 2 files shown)


weslleyspereira (commented):
The bug fix seems ok to me. Thanks!

weslleyspereira (approved):

PR #510, opened 2021-03-11
Fix for bug #481

scr2016:
1) Fix for bug #481 (If compiled with FFLAGS = -O0 -frecursive -ggdb3 -fcheck=bounds,the TSQR tests fail - array index out of bounds).
2) Comment fixes.


langou (approved):

langou:
Good job @scr2016. Thanks. 
@weslleyspereira is looking in putting the -fcheck=bounds now.

PR #513, opened 2021-03-14
Add a sample configuration file for the nagfor compiler

martin-frbg:


langou (approved):

ThemosTsikas (commented):
My run with -O4 in FFLAGS (instead of -O2) does not show test errors (on Linux x86_64 with gcc 8.3.1). 

Themos Tsikas, NAG Ltd.

martin-frbg:
That's good to know - I am only on a trial licence of your product at the moment (mainly for adapting the OpenBLAS build files) and cannot test other platforms besides Linux/x86_64 with that.

Issue #440, opened 2020-09-06
Errors with -fcheck=all regarding string length


h-vetinari:
While trying to find all inconsistencies in SLASET/DLASET-calls in the context of #438 (cf. #318, #425, #429), I ran a [build](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=205734&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d) with `-fcheck=all` to see if I had missed any (non-obvious) callsites.

While I could verify that #438 works as planned, this uncovered some additional errors:

* `14 - CBLAS-xscblat2 (Failed)`:
  ```
  At line 157 of file $SRC_DIR/BLAS/SRC/sgemv.f
  Fortran runtime error: Actual string length is shorter than the declared one for dummy argument 'trans' (-525502495/1)
  ```
* `15 - CBLAS-xscblat3 (Failed)`:
  ```
  At line 188 of file $SRC_DIR/BLAS/SRC/sgemm.f
  Fortran runtime error: Actual string length is shorter than the declared one for dummy argument 'transa' (-1730907187/1)
  ```
* `17 - CBLAS-xdcblat2 (Failed)`:
  ```
  At line 157 of file $SRC_DIR/BLAS/SRC/dgemv.f
  Fortran runtime error: Actual string length is shorter than the declared one for dummy argument 'trans' (-1442584295/1)
  ```
* `18 - CBLAS-xdcblat3 (Failed)`:
  ```
  At line 188 of file $SRC_DIR/BLAS/SRC/dgemm.f
  Fortran runtime error: Actual string length is shorter than the declared one for dummy argument 'transb' (0/1)
  ```
* `20 - CBLAS-xccblat2 (Failed)`:
  ```
  At line 159 of file $SRC_DIR/BLAS/SRC/cgemv.f
  Fortran runtime error: Actual string length is shorter than the declared one for dummy argument 'trans' (-1401260171/1)
  ```
* `21 - CBLAS-xccblat3 (Failed)`:
  ```
  At line 188 of file $SRC_DIR/BLAS/SRC/cgemm.f
  Fortran runtime error: Actual string length is shorter than the declared one for dummy argument 'transa' (0/1)
  ```
* `23 - CBLAS-xzcblat2 (Failed)`:
  ```
  At line 188 of file $SRC_DIR/BLAS/SRC/zgemm.f
[...]

PR #521, opened 2021-03-18
cblas: mark cblas_xerbla and F77_xerbla symbols as weak

veprbl:
This should fix an issue reported by @bobrik in https://github.com/Reference-LAPACK/lapack/issues/440#issuecomment-792117758, although I discovered it was also appearing on x86_64-darwin. Thanks to @langou for the useful summary https://github.com/Reference-LAPACK/lapack/issues/440#issuecomment-795909116 .


weslleyspereira:
Hi! Thanks for working on that! Please see my comment here: https://github.com/Reference-LAPACK/lapack/issues/440#issuecomment-802328221

weslleyspereira (approved):

PR #531, opened 2021-04-10
 Use cmake builtin tool for "env" to ensure cross-platform compatibility 

### LINKED ISSUE #530: Turn on BLAS++, blaspp build error.
Turn on BLAS++, blaspp build error:
FAILED: blaspp-prefix/src/blaspp-stamp/blaspp-configure

Trun off BLAS++, lapackpp also build error:
FAILED: lapackpp-prefix/src/lapackpp-stamp/lapackpp-configure

WIN10X64+ifort.exe

martin-frbg:
fixes #530


weslleyspereira:
Thanks, @martin-frbg. Should we turn on BLASPP and LAPACKPP on the CI tools?

weslleyspereira (approved):

martin-frbg:
I don't really know - both would do a download from Mark Gates' bitbucket site,  which could cause spurious failures in case of network problems

weslleyspereira:
Hm... I see. In that case, I think it may not be a good idea.

h-vetinari:
The CI here doesn't run crazy often; even if a network issue appear, it can be restarted. I'd say better to test things that are expected to work with slight attendant inconvenience, than not at all.

PR #535, opened 2021-04-15
Corrections to claqr4.f and slaqr4.f code, and update xLAQR comments

eshpc:
**Description**

The LAPACK 3.9.1 changes to the xLAQR routines had some inconsistencies in the code. I've modified claqr4.f and slaqr4.f to match the changes in other routines. I've also updated comments in all the xLAQR0 and xLAQR4 routines to match the recent code changes.

**Checklist**

- [ ] The documententation has been updated

### DIFF
--- a/SRC/claqr0.f
+++ b/SRC/claqr0.f
@@ -555,7 +555,7 @@
 *
 *                 ==== Got NS/2 or fewer shifts? Use CLAQR4 or
 *                 .    CLAHQR on a trailing principal submatrix to
-*                 .    get more. (Since NS.LE.NSMAX.LE.(N+6)/9,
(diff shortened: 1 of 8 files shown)


thijssteel:
Thanks for catching these errors, i usually program in double precision, so the single precision gets less attention.

I hope these mistakes do not cause any problems for users.

The changes in `NSR` and `NSMAX` might result in slightly lower performance, but should not result in an error as `(N+6)/9 < (N-3)/6` for all valid `N` (luckily, `NTINY` was correctly updated).

The mistake in `KDU` is potentially problematic, as this calculated workspace may be too large for the matrix (since `NSMAX` was increased).

langou (approved):

PR #536, opened 2021-04-15
Change OMP_GET_NUM_THREADS to OMP_GET_MAX_THREADS in xCHKEE

eshpc:
**Description**

The xCHKEE programs were modified at LAPACK 3.9.1 to force OpenMP to use 1 thread when testing the routine error exits, and then switch back to the original desired thread count for subsequent tests. Use of OMP_GET_NUM_THREADS() is not correct in this context (it will return 1 as this code is not inside a parallel region), we should use OMP_GET_MAX_THREADS() instead.


weslleyspereira (approved):
Thanks!

PR #539, opened 2021-04-16
Copy recent changes from xUNGBR to xORGBR

eshpc:
**Description**

I suggest we copy the change made to xUNGBR to fix the issue in https://github.com/Reference-LAPACK/lapack/issues/467 to the real versions of these routines (xORGBR).


weslleyspereira (approved):
Seems good to me. Thanks!

PR #540, opened 2021-04-22
Fix error on LAPACKE_*tpmqrt_work for ROW MAJOR matrices

### LINKED ISSUE #404: LAPACKE_dtpmqrt_work wrong computation for row major layout
The transposition of input matrices to have them in column major layout is not performed correctly. 
Consider we want to apply Q from the left.
The array 'a' for example is a  k x n matrix. In the line 105, it transposed into k x m matrix! (Checking the dimension, line 61, is wrong as well.)
The 't' array is nb x k matrix with leading dimension at least k in row major layout. Lines 54, 71, and 104 should be corrected. The same with the array 'v', lines 55 and 103 should be corrected. 
The dimensions of the transposed arrays are wrong and need to be corrected.

weslleyspereira:
Fixes #404

**Description**
There are many bugs in `LAPACKE_*tpmqrt_work` routines for ROW MAJOR matrices, as pointed out by @haldaas in #404.

**Checklist**

- [x] Fix `LAPACKE_dtpmqrt_work`
- [x] Fix the remaining `LAPACKE_*tpmqrt_work` whenever my solution is validated for the double algorithm
- [ ] Fix similar issues in other methods (if someone identifies it)

### DIFF
--- a/LAPACKE/src/lapacke_dtpmqrt_work.c
+++ b/LAPACKE/src/lapacke_dtpmqrt_work.c
@@ -48,16 +48,24 @@ lapack_int LAPACKE_dtpmqrt_work( int matrix_layout, char side, char trans,
             info = info - 1;
         }
     } else if( matrix_layout == LAPACK_ROW_MAJOR ) {
-        lapack_int lda_t = MAX(1,k);
+        lapack_int nrowsA, ncolsA, nrowsV;
+        if      ( side == 'L' ) { nrowsA = k; ncolsA = n; nrowsV = m; }
+        else if ( side == 'R' ) { nrowsA = m; ncolsA = k; nrowsV = n; }
+        else {
+            info = -2;
+            LAPACKE_xerbla( "LAPACKE_dtpmqrt_work", info );
+            return info;
+        }
+        lapack_int lda_t = MAX(1,nrowsA);
         lapack_int ldb_t = MAX(1,m);
-        lapack_int ldt_t = MAX(1,ldt);
-        lapack_int ldv_t = MAX(1,ldv);
+        lapack_int ldt_t = MAX(1,nb);
+        lapack_int ldv_t = MAX(1,nrowsV);
         double* v_t = NULL;
         double* t_t = NULL;
         double* a_t = NULL;
         double* b_t = NULL;
         /* Check leading dimension(s) */
-        if( lda < m ) {
+        if ( lda < ncolsA ) {
             info = -14;
             LAPACKE_xerbla( "LAPACKE_dtpmqrt_work", info );
             return info;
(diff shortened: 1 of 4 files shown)


LAPACKE/src/lapacke_dtpmqrt_work.c
```diff
@@ -48,16 +48,24 @@ lapack_int LAPACKE_dtpmqrt_work( int matrix_layout, char side, char trans,
             info = info - 1;
         }
     } else if( matrix_layout == LAPACK_ROW_MAJOR ) {
-        lapack_int lda_t = MAX(1,k);
+        lapack_int nrowsA, ncolsA, nrowsV;
+        if      ( side == 'L' ) { nrowsA = k; ncolsA = n; nrowsV = m; }
```
angsch, line 52: to avoid case sensitivity `LAPACKE_lsame(side, 'l')`

LAPACKE/src/lapacke_dtpmqrt_work.c
```diff
@@ -99,19 +107,19 @@ lapack_int LAPACKE_dtpmqrt_work( int matrix_layout, char side, char trans,
...
                         &ldt_t, a_t, &lda_t, b_t, &ldb_t, work, &info );
         if( info < 0 ) {
             info = info - 1;
         }
         /* Transpose output matrices */
-        LAPACKE_dge_trans( LAPACK_COL_MAJOR, k, m, a_t, lda_t, a, lda );
-        LAPACKE_dge_trans( LAPACK_COL_MAJOR, m, n, b_t, ldb_t, b, ldb );
+        LAPACKE_dge_trans( LAPACK_COL_MAJOR, ncolsA, nrowsA, a_t, lda_t, a, lda );
+        LAPACKE_dge_trans( LAPACK_COL_MAJOR, n, m, b_t, ldb_t, b, ldb );
         /* Release memory and exit */
```
angsch, line 123: The subroutine flips the row and column count internally when the matrix layout is column major prior to the transposition. This should read
```
LAPACKE_dge_trans( LAPACK_COL_MAJOR, nrowsA, ncolsA, a_t, lda_t, a, lda );
LAPACKE_dge_trans( LAPACK_COL_MAJOR, m, n, b_t, ldb_t, b, ldb );
```

angsch (commented):
The fixes of the leading dimensions look good. I like your solution to abstract from `side` by setting `nrowsA`, `ncolsA` and `nrowsV`.

weslleyspereira:
Thanks a lot for reviewing this PR, @angsch! I applied all your fixes in this last commit.

I will now apply the changes to the other `LAPACKE_*tpmqrt_work`.

angsch (commented):
Looks good to me :+1:

The fix likely applies to`LAPACKE_?tprfb_work`, too. The dimensions of `A` depend on `side`. When `side = 'Left'`, `A` has minimum dimensions `(K, N)`. When `side = 'Right'`, `A` is `(M, K)`.  The row major interface seems to only support `side = 'Left'`:
https://github.com/Reference-LAPACK/lapack/blob/28f7e8309608b92aaec2e2556d4b25d758ccada9/LAPACKE/src/lapacke_ctprfb_work.c#L52-L53

Perhaps it make sense to move this to a separate issue.

langou (approved):

PR #541, opened 2021-04-22
Applies @mdarnold1's solution to fix #393

### LINKED ISSUE #393: cblas.h incompatible with similar libraries
cblas.h is incompatible with ATLAS clapack.h.  The latter also includes cblas.h  but the definitions are not currently compatible which can cause confusion on systems with both libraries installed.  One way they are incompatible is ATLAS declares function arguments like enum CBLAS_... but lapack cblas.h has typedef enum {...} CBLAS_... (lines 19-25).

One way to improve compatibility would be a subtle modification to lapack's cblas.h:
typedef enum CBLAS... {...} CBLAS_...
...
#define CBLAS_ORDER CBLAS_LAYOUT

weslleyspereira:
Fix #393

**Description**
@mdarnold1 identified `cblas.h` was incompatible with similar libraries, and proposed a solution on #393. This PR applies his solution.

### DIFF
--- a/CBLAS/include/cblas.h
+++ b/CBLAS/include/cblas.h
@@ -16,13 +16,13 @@ extern "C" {            /* Assume C declarations for C++ */
    #define CBLAS_INDEX int
 #endif
 
-typedef enum {CblasRowMajor=101, CblasColMajor=102} CBLAS_LAYOUT;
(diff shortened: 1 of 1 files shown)


langou (approved):

langou:
Thanks @weslleyspereira and @mdarnold1 . Looks great!

PR #543, opened 2021-04-23
Add Explicit Casts For Complex To Real Cases

Matthew-Badin:
This change adds explicit casts to many of the complex routines.  This is done because if the fortran is translated to a different language, the rules in that language for mixed types may not be the same.  This change aids compilation in the translated language.  Additionally, for complex to integer cases I left the change as complex to real as real to integer is handled by the translated language.


langou:
I did not review the full list of changes but I think "explicit casts" is a good idea.

langou (approved):

langou:
Thanks @Matthew-Badin !

Matthew-Badin:
Thank you =).

Issue #339, opened 2019-05-08
C -> Fortran ABI issues in CBLAS & LAPACKE


jabl:
Hello,

since it's inception CBLAS seems to have assumed it's Ok to pass a length=1 string to the Fortran BLAS without passing a hidden string length argument. And the newer LAPACKE seems to have followed suit.

While this has worked (apparently adequately?) for several decades, in light of increasingly sophisticated compiler optimizations, for instance interprocedural or link-time optimizations, such sins are unfortunately coming back to haunt us.

The R developers noticed something fishy in their LAPACK interface with recent versions of GFortran.  They don't use LAPACKE, but the issue is due to them also omitting the string length argument.  See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90329 and links to mailing list threads therein for details.

While some workarounds exist, they aren't really bulletproof, and the proper and robust fix is to change the interfaces so that they match between the caller and the callee.  The problem is that there is no "standard" Fortran ABI, and different compilers use different ways to pass the hidden string length information.  A couple of ways around this:

1) Use the Fortran 2003 ISO_C_BINDING feature to create a C ABI compatible wrapper routine that can then call the actual Fortran implementation. This would be robust, but the downside is that this approach requires a Fortran compiler with the ISO_C_BINDING feature (which is all(?) actively maintained compilers these days, but if you want to retain support for old compilers this could be a problem). Also there would be an extra layer of indirection, e.g. instead of "cblas_foo -> foo" it would be "cblas_foo -> cblas_icb_foo -> foo", although it should be possible for the compiler to inline the extra wrapper.

2) Use knowledge of compiler-specific argument passing conventions and macro magic. This is somewhat brittle, and requires fiddling with build machinery and macros, but doesn't need ISO_C_BINDING, and avoids the extra wrapper routines.

jabl:
Having thought about this issue for a while, there's also the option

3. Rewrite CBLAS and LAPACKE in Fortran using BIND(C). That would avoid the extra wrapper issue with option 1. and the fragility of option 2.

weslleyspereira:
#515 closed the CBLAS part. We still need to change `LAPACKE/include/lapack.h` to solve this issue.

eshpc:
[...]

PR #551, opened 2021-05-11
Fix out-of-bounds in xORBDB2 and xORBDB3

### LINKED ISSUE #549: xUNCSD2BY1 does not handle submatrices with empty rows properly
**Description**

The 2x1 CS decomposition decomposes a matrix `Q = [Q1; Q2]`, where `Q^* Q = I`. There are two possibilities for the CS decomposition `UΣV = Q1` in the special case where `Q2` has no rows:
* `U = Q1`, `Σ = I`, `V = I`,
* `U = I`, `Σ = I`, `V = Q1`.

That is, the only thing to do by the algorithm is copying the input matrix into U or V and set the other matrix to the identity matrix.

If
* one of the matrices `Q1` or `Q2` has no rows and
* the other matrix is **square**,

then SORCSD2BY1 performs out-of-bounds array accesses. I assume the same behavior occurs also for DORCSD2BY1, CUNCSD2BY1, and ZUNCSD2BY1. My expectation is that xORCSD2BY1 does not perform out-of-bounds accesses and returns one of the two possible decompositions shown above.

```fortran
      PROGRAM testCSD2BY1
      IMPLICIT NONE

      INTEGER            INFO, LDA, LDU1, LDU2, LDV, M, N, P, LWORK, I
      PARAMETER        ( LDA = 1, LDU1 = 5, LDU2 = 5, LDV = 5,
     $                   M = 0, P = 1, N = 1, LWORK = 1024)

      INTEGER            IWORK(M + N + P)
      REAL               THETA( N ), A( LDA, N ),
     $                   U1( LDU1, M ), U2( LDU2, P ), V( LDV, N ),
     $                   WORK( LWORK )

      EXTERNAL           SGGQRCS

      A = 0.0
      DO I = 1, MIN( P, N )
         A( I, I ) = 1.0E0
      END DO

      CALL SORCSD2BY1( 'Y', 'Y', 'Y', M + P, M, N, 
     $                 A( 1, 1 ), LDA, A( M + 1, 1 ), LDA,
     $                 THETA,
     $                 U1, LDU1, U2, LDU2, V, LDV,
     $                 WORK, LWORK, IWORK, INFO )
      END PROGRAM testCSD2BY1
```

Output:
```
$ make
gfortran  -frecursive -fcheck=all -Wextra -Wall -pedantic bug-csd2by1.f -llapack -L/tmp/build.lapack/lib
$ env LD_LIBRARY_PATH=/tmp/build.lapack/lib ./a.out
At line 318 of file /home/christoph/lapack/SRC/sorbdb2.f
Fortran runtime error: Index '2' of dimension 1 of array 'x21' above upper bound of 1

Error termination. Backtrace:
#0  0x7fb2529d0c3b in sorbdb2_
	at /home/christoph/lapack/SRC/sorbdb2.f:318
#1  0x7fb2529dd2ce in sorcsd2by1_
	at /home/christoph/lapack/SRC/sorcsd2by1.f:549
#2  0x4009cd in ???
#3  0x400a0b in ???
#4  0x7fb251655bf6 in ???
#5  0x400789 in ???
#6  0xffffffffffffffff in ???
```

**Checklist**

weslleyspereira:
Fix #549.

**Description**
In #549, we verified xORBDB2 and xORBDB3 use subarrays of size 0 starting at `X(N+1)`, where N is the size of X. These arrays are passed as arguments to other subroutines that never reference the arrays. Therefore, this is a false positive case of out-of-bounds. The code provokes no errors in practice, but some compilers (like gfortran) identify and return an `out-of-bounds error`. This PR tries to rewrite the code while keeping the check of bounds. __[edited]__

**Checklist**

- [x] Fix sORBDB2
- [x] Fix sORBDB3
- [x] Fix dORBDB2 and dORBDB3

### DIFF
--- a/SRC/cunbdb1.f
+++ b/SRC/cunbdb1.f
@@ -222,7 +222,7 @@
 *     .. Local Scalars ..
       REAL               C, S
       INTEGER            CHILDINFO, I, ILARF, IORBDB5, LLARF, LORBDB5,
-     $                   LWORKMIN, LWORKOPT, I1, I2, I3, I4
+     $                   LWORKMIN, LWORKOPT, I1, I2
       LOGICAL            LQUERY
 *     ..
 *     .. External Subroutines ..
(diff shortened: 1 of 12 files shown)


weslleyspereira:
The strategy was:

1. Remove `I=P` and `I=Q` from the loops.

2. Replace

```fortran
SLARFGP( 1, ALPHA, OUTOFBOUNDS_V, LDV, TAU )
```

by

```fortran
TAU = 1 - SIGN( 1, ALPHA )
ALPHA = ABS( ALPHA )
```

3. Remove calls of `SLARF` where the output argument was invalid.

christoph-conrads:
> I do not know what to do about it since I do not know whether the code relies on the P or not.

The computed beta values are used as `atan2` inputs. Without further review, it does rely on the beta values being positive.

> **Description**
> In #549, it was highlighted xORBDB2 and xORBDB3 had several out-of-bounds accesses. This PR tries to fix some of them.

There are no out-of-bounds accesses. Given an array `X` with valid indices `1, 2, ..., N`, the code references the subarray of length zero starting at X( N + 1)`.

> **Checklist**
> 
>     * [x]  Fix sORBDB2
> 
>     * [x]  Fix sORBDB3
> 
>     * [ ]  Fix dORBDB2 and dORBDB3

Add to that:
* [ ] SORBDB1 (for when the column count `Q` is the smallest dimension)
* [ ] The array accesses other do-loop in SORBDB{1,2,3} when the smaller submatrix has more than zero rows
* [ ] DORBDB{1,2,3}
* [ ] CUNBDB{1,2,3}
* [ ] ZUNBDB{1,2,3}

weslleyspereira:
I see... If one wants to fix all of them, we would need to rewrite a number of algorithms with no real improvement in practice. I couldn't think of a good way to continue using `-fcheck=bounds` and avoid the false positives. Any ideas?

christoph-conrads:
> I see... If one wants to fix all of them, we would need to rewrite a number of algorithms with no real improvement in practice. I couldn't think of a good way to continue using `-fcheck=bounds` and avoid the false positives. Any ideas?
> 
> * I updated the description of this PR.

Valgrind and address sanitizers work well for me. They also avoid issue #339 when calling LAPACK code from C with the function prototypes in [`LAPACKE/include/lapack.h`](https://github.com/Reference-LAPACK/lapack/blob/2dafa3d2756a7825c23a8c8456781561e36668ae/LAPACKE/include/lapack.h) (the header omits the hidden arguments of type `size_t` that contain the length of the strings passed to the callee; consequently C callers do not initialize these values).

Valgrind can be used with any executable: `valgrind --valgrind-arg1 --valgrind-arg2 -- my-program --my-arg`. Address sanitizers can be used by passing `-fsanitize=address` on the compiler and linker command line. It is also possible to use address sanitizers without re-building the code by pre-loading them (e.g., `env LD_PRELOAD=/path/to/libasan.so my-executable --my-args`) but I am not sure which problems can be detected this way. On some Linux distributions, the address sanitizer libraries are not part of the GCC package installed by the system's package manager (e.g., on CentOS); use of the sanitizer will then cause linker errors.

christoph-conrads:
> I see... If one wants to fix all of them, we would need to rewrite a number of algorithms with no real improvement in practice. I couldn't think of a good way to continue using `-fcheck=bounds` and avoid the false positives. Any ideas?

Just found in `sgeqr2.f`: Use `X( MIN(I+1, N) )`.

https://github.com/Reference-LAPACK/lapack/blob/2dafa3d2756a7825c23a8c8456781561e36668ae/SRC/sgeqr2.f#L177-L182

weslleyspereira:
Thanks, @christoph-conrads! That was easy to implement. I followed your suggestions and created a new commit. I am now applying the fixes to xORBDB1, and then I will look at the complex cases.

weslleyspereira:
Done! I hope I didn't miss anything.

christoph-conrads:
> Thanks, @christoph-conrads! That was easy to implement. I followed your suggestions and created a new commit. I am now applying the fixes to xORBDB1, and then I will look at the complex cases.

This was not a suggestion, I just responded to your request for ideas, and I think this is a bad idea.

These changes trade readability for `-fcheck=bounds` shutting up about alleged out-of-bounds accesses in the special case where the matrix at hand is not a submatrix of a bigger matrix. Valgrind and the address sanitizer can detect OOB access in this special case, too. In any other case, `-fcheck=bounds` will not help. Moreover, if there is any numerical mistake in those 250+ lines touched by this commit, none of these tools will help and given variable names like `I`, `I1`, `I2`, `I3`, and `I4`, the humans cannot help either.

langou (approved):

weslleyspereira:
I will drop this PR. I don't think this is a good idea anymore. I agree with the sentence:

> These changes trade readability for `-fcheck=bounds` shutting up about alleged out-of-bounds accesses in the special case where the matrix at hand is not a submatrix of a bigger matrix.

I think it would be good to have a routine in the CI to run the tests under Valgrind. Thanks @christoph-conrads!

PR #555, opened 2021-05-16
Correct XERBLA argument in SLASQ2

msk:
**Description**

`SLASQ2` calls `XERBLA` with `'DLASQ2'`, instead of `'SLASQ2'`, as the first argument.


martin-frbg (approved):
Approved as obvious (once you see it), good catch, thanks

PR #571, opened 2021-05-30
modify xGBT02 and xGET02 to use 1-norm of op(A) not A (no.2 from PR #562)

jip:
1. Comments improved, unified and fixed.
2. Code improved and fixed.

To unify with routines xGBT01, xGTT02, xLANGB, xTBT02, xTPT02, and xTRT02.

Some remarks:
1. It turned out that xLANGB, xLANGT, xLANTB, and xLANTP subroutines accept only a square matrix A, unlike xLANGE and xLANTR which allow A to be non-square.
2. xLANGB uses inline loop not SCSUM1/DZSUM1 for 1-norm and inf-norm computing.
3. xGET02 uses magnitude-based norms |y| unlike xGBT02 which uses taxicab-based norms |Re(y)| + |Im(y)|.

### DIFF
--- a/TESTING/EIG/cget02.f
+++ b/TESTING/EIG/cget02.f
@@ -28,9 +28,10 @@
 *> \verbatim
 *>
 *> CGET02 computes the residual for a solution of a system of linear
-*> equations  A*x = b  or  A'*x = b:
-*>    RESID = norm(B - A*X) / ( norm(A) * norm(X) * EPS ),
-*> where EPS is the machine epsilon.
+*> equations op(A)*X = B:
+*>    RESID = norm(B - op(A)*X) / ( norm(op(A)) * norm(X) * EPS ),
+*> where op(A) = A, A**T, or A**H, depending on TRANS, and EPS is the
+*> machine epsilon.
 *> \endverbatim
 *
 *  Arguments:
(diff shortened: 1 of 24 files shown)


langou:
> xGET02 uses magnitude-based norms |y| unlike xGBT02 which uses taxicab-based norms |Re(y)| + |Im(y)|.

Good point. You are correct.

The fact that we use two different measures is awkward and I do not know why this has been like that.

I wonder whether using xLANGB in xGBT02 would not (1) simplify the code of xGBT02, and (2) use magnitude-based norms |y|.

jip:
> I wonder whether using xLANGB in xGBT02 would not (1) simplify the code of xGBT02, and (2) use magnitude-based norms |y|.

It would, but this change will require to extend xLANGB interface to accept non-square matrices. And also to extend xLANGT, xLANTB, and xLANTP in the same way, to maintain consistency.

langou:
I see. Thanks @jip for explaining. 

Another idea is to have SCASUM (and DZASUM) that works with complex ABS( z ) instead of ABS(REAL( z )) + ABS(IMAG( z )). Having such SCASUM and DZASUM could actually be useful for a variety of reason. Then we would simply replace the calls to SCASUM (and DZASUM) in CGBT02 (and ZGBT02) by these new routines. 

So, well, I think that's great work @jip and we can leave it at that for now. And revisit as needed in the future.

langou (approved):

PR #572, opened 2021-06-03
Fix some typos in code and comments

jip:
These typos were discovered while working on the PR #571.


langou:
Thanks. I did `grep REAL z*f` in LAPACK/SRC/ and there are a few. We'll try to fix these and replace the REAL() with DBLE(). I am looking at your other changes. Looking good. Thanks.

langou (approved):

PR #581, opened 2021-06-09
fixed comments: specify which norm is used (no.1 from PR #562)

jip:
Despite the recommendation, no change has been made in xGBT02, xGET02, xGTT02, xTBT02, xTPT02, and xTRT02 for C,Z datatypes yet: as it was noted afterwards in PR #571, those routines are using mixed types of norm: 1-norm for op(A) and taxicab-based norm |Re(z)|+|Im(z)| for residual.


weslleyspereira (approved):
Nice job! Thanks.

langou (approved):

PR #582, opened 2021-06-14
Fix some typos in comments

jip:
These typos were overlooked while working on the PR #581. I hope they are the last.

### DIFF
--- a/TESTING/LIN/ctbt02.f
+++ b/TESTING/LIN/ctbt02.f
@@ -29,13 +29,11 @@
 *> \verbatim
 *>
 *> CTBT02 computes the residual for the computed solution to a
-*> triangular system of linear equations  A*x = b,  A**T *x = b,  or
-*> A**H *x = b  when A is a triangular band matrix.  Here A**T denotes
-*> the transpose of A, A**H denotes the conjugate transpose of A, and
-*> x and b are N by NRHS matrices.  The test ratio is the maximum over
-*> the number of right hand sides of
+*> triangular system of linear equations op(A)*X = B, when A is a
+*> triangular band matrix. The test ratio is the maximum over the
+*> number of right hand sides of
 *>    norm(b - op(A)*x) / ( norm(op(A)) * norm(x) * EPS ),
-*> where op(A) denotes A, A**T, or A**H, and EPS is the machine epsilon.
+*> where op(A) = A, A**T, or A**H, and EPS is the machine epsilon.
 *> \endverbatim
 *
 *  Arguments:
--- a/TESTING/LIN/ctbt02.f
+++ b/TESTING/LIN/ctbt02.f
@@ -32,7 +32,7 @@
 *> triangular system of linear equations op(A)*X = B, when A is a
 *> triangular band matrix. The test ratio is the maximum over the
 *> number of right hand sides of
-*>    norm(op(A)*x - b) / ( norm(op(A)) * norm(X) * EPS ),
+*>    norm(op(A)*X - B) / ( norm(op(A)) * norm(X) * EPS ),
 *> where op(A) = A, A**T, or A**H, and EPS is the machine epsilon.
 *> \endverbatim
 *
--- a/TESTING/LIN/ctbt02.f
+++ b/TESTING/LIN/ctbt02.f
@@ -32,7 +32,7 @@
 *> triangular system of linear equations op(A)*X = B, when A is a
 *> triangular band matrix. The test ratio is the maximum over the
 *> number of right hand sides of
-*>    norm(b - op(A)*x) / ( norm(op(A)) * norm(x) * EPS ),
+*>    norm(op(A)*x - b) / ( norm(op(A)) * norm(X) * EPS ),
 *> where op(A) = A, A**T, or A**H, and EPS is the machine epsilon.
 *> \endverbatim
 *
--- a/TESTING/LIN/ctbt02.f
+++ b/TESTING/LIN/ctbt02.f
@@ -140,7 +140,7 @@
 *> \verbatim
 *>          RESID is REAL
 *>          The maximum over the number of right hand sides of
-*>          norm(B - op(A)*X) / ( norm(op(A)) * norm(X) * EPS ).
+*>          norm(op(A)*x - b) / ( norm(op(A)) * norm(x) * EPS ).
 *> \endverbatim
 *
 *  Authors:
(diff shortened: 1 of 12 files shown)


TESTING/LIN/ctbt02.f
```diff
@@ -29,13 +29,11 @@
...
-*> A**H *x = b  when A is a triangular band matrix.  Here A**T denotes
-*> the transpose of A, A**H denotes the conjugate transpose of A, and
-*> x and b are N by NRHS matrices.  The test ratio is the maximum over
-*> the number of right hand sides of
-*>    norm(b - op(A)*x) / ( norm(op(A)) * norm(x) * EPS ),
-*> where op(A) denotes A, A**T, or A**H, and EPS is the machine epsilon.
+*> triangular system of linear equations op(A)*X = B, when A is a
+*> triangular band matrix. The test ratio is the maximum over the
+*> number of right hand sides of
+*>    norm(op(A)*X - B) / ( norm(op(A)) * norm(X) * EPS ),
```
weslleyspereira, line 35: I think this is still incorrect. The expression `norm(op(A)*X - B) / ( norm(op(A)) * norm(X) * EPS )` is a number, so it does not have right-hand sides. I suggest something like:

```
The test ratio is the maximum over
norm(b - op(A)*x) / ( norm(op(A)) * norm(x) * EPS )
for every column b of B,
where op(A) = A, A**T, or A**H, x is the solution vector, and EPS is the machine epsilon.
```

What do you think?

jip: Yes, this looks better. Two adjoining "where" sections (the first one - "for every column b of B" - is implicit, and the second one is explicit) could be merged:
```
The test ratio is the maximum over
  norm(b - op(A)*x) / ( norm(op(A)) * norm(x) * EPS ),
where op(A) = A, A**T, or A**H, b is the column of B, x is
the solution vector, and EPS is the machine epsilon.
```
but this is a matter of taste.

I believe this variant is incorrect, too. The original comment states "x and b are N by NRHS matrices". So will be the residual `R := b - op(A)*x`. The code computes taxicab-based norm `|| R(j) ||` for each j-th column R(j) from R. Then it takes a `max(j) || R(j) ||` which means a computing of kinda 1-norm, but based on taxicab-based norm `|Re(z)| + |Im(z)|`, not magnitude-based norm `|z|`. So, the phrase "the maximum over" is excessive.

And this variant may be improved further:
1. Since x and b are matrices not vectors here, capital letters X and B are used.
2. I'm using `norm(op(A)*X - B)` not `norm(B - op(A)*X)` since the first is the exact expression which is computed. I know they give identical results. I don't insist on this summands swapping.

So, the result may look like this:
```
The test ratio is set to
  norm(op(A)*X - B) / ( norm(op(A)) * norm(X) * EPS ),
where op(A) = A, A**T, or A**H, B is the N by NRHS matrix of
right-hand sides, X is the N by NRHS matrix of solution vectors, and
EPS is the machine epsilon.
```

When "kinda 1-norms" in both `norm(op(A)*X - B)` and `norm(X)` will be replaced by "true 1-norm", then the following statement could be added: "The norm used is the 1-norm.".

I'm ready to make a PR for any of variants discussed above.

weslleyspereira: > Yes, this looks better. Two adjoining "where" sections (the first one - "for every column b of B" - is implicit, and the second one is explicit) could be merged:
> 
> ```
> The test ratio is the maximum over
>   norm(b - op(A)*x) / ( norm(op(A)) * norm(x) * EPS ),
> where op(A) = A, A**T, or A**H, b is the column of B, x is
> the solution vector, and EPS is the machine epsilon.
> ```
> 
> but this is a matter of taste.

I would go with either this one or the version I proposed. Maybe you can modify it a bit:

```
The test ratio is the maximum over
  norm(b - op(A)*x) / ( ||op(A)||_1 * norm(x) * EPS ),
where op(A) = A, A**T, or A**H, b is the column of B, x is
the solution vector, and EPS is the machine epsilon.
```

As you pointed out, the `norm` operator is that "kinda 1-norm" for vectors of size N, in which `|Re(z)| + |Im(z)|` replaces the magnitude `|z|` for all complex scalars z. This is what the subroutine computes if I understand the algorithm correctly. 

> I believe this variant is incorrect, too. The original comment states "x and b are N by NRHS matrices". So will be the residual `R := b - op(A)*x`. The code computes taxicab-based norm `|| R(j) ||` for each j-th column R(j) from R. Then it takes a `max(j) || R(j) ||` which means in toto a computing of kinda 1-norm, but based on taxicab-based norm `|Re(z)| + |Im(z)|`, not magnitude-based norm `|z|`. So, the phrase "the maximum over" is excessive.
> 
> And this variant may be improved further:
> 
>     1. Since x and b are matrices not vectors here, capital letters X and B are used.
> 
>     2. I'm using `norm(op(A)*X - B)` not `norm(B - op(A)*X)` since the first is the exact expression which is computed. I know they give identical results. I don't insist on this summands swapping here.
> 
> 
> So, the result may look like this:
> 
> ```
> The test ratio is set to
>   norm(op(A)*X - B) / ( norm(op(A)) * norm(X) * EPS ),
> where op(A) = A, A**T, or A**H, B is the N by NRHS matrix of
> right-hand sides, X is the N by NRHS matrix of solution vectors, and
> EPS is the machine epsilon.
> ```

I don't agree with this proposal. The algorithm does not compute `norm(X)`, where "capital" X is the original input matrix. Maybe I am missing something, please correct me if I am wrong.

weslleyspereira:
I just noticed another thing that is not related to your changes, but maybe you can help me. Is there a bug here?

https://github.com/Reference-LAPACK/lapack/blob/2eeed4b25d89fb4bb76dfc39b481115d237fc464/TESTING/LIN/ctbt02.f#L231-L235

I think so. Suppose that the last column of `X` is full of zeros. Then, `XNORM = ZERO` in the last iteration ( `J = NRHS` ).
Then, the subroutine computes `RESID = ONE / EPS`. There are many files with the same pattern.

I can propose a PR to solve this (if this is actually a bug).

langou (approved):

TESTING/LIN/ctbt02.f
jip: (Sorry for delayed answer)

I think you are right. I'll use the last variant you've proposed:

> I would go with either this one or the version I proposed. Maybe you can modify it a bit:
> 
> ```
> The test ratio is the maximum over
>   norm(b - op(A)*x) / ( ||op(A)||_1 * norm(x) * EPS ),
> where op(A) = A, A**T, or A**H, b is the column of B, x is
> the solution vector, and EPS is the machine epsilon.
> ```

I'll try either to update this PR or to create a new one.

jip:
> I just noticed another thing that is not related to your changes, but maybe you can help me. Is there a bug here?
> [...]
> I can propose a PR to solve this (if this is actually a bug).

I don't see any problems here. If any column of X is full of zeros, then the RESID takes a maximal value, and the rest columns of X doesn't matter. IMHO.

PR #591, opened 2021-06-24
Use flags -mfpmath=sse (x86-64) -msse2 (x86) in cmake.yml

weslleyspereira:
The flags enable the use of scalar floating-point instructions present in the SSE instruction set for the GCC. Based on the discussion in #575. See also https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

- I let `makefile.yml` without those flags so as to continue testing the default options for GCC.

- Should we add other flags to improve the compatibility of LAPACK with a larger class of compilers?

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -34,8 +34,8 @@ on:
     - '!**md'
 
 env:
-  CFLAGS: "-Wall -pedantic"
-  FFLAGS: "-fimplicit-none -frecursive -fcheck=all"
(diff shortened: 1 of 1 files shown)


christoph-conrads:
SSE and SSE2 are enabled by default on x86-64 with GCC. From the documentation you linked:
> For the x86-32 compiler, you must use `-march=cpu-type`, `-msse` or `-msse2` switches to enable SSE extensions and make this option effective. For the x86-64 compiler, these extensions are enabled by default. 

(This is the first time that I read the term `x86-32`; this should read as `x86`.)

PR #593, opened 2021-06-29
Fix the number of matrices to test in TESTING/{s,c,d,z}ed.in (6->7)

nakatamaho:
The number of matrices to test in TESTING/{s,c,d,z}ed.in look wrong. 
7 seems correct but 6 looks wrong.


weslleyspereira (dismissed):
Thanks!

weslleyspereira (approved):

PR #605, opened 2021-07-26
Add (s,d)ROUNDUP_LWORK functions to fix the workspace computation

### LINKED ISSUE #600: Loss of precision in all single precision workspace queries, implicit and explicit
Hello everybody,

recently I stumbled across an error when using ssyevd via the lapacke interface.
It points to a problem in the lapack interface in general. It goes like this:

According the the lapack standard interface, many routines like ssyevd you have to call twice:
Once for asking the routine how much scratch memory it needs for a certain matrix size,
and only then call the routine in earnest with the required scratch memory sections
as parameters.

If you look closely at this first call, which should return the required memory size, e.g.
for a routine like ssyevd, you see that even according to the lapack documentation,
the memory requirement is passed back via a pointer to a **float** value.
So when calculating the memory it goes through a series of values:

    calulate memory       ->    store value in refernce   ->  retrieve the value for use (allocation)
        int64                            float                              int64

It is int64 is in case of an ilp64 interface, otherwise it would be int32.
So in essence, we have an intermediary shortening of the memory value from
**63 bit to 24 bit !!!**
(or more precise, to 24bit between the outermost set bits in IEEE 754 float representation)
Even in the case of 32bit integers, you have a shortening of 31 bits to 24 bits.

So, if you would call the memory requirement calculation as in the past 'by hand'
you might have chance to see where it goes wrong, but still you cannot prevent the shorting
to a float value. If you use the automatic memory allocation via the modern lapacke interface
you don't even have an idea what could be wrong, as the routine advertises to take care
of all memory management by itself!

This happens in **all** single precision routines (s/c) in lapack, which calculate
the memory requirement as an intermediary step.
Switching to double precison would then use a double precision reference instead,
increasing the intermediary value to 53 bits instead, which is still not close to the 64 bits
one would assume with a 64bit interface.

Workaround, four possible ways:
1. If you want to use single or complex lapack  routines, do not use the automatic memory allocation via the C lapacke interface
2. If you use the two-call lapack function method, for the memory calculation use the double(!) routine
3.  Have a look at the reference implementation of the lapack routine and calculate the required memory on your own
4. Only use small matrix sizes when using float/complex matrices

Other people have stumbled upon this, but didn't follow it through to the real cause, e.g.
[openBLAS build with int64 support fails on valid input for ssyevd](https://github.com/xianyi/OpenBLAS/issues/2745)

One has to stress, that at least with the two-call method, this is not a bug, but a design flaw.
In case of the lapacke automatic memory allocation, it has to be considered a fairly severe bug.

Regards,

oxydicer

weslleyspereira:
__[edited]__

**Description**

There is a known bug in LAPACK related to workspaces smaller than the value computed by the routine. This is because the routines use a floating-point variable to return the integer workspace size. See #600. The intent here is to mitigate the problem of overestimating the workspace when doing the conversion INTEGER -> REAL (or DOUBLE PRECISION) highlighted in #600. I do not solve the problem of a possible overflow in LWORK.

This PR applies the same strategy from the MAGMA package to solve the issue (See https://github.com/Reference-LAPACK/lapack/issues/600#issuecomment-884426097 from @mgates3). 

**Checklist**

- [x] Fix all *GESDD routines.
- [ ] Fix all the routines in LAPACK.

### DIFF
--- a/INSTALL/Makefile
+++ b/INSTALL/Makefile
@@ -48,3 +48,6 @@ cleantest:
 
 slamch.o: slamch.f ; $(FC) $(FFLAGS_NOOPT) -c -o $@ $<
 dlamch.o: dlamch.f ; $(FC) $(FFLAGS_NOOPT) -c -o $@ $<
+
+sroundup_lwork.o: sroundup_lwork.f ; $(FC) $(FFLAGS_NOOPT) -c -o $@ $<
+droundup_lwork.o: droundup_lwork.f ; $(FC) $(FFLAGS_NOOPT) -c -o $@ $<
--- a/INSTALL/droundup_lwork.f
+++ b/INSTALL/droundup_lwork.f
@@ -64,12 +64,21 @@
 * =====================================================================
 *     ..
 *     .. External Functions ..
-      DOUBLE PRECISION   DLAMCH
-      EXTERNAL           DLAMCH
+      DOUBLE PRECISION  DLAMCH
+      EXTERNAL          DLAMCH
+*     ..
+*     .. Intrinsic Functions ..
+      INTRINSIC         DIGITS, RADIX
 *     ..
 *     .. Executable Statements ..
 *     ..
-      DROUNDUP_LWORK = LWORK * ( 1 + DLAMCH('EPS') )
+      DROUNDUP_LWORK = LWORK
+*
+      IF( DROUNDUP_LWORK .GE. DBLE(RADIX(0.0D+0))**DIGITS(0.0D+0) ) THEN
+*         If LWORK can't be represented exactly in double precision
+          DROUNDUP_LWORK = LWORK * ( 1.0D+0 + DLAMCH('EPS') )
+      ENDIF
+*
       RETURN
 *
 *     End of DROUNDUP_LWORK
--- a/INSTALL/droundup_lwork.f
+++ b/INSTALL/droundup_lwork.f
@@ -21,13 +21,17 @@
 *> \verbatim
 *>
 *> DROUNDUP_LWORK deals with a subtle bug with returning LWORK as a Float.
-*> If LWORK > 2**24, then it will get rounded as a Float.
 *> This routine guarantees it is rounded up instead of down by
-*> multiplying LWORK by 1+eps, where eps is the relative machine precision.
+*> multiplying LWORK by 1+eps when it is necessary, where eps is the relative machine precision.
+*> E.g.,
+*>
+*>        float( 9007199254740993            ) == 9007199254740992
+*>        float( 9007199254740993 ) * (1.+eps) == 9007199254740994
 *>
 *> \return DROUNDUP_LWORK
 *> \verbatim
 *>         DROUNDUP_LWORK >= LWORK.
+*>         DROUNDUP_LWORK is guaranteed to have zero decimal part.
 *> \endverbatim
 *
 *  Arguments:
@@ -64,14 +68,14 @@
 * =====================================================================
 *     ..
 *     .. Intrinsic Functions ..
-      INTRINSIC         DIGITS, RADIX, EPSILON
+      INTRINSIC         EPSILON, DBLE, INT
 *     ..
 *     .. Executable Statements ..
 *     ..
-      DROUNDUP_LWORK = LWORK
+      DROUNDUP_LWORK = DBLE( LWORK )
 *
-      IF( DROUNDUP_LWORK .GE. DBLE(RADIX(0.0D+0))**DIGITS(0.0D+0) ) THEN
-*         If LWORK can't be represented exactly in double precision
+      IF( INT( DROUNDUP_LWORK ) .LT. LWORK ) THEN
+*         Force round up of LWORK
           DROUNDUP_LWORK = DROUNDUP_LWORK * ( 1.0D+0 + EPSILON(0.0D+0) )
       ENDIF
 *
--- a/INSTALL/droundup_lwork.f
+++ b/INSTALL/droundup_lwork.f
@@ -1,0 +1,77 @@
+*> \brief \b DROUNDUP_LWORK
+*
+*  =========== DOCUMENTATION ===========
+*
+* Online html documentation available at
+*            http://www.netlib.org/lapack/explore-html/
+*
+*  Definition:
+*  ===========
+*
+*      DOUBLE PRECISION FUNCTION DROUNDUP_LWORK( LWORK )
+*
+*     .. Scalar Arguments ..
+*      INTEGER          LWORK
+*     ..
+*
+*
+*> \par Purpose:
+*  =============
+*>
+*> \verbatim
+*>
+*> DROUNDUP_LWORK deals with a subtle bug with returning LWORK as a Float.
+*> If LWORK > 2**24, then it will get rounded as a Float.
+*> This routine guarantees it is rounded up instead of down by
+*> multiplying LWORK by 1+eps, where eps is the relative machine precision.
+*>
+*> \return DROUNDUP_LWORK
+*> \verbatim
+*>         DROUNDUP_LWORK >= LWORK.
+*> \endverbatim
+*
+*  Arguments:
+*  ==========
+*
+*> \param[in] LWORK Workspace size.
+*
+*  Authors:
+*  ========
+*
+*> \author Weslley Pereira, University of Colorado Denver, USA
+*
+*> \ingroup auxOTHERauxiliary
+*
+*> \par Further Details:
+*  =====================
+*>
+*> \verbatim
+*>  This routine was inspired in the method `magma_zmake_lwork` from MAGMA.
+*>  \see https://bitbucket.org/icl/magma/src/master/control/magma_zauxiliary.cpp
+*> \endverbatim
+*
+*  =====================================================================
+      DOUBLE PRECISION FUNCTION DROUNDUP_LWORK( LWORK )
+*
+*  -- LAPACK auxiliary routine --
+*  -- LAPACK is a software package provided by Univ. of Tennessee,    --
+*  -- Univ. of California Berkeley, Univ. of Colorado Denver and NAG Ltd..--
+*
+*     .. Scalar Arguments ..
+      INTEGER          LWORK
+*     ..
+*
+* =====================================================================
+*     ..
+*     .. External Functions ..
+      DOUBLE PRECISION   DLAMCH
+      EXTERNAL           DLAMCH
+*     ..
+*     .. Executable Statements ..
+*     ..
+      DROUNDUP_LWORK = LWORK * ( 1 + DLAMCH('EPS') )
+      RETURN
+*
+*     End of DROUNDUP_LWORK
+*
+      END
--- a/INSTALL/droundup_lwork.f
+++ b/INSTALL/droundup_lwork.f
@@ -63,12 +63,8 @@
 *
 * =====================================================================
 *     ..
-*     .. External Functions ..
-      DOUBLE PRECISION  DLAMCH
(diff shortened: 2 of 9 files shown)


luszczek:
Please see my comment to #600 If integer LWORK overflow then adding EPS won't help.

weslleyspereira:
Thanks for the comments @luszczek. In fact, these changes have nothing to do with
https://icl.cs.utk.edu/lapack-forum/viewtopic.php?t=1418 or
http://icl.cs.utk.edu/lapack-forum/archives/lapack/msg00827.html
I will remove them from the description. 

The intent here is to mitigate the loss of precision when doing the conversion INTEGER -> REAL (or DOUBLE PRECISION) highlighted in #600.

weslleyspereira:
Two commits with updates in the ROUNDUP_LWORK functions:
1. ROUNDUP_LWORK functions now always return an integer
2. Fix bug: ROUNDUP_LWORK must use (1+EPSILON(0)) instead of (1+LAMCH(EPS))

With (1), the roundup routines always return a floating-point number that is an integer. I do that by checking whether
```fortran
SROUNDUP_LWORK .GE. REAL(RADIX(0.0E+0))**DIGITS(0.0E+0)  
```
The latter is valid for the single-precision case. I used a similar test for the double-precision case. 

The [code in MAGMA](https://bitbucket.org/icl/magma/src/master/control/magma_zauxiliary.cpp) uses DLAMCH('E'). I am not sure this is the correct option. I ran a few tests (below) and saw that one must use EPSILON(0) (or 2*LAMCH(EPS)) to get the desired rounding up.

__Some tests:__

I compiled the LAPACK with `-fdefault-integer-8` and used [this gist](https://gist.github.com/weslleyspereira/7eed7d01dbb3845ea43fda9974d5ff6d) to obtain:

```
 SROUNDUP_LWORK(0)        =    0.00000000    
 SROUNDUP_LWORK(1)        =    1.00000000    
 SROUNDUP_LWORK(317)      =    317.000000    
 DROUNDUP_LWORK(0)        =    0.0000000000000000     
 DROUNDUP_LWORK(1)        =    1.0000000000000000     
 DROUNDUP_LWORK(317)      =    317.00000000000000     
 
 Tests with X = 16777216 = 2**24
X-2 = 16777214; SROUNDUP_LWORK(X-2) = 16777214.0
X-1 = 16777215; SROUNDUP_LWORK(X-1) = 16777215.0
X   = 16777216; SROUNDUP_LWORK(X  ) = 16777218.0
X+1 = 16777217; SROUNDUP_LWORK(X+1) = 16777218.0
X+2 = 16777218; SROUNDUP_LWORK(X+2) = 16777220.0
 
 Tests with X = 9007199254740992 = 2**53
X-2 = 9007199254740990; SROUNDUP_LWORK(X-2) = 9007200328482816.0
X-1 = 9007199254740991; SROUNDUP_LWORK(X-1) = 9007200328482816.0
X   = 9007199254740992; SROUNDUP_LWORK(X  ) = 9007200328482816.0
X+1 = 9007199254740993; SROUNDUP_LWORK(X+1) = 9007200328482816.0
X+2 = 9007199254740994; SROUNDUP_LWORK(X+2) = 9007200328482816.0
 
 Tests with X = 9007199254740992 = 2**53
X-2 = 9007199254740990; DROUNDUP_LWORK(X-2) = 9007199254740990.0
X-1 = 9007199254740991; DROUNDUP_LWORK(X-1) = 9007199254740991.0
X   = 9007199254740992; DROUNDUP_LWORK(X  ) = 9007199254740994.0
X+1 = 9007199254740993; DROUNDUP_LWORK(X+1) = 9007199254740994.0
X+2 = 9007199254740994; DROUNDUP_LWORK(X+2) = 9007199254740996.0
```

just to be sure the output has only 0 decimal parts.

weslleyspereira:
Verifying the round down/up in the conversion from INTEGER(8) to REAL and DOUBLE PRECISION:

1) Tests with X = 16777216 = 2**24
 
 I |    X + I |  REAL(X+1) |  DBLE(X+1)
---|----------|------------|-----------
-2 | 16777214 | 16777214.0 | 16777214.0
-1 | 16777215 | 16777215.0 | 16777215.0
 0 | 16777216 | 16777216.0 | 16777216.0
 1 | 16777217 | 16777216.0 | 16777217.0
 2 | 16777218 | 16777218.0 | 16777218.0
 3 | 16777219 | 16777220.0 | 16777219.0
 4 | 16777220 | 16777220.0 | 16777220.0
 5 | 16777221 | 16777220.0 | 16777221.0
 6 | 16777222 | 16777222.0 | 16777222.0
 7 | 16777223 | 16777224.0 | 16777223.0
 8 | 16777224 | 16777224.0 | 16777224.0
 9 | 16777225 | 16777224.0 | 16777225.0
10 | 16777226 | 16777226.0 | 16777226.0
11 | 16777227 | 16777228.0 | 16777227.0
12 | 16777228 | 16777228.0 | 16777228.0
13 | 16777229 | 16777228.0 | 16777229.0
14 | 16777230 | 16777230.0 | 16777230.0
15 | 16777231 | 16777232.0 | 16777231.0
16 | 16777232 | 16777232.0 | 16777232.0
17 | 16777233 | 16777232.0 | 16777233.0
18 | 16777234 | 16777234.0 | 16777234.0
19 | 16777235 | 16777236.0 | 16777235.0
20 | 16777236 | 16777236.0 | 16777236.0

 2) Tests with X = 9007199254740992 = 2**53

 I |            X + I |          REAL(X+1) |          DBLE(X+1)
---|------------------|--------------------|-------------------
-2 | 9007199254740990 | 9007199254740992.0 | 9007199254740990.0
-1 | 9007199254740991 | 9007199254740992.0 | 9007199254740991.0
 0 | 9007199254740992 | 9007199254740992.0 | 9007199254740992.0
 1 | 9007199254740993 | 9007199254740992.0 | 9007199254740992.0
 2 | 9007199254740994 | 9007199254740992.0 | 9007199254740994.0
 3 | 9007199254740995 | 9007199254740992.0 | 9007199254740996.0
 4 | 9007199254740996 | 9007199254740992.0 | 9007199254740996.0
 5 | 9007199254740997 | 9007199254740992.0 | 9007199254740996.0
 6 | 9007199254740998 | 9007199254740992.0 | 9007199254740998.0
 7 | 9007199254740999 | 9007199254740992.0 | 9007199254741000.0
 8 | 9007199254741000 | 9007199254740992.0 | 9007199254741000.0
 9 | 9007199254741001 | 9007199254740992.0 | 9007199254741000.0
10 | 9007199254741002 | 9007199254740992.0 | 9007199254741002.0
11 | 9007199254741003 | 9007199254740992.0 | 9007199254741004.0
12 | 9007199254741004 | 9007199254740992.0 | 9007199254741004.0
13 | 9007199254741005 | 9007199254740992.0 | 9007199254741004.0
14 | 9007199254741006 | 9007199254740992.0 | 9007199254741006.0
15 | 9007199254741007 | 9007199254740992.0 | 9007199254741008.0
16 | 9007199254741008 | 9007199254740992.0 | 9007199254741008.0
17 | 9007199254741009 | 9007199254740992.0 | 9007199254741008.0
18 | 9007199254741010 | 9007199254740992.0 | 9007199254741010.0
19 | 9007199254741011 | 9007199254740992.0 | 9007199254741012.0
20 | 9007199254741012 | 9007199254740992.0 | 9007199254741012.0

weslleyspereira:
Another related test shows how `REAL(INT( ... ))` and `DBLE(INT( ... ))` behave for huge floating-point numbers. Note in the results that `INT(REAL(HUGE( X )))` and `INT(REAL(HUGE( X )) * 2)` both return the lowest negative integer of the type of X. 

```Fortran
      PROGRAM test_convert
      IMPLICIT NONE

      INTEGER(4)           intX
      INTEGER(8)           X
      REAL                 realX
      DOUBLE PRECISION     dbleX

      INTRINSIC            REAL, DBLE, HUGE
      realX = REAL(HUGE(intX))
      dbleX = DBLE(HUGE(intX))
      WRITE(*,992) HUGE(intX), realX, INT(realX,4), REAL(INT(realX,4))
      WRITE(*,992) HUGE(intX), dbleX, INT(dbleX,4), DBLE(INT(dbleX,4))
      
      realX = REAL(HUGE(intX)) * 2
      dbleX = DBLE(HUGE(intX)) * 2
      WRITE(*,992) HUGE(intX), realX, INT(realX,4), REAL(INT(realX,4))
      WRITE(*,992) HUGE(intX), dbleX, INT(dbleX,4), DBLE(INT(dbleX,4))

      realX = REAL(HUGE(X))
      dbleX = DBLE(HUGE(X))
      WRITE(*,993) HUGE(X), realX, INT(realX,8), REAL(INT(realX,8))
      WRITE(*,993) HUGE(X), dbleX, INT(dbleX,8), DBLE(INT(dbleX,8))

      realX = REAL(HUGE(X)) * 2
      dbleX = DBLE(HUGE(X)) * 2
      WRITE(*,993) HUGE(X), realX, INT(realX,8), REAL(INT(realX,8))
      WRITE(*,993) HUGE(X), dbleX, INT(dbleX,8), DBLE(INT(dbleX,8))

  992 FORMAT( I11, "; ", F13.1, "; ", I11, "; ", F13.1 )
  993 FORMAT( I20, "; ", F22.1, "; ", I20, "; ", F22.1 )

      END PROGRAM test_convert
```

Output (compiled with `gfortran -fdefault-integer-8` gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)):

```
 2147483647;  2147483648.0; -2147483648; -2147483648.0
 2147483647;  2147483647.0;  2147483647;  2147483647.0
 2147483647;  4294967296.0; -2147483648; -2147483648.0
 2147483647;  4294967294.0; -2147483648; -2147483648.0
 9223372036854775807;  9223372036854775808.0; -9223372036854775808; -9223372036854775808.0
 9223372036854775807;  9223372036854775808.0; -9223372036854775808; -9223372036854775808.0
 9223372036854775807; 18446744073709551616.0; -9223372036854775808; -9223372036854775808.0
 9223372036854775807; 18446744073709551616.0; -9223372036854775808; -9223372036854775808.0
```

weslleyspereira:
What about forcing round up only if `INT( REAL( LWORK ) ) .LT. LWORK`? See 51eee01. This new implementation should be optimal: It guarantees that *ROUNDUP_LWORK returns the smallest floating-point number that:
1. has zero decimal part.
2. is greater than or equal to LWORK.

This test can be less performant than testing `REAL( LWORK ) .GE. REAL(RADIX(0.0E+0))**DIGITS(0.0E+0)`. I say that because we have one extra operation `INT( ... )` when compared to cea9568. But I prefer 51eee01 since the output should be optimal.

mgates3:
Regarding MAGMA, after precision-generation, `magma_sauxiliary.cpp` uses `lapackf77_slamch("eps")`, but it does the computation in *double*. If it computed 1.0f + slamch("eps") in single, it would get 1.0f, which isn't useful. Note real_Double_t is double.
```
> less magma/control/magma_sauxiliary.cpp 
...
float magma_smake_lwork( magma_int_t lwork )
{
    ...
    real_Double_t one_eps = 1. + lapackf77_slamch("Epsilon");
    return MAGMA_S_MAKE( float(lwork*one_eps), 0 );
    ...
}
```
This works up to lwork around 2^53 (33,554,432 GiB in single). Using 1 + epsilon, for epsilon = 1.19e-07 as defined by Fortran, C, C++, Matlab, etc. works for even larger lwork (I checked up to 2^61). Note slamch("eps") returns unit roundoff, u = 5.96e-08, which is epsilon/2.

weslleyspereira:
Thanks for the clarification, @mgates3 !

langou (approved):

PR #607, opened 2021-08-08
Make lapy and nrm2 consistent

angsch:
**Description**

If some of the inputs are `inf`, `{s,d}lapy{2,3}` and the new` {s,d}nrm2` routines return inconsistent results, `NaN` in contrast to `inf`.

- lapy2 if both inputs are `inf`
- lapy3 if at least one of the 3 inputs is `inf`

This commit adds a check to avoid the divisions `inf / inf` that introduce `NaN`.

**Checklist**

### DIFF
--- a/SRC/dlapy2.f
+++ b/SRC/dlapy2.f
@@ -78,12 +78,15 @@
       PARAMETER          ( ONE = 1.0D0 )
 *     ..
 *     .. Local Scalars ..
-      DOUBLE PRECISION   W, XABS, YABS, Z
+      DOUBLE PRECISION   W, XABS, YABS, Z, HUGEVAL
       LOGICAL            X_IS_NAN, Y_IS_NAN
 *     ..
 *     .. External Functions ..
       LOGICAL            DISNAN
       EXTERNAL           DISNAN
+*     ..
+*     .. External Subroutines ..
+      DOUBLE PRECISION   DLAMCH
 *     ..
 *     .. Intrinsic Functions ..
       INTRINSIC          ABS, MAX, MIN, SQRT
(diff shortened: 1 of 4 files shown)


langou (commented):
Thanks!

langou (commented):
1. Instead of using the FORTRAN INTRINSIC HUGE, it might be better to use DLAMCH( 'Overflow' ).
2. Instead of calling MAXN, maybe call this variable HUGEVAL. (I agree that MAX is better to describe this variable, but HUGE is what LAPACK and FORTRAN seem to be using, so I think we want to be consistent.)
3. Instead of using ==, it might be better to use .EQ. (for consistency throughout LAPACK)

langou (commented):
General question. Why does LAPACK not have DISPOSINFTY ? There is DISNAN, but not DISPOSINFTY. Such that the W > MAXN of @angsch would be a  DISPOSINFTY( W ). (And we can discuss on the same of this subroutine.)

langou (commented):
Very nice PR @angsch ! "Correct" behavior of LAPACK routines under NaN and Inf is something we are working on right now, and this is a great contribution! 

The first part of the current effort is to define what the "correct" behavior of LAPACK routines under NaN and Inf should be. The second part is to make the "correct" behavior happen in the routines. And finally the third and last part is to write some test that check for  "correct" behavior under NaN and Inf.

Perfect timing for this contribution!

As far what the "correct" behavior should be, I certainly agree that the output of SQRT( ( +/- INF ** 2 ) + ( +/- INF ) ** 2 ) ) is better defined as an +INF than a NaN.

angsch:
Where should the tests go? The testing directory appears to so far not cover any auxiliary routines.

langou:
Hi @angsch, it is not clear where these tests will go. Maybe in LAPACK/TESTING proper, maybe somewhere else. I think it is great that you are identifying a problem and give a PR with a good solution. It would be good to have the related TESTING but we are not in place yet, so I think the best as far as TESTING goes is to wait. Cheers, Julien.

weslleyspereira (approved):

Issue #606, opened 2021-07-26
lapack > 3.8.0 fails to build with IBM XL compiler


quellyn:
Beginning with lapack version 3.9.0, I am unable to build using the IBM XL compiler on a Power9 Linux platform.

To reproduce:

1. Login to a Power9 Linux platform with a relatively recent version of the IBM XL compiler and CMake:
```
$ arch
ppc64le
```
```
$ xlf -qversion
IBM XL Fortran for Linux, V16.1.1 (5725-C75, 5765-J15)
Version: 16.01.0001.0007
```
```
$ cmake --version
cmake version 3.19.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```
2. Download lapack > 3.8.0:
```
$ wget https://github.com/Reference-LAPACK/lapack/archive/refs/tags/v3.10.0.tar.gz
$ tar xvzf https://github.com/Reference-LAPACK/lapack/archive/refs/tags/v3.10.0.tar.gz
```
3. Attempt cmake build:
```
$ mkdir build; cd build
[...]

PR #621, opened 2021-09-15
Address issues with IBM XL builds

quellyn:
**Description**
This PR addresses #606, by adding correct detection for the IBM XL Fortran (`xlf`) recursive option (`-qrecur`). 

Note that the current logic will test ALL possible recursive flags against a target compiler; in the case of `xlf` this unexpectedly "succeeds" when the Intel-specific flag `-recursive` is passed in. Actually it doesn't really succeed, but `xlf` does not correctly report failure, so `cmake CheckFortranCompilerFlag` ends up tacking `-recursive` onto `CMAKE_Fortran_FLAGS`. The resulting attempt to compile with this bogus flag fails.

You can see the errant compiler behavior with a simple example "Hello World" program:
```
$ xlf_r -recursive -o hello-dumb hello.f90
** helloworld   === End of Compilation 1 ===
1501-510  Compilation successful for file hello.f90.
ld: cannot find -lgcc_s
ld: cannot find -lgcc_s

$ ls
hello.f90  hello.o

$ file hello.o
hello.o: ELF 64-bit LSB relocatable, 64-bit PowerPC or cisco 7500, version 1 (SYSV), not stripped
```
The `xlf` compiler produces an intermediate object code that will not link.

P.S. I've sent this patch to the [Spack project](https://github.com/spack/spack/pull/25793). They've accepted it but of course it would be best to implement the fixes upstream.

**Checklist**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,28 +91,40 @@ include(PreventInBuildInstalls)
 
 # Check if recursive flag exists
 include(CheckFortranCompilerFlag)
-check_fortran_compiler_flag("-recursive" _recursiveFlag)
-check_fortran_compiler_flag("-frecursive" _frecursiveFlag)
-check_fortran_compiler_flag("-Mrecursive" _MrecursiveFlag)
+if(CMAKE_Fortran_COMPILER_ID STREQUAL Flang)
+  check_fortran_compiler_flag("-Mrecursive" _MrecursiveFlag)
+elseif(CMAKE_Fortran_COMPILER_ID STREQUAL GNU)
+  check_fortran_compiler_flag("-frecursive" _frecursiveFlag)
+elseif(CMAKE_Fortran_COMPILER_ID STREQUAL Intel)
+  check_fortran_compiler_flag("-recursive" _recursiveFlag)
+elseif(CMAKE_Fortran_COMPILER_ID STREQUAL XL)
+  check_fortran_compiler_flag("-qrecur" _qrecurFlag)
+endif()
 
 # Add recursive flag
-if(_recursiveFlag)
-  string(REGEX MATCH "-recursive" output_test <string> "${CMAKE_Fortran_FLAGS}")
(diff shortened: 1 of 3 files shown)


quellyn:
I should mention @martin-frbg, who helped me debug this issue.

CMakeLists.txt
```diff
@@ -91,28 +91,40 @@ include(PreventInBuildInstalls)
...
-check_fortran_compiler_flag("-Mrecursive" _MrecursiveFlag)
+if(CMAKE_Fortran_COMPILER_ID STREQUAL Flang)
+  check_fortran_compiler_flag("-Mrecursive" _MrecursiveFlag)
+elseif(CMAKE_Fortran_COMPILER_ID STREQUAL GNU)
+  check_fortran_compiler_flag("-frecursive" _frecursiveFlag)
+elseif(CMAKE_Fortran_COMPILER_ID STREQUAL Intel)
+  check_fortran_compiler_flag("-recursive" _recursiveFlag)
+elseif(CMAKE_Fortran_COMPILER_ID STREQUAL XL)
+  check_fortran_compiler_flag("-qrecur" _qrecurFlag)
+endif()
```
weslleyspereira, line 102: I think it is a good idea to have a non-empty `else()` for other compiler IDs. I would considers using either:
```cmake
else()
  check_fortran_compiler_flag("-recursive" _recursiveFlag)
  check_fortran_compiler_flag("-frecursive" _frecursiveFlag)
  check_fortran_compiler_flag("-Mrecursive" _MrecursiveFlag)
  check_fortran_compiler_flag("-qrecur" _qrecurFlag)
endif()
```
or:
```cmake
else()
  message( WARNING "Fortran local arrays should be allocated on the stack."
    " Please assure your compiler guarantees that."
    " See https://github.com/Reference-LAPACK/lapack/pull/188 and references therein." )
endif()
```

weslleyspereira (approved):
Thanks!

PR #622, opened 2021-09-19
Regression test for illegal modification of Y in xGEMV.

matcross:
At NAG we've observed that Apple vecLib modifies Y on output from SGEMV and DGEMV when TRANS is 'T', M is zero, and N is nonzero.
Here is a test case for all four xGEMV's and four xGBMV's.
(It looked unnecessary to try to incorporate this into the data-file mechanism that the xblat2 tests use, so I hope this is OK.)


weslleyspereira (approved):
Thanks! Your tests make sense to me, I don't think it is necessary to change the input files.

PR #623, opened 2021-09-22
Test the Fortran intrinsic ABS and complex divisions and report fails during build

weslleyspereira:
**Description**

After the long discussion in #575, I started testing the intrinsic routines complex ABS and complex division of some Fortran compilers. It is good that the user are aware of the compiler behavior.

This PR puts 2 new tests in LAPACK. The tests run at the build process. The build system reports any failures via command line, and the details are stored in disk inside the INSTALL directory. With this information, the user can better define/configure its Fortran compiler. 

**Checklist**

- [x] The changes were implemented on the CMake build system.
- [x] The changes were implemented on the Makefile build system.


weslleyspereira:
I am proposing a short output during the build. For example, when compiling with gfortran 9.3.0 :

```bash
./test_zcomplexabs 2> test_zcomplexabs.err
./test_zcomplexdiv 2> test_zcomplexdiv.err
 !! Some (x+x*I)/(x+x*I) differ from 1
 !! Some (x+x*I)/(x-x*I) differ from I
 # Please check the failed divisions in [stderr]
```

The complete description of the failing tests si in the stderr file:

```
[ c3] X =  1.7976931348623157E+308 : (x+x*I)/(x+x*I) =                      NaN+0.0000000000000000E+000*I differs from +1.0000000000000000E+000
[ c4] X =  8.9884656743115795E+307 : (x+x*I)/(x+x*I) =                      NaN+0.0000000000000000E+000*I differs from +1.0000000000000000E+000
[ f3] X =  1.7976931348623157E+308 : (x+x*I)/(x-x*I) =  0.0000000000000000E+000                     NaN*I differs from +0.0000000000000000E+000+1.0000000000000000E+000*I
[ f4] X =  8.9884656743115795E+307 : (x+x*I)/(x-x*I) =  0.0000000000000000E+000                     NaN*I differs from +0.0000000000000000E+000+1.0000000000000000E+000*I
```

The identifiers, e.g., `[c3]`, relate to the failing test 'C' with input data '3'. Please, look at the documentation of each routine to find information about it.

weslleyspereira:
I have just added tests for the intrinsic MIN, MAX and for the complex multiplication. Interesting, I found that

```
[i7] MIN(   0.,  NaN) =   NaN
[i7] MAX(   0.,  NaN) =   NaN
[i8] MIN(  NaN,   0.) =    0.
[i8] MAX(  NaN,   0.) =    0.
```

using gfortran 9.3.0 on a Linux machine `(Ubuntu 9.3.0-17ubuntu1~20.04)`. Following the standard IEEE754-2019, this is "a common way of defining min and max in a high-level language" (See [nan_propagation.pdf](https://www.agner.org/optimize/nan_propagation.pdf) for instance). Since MIN and MAX are currently used in LAPACK, we should signal these intrinsic functions may not propagate NaNs.

weslleyspereira:
Personal note about complex multiplications in C and Fortran using GCC:
In GCC 11.2 on a x86-64 machine, the multiplications `real*complex` and `complex*complex` generate two different codes when using the C language (see https://godbolt.org/z/r9Esdvnfr). In this code, `(1 * (Inf+0*I)) = (Inf+0*I)` and `((1+0*I) * (Inf+0*I)) = (Inf+NaN*I)`. Using the same compiler, but with a code in Fortran, both multiplications return `(Inf+NaN*I)`. (run the code in https://godbolt.org/z/fa839ca14)

langou (approved):

PR #617, opened 2021-09-01
use integer64 flag for Cray Fortran compiler

ajaypanyala:
**Description**
  use `default64` flag for Cray ftn compiler when BUILD_INDEX64 is ON.

**Checklist**

- [ X] The documentation has been updated.
- [X ] If the PR solves a specific issue, it is set to be closed on merge.

Issue #461, opened 2020-11-01
Allow a possible installation of index-64 library alongside standard index-32 library?


epsilon-0:
Currently debian has (some) separate headers for blas64 and cblas64 (though not from the reference implementation).

I am not sure if they are correct or not, with regards to the reference index64 API (they are from the blis library).

Would it be possible to add an option to cmake, something like `BUILD_INDEX64`, which is defaulted to `OFF`, but if has been turned on, it will create the index-64 libraries?   
If I make a PR for such an option, would it be entertained as a possibility?

Some things which I had in mind for allowing this to co-exist with the standard installation - name the libraries to `libblas64.so, libcblas64.so, liblapack64.so, liblapacke64.so`, this way there is no conflict between the library names (though of course, you can't link with both libblas and libblas64 at the same time).
Also the library would need to be compiled twice, once for index32 and once for index64 (but that is a perfectly normal scenario and not a deal breaker).
The only conflict I am unable to think of a good resolution is the header file names.
If following debians style, it might be wise to call the c headers ending with 64 as well.
(I'm maintaining this for Gentoo and would like to keep the ecosystem very close to debian, so that we have minimal problems for developers switching between systems)

I'm open to any suggestions before I make the PR :heart:

Thanks,
Aisha

langou:
Hi Aisha, that makes total sense to me, but let us see if we have feedback from others. So let us wait a few days. J

epsilon-0:
Most definitely, that sounds like a good plan.

epsilon-0:
OMG @langou 
you are so fast :heart: 

Just for completeness, I am writing down things that we still have to do:
- Figure out how to name headers so that 32bit API can co-exist with 64bit API
- Fix printf/fprintf statements so that they use correct qualifier for printing.

Any suggestions to solve the first point are welcome, I unfortunately don't have a "clean" solution.

epsilon-0:
A couple of questions, which will help me handle the naming for the files
- There seems to be a tonne of duplicate definitions between `cblas_f77.h` and `cblas_test.h`. Do we really need that?
- Does `cblas_test.h` need to be installed? Given its name (and the files that it is used in) I presume it to be used only during the testing phase. Maybe we should not install this file on a system-wide level?

christoph-conrads:
Hi @epsilon-0,

[...]

PR #624, opened 2021-09-27
ILP windows fix

ericlars:
**Description**
CBLAS and LAPACKE layers are not built with 8 byte integers when cmake option BUILD_INDEX64=ON is passed. 
The reason is because long's in windows are 4 bytes (MSVC c/c++, intel compilers). This patch uses int64_t and int32_t
to ensure ILP interface is built.

Additionally, this patch adds support ifort's 8 byte integer flag to cmake system for building BLAS and LAPACK.

**Checklist**


weslleyspereira (approved):
Thanks @ericlars! This issue also appears in the discussion from #461, maybe you want to take a look.

CMAKE/CheckLAPACKCompilerFlags.cmake
```diff
@@ -14,6 +14,19 @@ macro( CheckLAPACKCompilerFlags )
...
+# FORTRAN ILP default
+if ( FORTRAN_ILP )
+    if( CMAKE_Fortran_COMPILER_ID STREQUAL "Intel" )
+        if ( WIN32 )
+            set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} /integer-size:64")
+        else ()
+            set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -integer-size 64")
+        endif()
+    else()
+        set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fdefault-integer-8")
```
ajaypanyala, line 26: @weslleyspereira The [logic](https://github.com/Reference-LAPACK/lapack/pull/617/files) in #617 needs to go here.

PR #631, opened 2021-10-15
New algorithms for computing Givens rotations

### LINKED ISSUE #629: accuracy problems with the new LAPACK 3.10  *LARTG implementations
**Description**
Under certain conditions the new *LARTG implementations can essentially increase the 2-norm  of initial vector. It can lead to significant loss of accuracy of singular vectors or eigenvectors in ScaLAPACK or LAPACK if the matrix size is large enough. The attached test demonstrates the problem.   The test is for the single complex precision but the precision doesn’t matter. 

In the test  5000 of plane rotations are applied to the vector of  length 2 with the initial 2-norm of 10. If the test is linked with the  reference LAPACK 3.10 built with    
 -bash-4.2$ gcc --version
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-4)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.    
           
 then the 2-norm of  the resulting vector  is   0.10001301E+02

-bash-4.2$ gfortran test.f liblapack.a librefblas.a 
-bash-4.2$ ./a.out
… 
  The 2-norm of the first vector = 
    0.10001301E+02
…
Max of all errors printed above 
    0.13008118E-02

Everything is excellent with the LAPACK 3.9 clartg and here is the output for the release

-bash-4.2$ ./a.out 
...                                                                                                                                                  
   The 2-norm of the first vector = 
    0.99999943E+01
..
  Error abs(norm of the first column  - initial norm )
    0.57220459E-05
...
  Max of all errors printed above 
    0.57220459E-05

**Checklist**

- [ 
[test.zip](https://github.com/Reference-LAPACK/lapack/files/7325814/test.zip)
] I've included a minimal example to reproduce the issue

weslleyspereira:
Closes #629

New algorithms for computing Givens rotations
---------------------------------------------

@sergey-v-kuznetsov highlighted in #629 that the new Givens rotations operations may have lower accuracy than the ones that were in LAPACK up to release 3.9. This was verified after applying several rotations to a initial unitary matrix. This PR proposes:

1. A new algorithm for computing complex Givens rotations.
2. A slight modification in the algorithms for computing real-valued Givens rotations.

Both modifications target the improvement of the output's accuracy.

@langou and I are preparing a report with the numerical analysis and experiments comparing the different algorithms for computing Givens rotations. We will share the document here when it is ready to review. 

**Minor modifications**

1. Use `rtmin = sqrt( safmin )` instead of `rtmin = sqrt( safmin / epsilon )`. The first condition is sufficient to guarantee all real variables used in the intermediate steps of the new algorithm belong to the interval `[safmin,safmax]`.
2. Set `rtmax` to either `sqrt( safmin/4 )`, `sqrt( safmin/2 )` or `sqrt( safmin )`. This variable depends on where it is in the algorithm. The value is the maximum possible in order that all real variables used in the intermediate steps of the new algorithm belong to the interval `[safmin,safmax]`.
3. Eliminate the intermediate computations like `p = one / d`, `uu = one / u` and `vv = one / v`. These operations reduce the number of divisions in the code at the cost of possibly increasing the accumulation error. I am trying to improve accuracy, so I remove the intermediate operations at the cost of having additional floating-point divisions.
4. When `f = 0`, check if `real(g) == 0` or `aimag(g) == 0` to avoid unnecessary `ABSSQ( g ); sqrt( g2 )`. This change reduces the accumulation error when `real(g) == 0` (analogously `aimag(g) == 0`) and `aimag(g)**2` (analogously `real(g)**2`) cannot be stored in the respective finite precision. We choose not to use the intrinsic complex `abs` because its implementation is compiler-dependent.

**Major changes**

The algorithm for computing complex Givens rotations was revisited. This is the new code in `(c,z)ROTG` and `(c,z)LARTG` for the unscaled part:

```f90
f2 = ABSSQ( f )
g2 = ABSSQ( g )
h2 = f2 + g2
! safmin <= f2 <= h2 <= safmax 
if( f2 >= h2 * safmin ) then
    ! safmin <= f2/h2 <= 1, and h2/f2 is finite
    c = sqrt( f2 / h2 )
    r = f / c
    rtmax = rtmax * 2
    if( f2 > rtmin .and. h2 < rtmax ) then
        ! safmin <= sqrt( f2*h2 ) <= safmax
        s = conjg( g ) * ( f / sqrt( f2*h2 ) )
    else
        s = conjg( g ) * ( r / h2 )
    end if
else
    ! f2/h2 <= safmin may be subnormal, and h2/f2 may overflow.
    ! Moreover,
    !  safmin <= f2*f2 * safmax < f2 * h2 < h2*h2 * safmin <= safmax,
    !  sqrt(safmin) <= sqrt(f2 * h2) <= sqrt(safmax).
    ! Also,
    !  g2 >> f2, which means that h2 = g2.
    d = sqrt( f2 * h2 )
    c = f2 / d
    if( c >= safmin ) then
        r = f / c
    else
        ! f2 / sqrt(f2 * h2) < safmin, then
        !  sqrt(safmin) <= f2 * sqrt(safmax) <= h2 / sqrt(f2 * h2) <= h2 * (safmin / f2) <= h2 <= safmax
        r = f * ( h2 / d )
    end if
    s = conjg( g ) * ( f / d )
end if
```

- The worst-case scenario analysis shows this algorithm is more accurate than both the algorithms from 3.9.1 and 3.10.0.
- All real variables used in intermediate computations belong to the interval `[safmin,safmax]`.

**Acknowledgements**

Thanks to people that contributed in the discussions about this code:
- @sergey-v-kuznetsov, @ecanesc, @alilotfi90 and @langou,
- People from @BallisticLA team, especially prof. Jim Demmel.

**Checklist**

- [ ] Time measurements.

### DIFF
--- a/SRC/clartg.f90
+++ b/SRC/clartg.f90
@@ -224,7 +224,6 @@ subroutine CLARTG( f, g, c, s, r )
             h2 = f2 + g2
          end if
          if( f2 > safmin * g2 ) then
-            ! Use a precise algorithm
             d = sqrt( w**2 + g2/f2 )
             c = w / d
             if( f2 > rtmin .and. h2 < rtmax ) then
--- a/SRC/clartg.f90
+++ b/SRC/clartg.f90
@@ -233,7 +233,6 @@ subroutine CLARTG( f, g, c, s, r )
          else
             c = ( f2*p )*w
          end if
-         c = ( f2*p )*w
          s = conjg( gs )*( fs*p )
          r = ( fs*( h2*p ) )*u
       end if
--- a/SRC/clartg.f90
+++ b/SRC/clartg.f90
@@ -117,7 +117,7 @@
 subroutine CLARTG( f, g, c, s, r )
    use LA_CONSTANTS, &
    only: wp=>sp, zero=>szero, one=>sone, two=>stwo, czero, &
-         rtmin=>srtmin, rtmax=>srtmax, safmin=>ssafmin, safmax=>ssafmax
+         safmin=>ssafmin, safmax=>ssafmax
 !
 !  -- LAPACK auxiliary routine (version 3.10.0) --
 !  -- LAPACK is a software package provided by Univ. of Tennessee,    --
@@ -129,7 +129,7 @@ subroutine CLARTG( f, g, c, s, r )
    complex(wp)        f, g, r, s
 !  ..
 !  .. Local Scalars ..
-   real(wp) :: d, f1, f2, g1, g2, h2, u, v, w
+   real(wp) :: d, f1, f2, g1, g2, h2, u, v, w, rtmin, rtmax
    complex(wp) :: fs, gs, t
 !  ..
 !  .. Intrinsic Functions ..
@@ -141,6 +141,9 @@ subroutine CLARTG( f, g, c, s, r )
 !  .. Statement Function definitions ..
    ABSSQ( t ) = real( t )**2 + aimag( t )**2
 !  ..
+!  .. Constants ..
+   rtmin = sqrt( safmin )
+!  ..
 !  .. Executable Statements ..
 !
    if( g == czero ) then
@@ -150,6 +153,7 @@ subroutine CLARTG( f, g, c, s, r )
    else if( f == czero ) then
       c = zero
       g1 = max( abs(real(g)), abs(aimag(g)) )
+      rtmax = sqrt( safmax/2 )
       if( g1 > rtmin .and. g1 < rtmax ) then
 !
 !        Use unscaled algorithm
@@ -170,6 +174,7 @@ subroutine CLARTG( f, g, c, s, r )
    else
       f1 = max( abs(real(f)), abs(aimag(f)) )
       g1 = max( abs(real(g)), abs(aimag(g)) )
+      rtmax = sqrt( safmax/4 )
       if( f1 > rtmin .and. f1 < rtmax .and. &
           g1 > rtmin .and. g1 < rtmax ) then
 !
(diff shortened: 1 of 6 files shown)


SRC/clartg.f90
```diff
@@ -224,6 +228,11 @@ subroutine CLARTG( f, g, c, s, r )
             d = sqrt( f2 )*sqrt( h2 )
          end if
          p = 1 / d
+         if( f2 > safmin * g2 ) then
+            c = (1 / sqrt( one + g2/f2 )) * w
+         else
+            c = ( f2*p )*w
+         end if
          c = ( f2*p )*w
```
langou, line 236: I think you forgot to remove line 236.

langou (dismissed):
Let us wait on feedback from Sergey though.

langou (approved):

SRC/dlartg.f90
```diff
@@ -37,16 +37,14 @@
 !> This version is discontinuous in R at F = 0 but it returns the same
 !> C and S as ZLARTG for complex inputs (F,0) and (G,0).
 !>
-!> This is a more accurate version of the BLAS1 routine DROTG,
+!> This is a more accurate version of the BLAS1 routine SROTG,
```
vladimir-ch, line 40: Why this change? D looks correct.

SRC/zlartg.f90
```diff
@@ -46,11 +50,8 @@
 !> to avoid overflow or underflow in computing the square root of the
 !> sum of squares.
 !>
-!> This is a faster version of the BLAS1 routine ZROTG, except for
-!> the following differences:
-!>    F and G are unchanged on return.
-!>    If G=0, then C=1 and S=0.
-!>    If F=0, then C=0 and S is chosen so that R is real.
+!> This is the same routine CROTG fom BLAS1, except that
```
vladimir-ch, line 53: ZROTG?

weslleyspereira: Thanks for the review!

weslleyspereira:
I have just updated the description (first comment) of this PR. @langou and I are preparing a report with the numerical analysis and experiments comparing the different algorithms for computing Givens rotations. We will share the document here when it is ready to review.

weslleyspereira:
And here is the report: https://arxiv.org/abs/2211.04010.
I am ready to merge it.

langou (approved):

PR #637, opened 2021-11-11
CMake selective build improvements

christoph-conrads:
**Description**

This merge request
* makes CMake always install `lapack.h` even when LAPACKE is not built,
* ensures all necessary object files are compiled when only a subset of the arithmetic (e.g., only real double-precision) is to be built, and
* ensures all LAPACK code required by LAPACKE is present (e.g., `LAPACKE_BUILD_SINGLE=ON` implies `BUILD_SINGLE=ON`).

**Checklist**


christoph-conrads:
You can check for missing code with selective compilation with
```sh
env LDFLAGS='-Wl,-z,defs' cmake -DBUILD_SHARED_LIBS=ON ../lapack
```

BLAS/SRC/CMakeLists.txt
```diff
@@ -38,12 +38,22 @@ set(CBLAS1 scabs1.f scasum.f scnrm2.f90 icamax.f caxpy.f ccopy.f
 set(DBLAS1 idamax.f dasum.f daxpy.f dcopy.f ddot.f dnrm2.f90
 	drot.f drotg.f90 dscal.f dsdot.f dswap.f drotmg.f drotm.f)
 
+set(DB1AUX sscal.f isamax.f)
```
weslleyspereira, line 41: This is the first time I am checking which routines use the auxiliary routines in `*B1AUX`. I couldn't find which double precision routines use `ISAMAX`. Would you please be able to tell? Thanks.

Moreover, I noticed that some of the dependencies, like `SASUM` in `CB1AUX`, are only needed for testing routines. Maybe we should refactor it.

christoph-conrads: > I couldn't find which double precision routines use ISAMAX. Would you please be able to tell? Thanks.

The dependency chain is `dsgesv` which depends on `sgetrf`, `sgetrf2`, `isamax`.

> Moreover, I noticed that some of the dependencies, like SASUM in CB1AUX, are only needed for testing routines. Maybe we should refactor it.

When working on this issue, I was surprised to see that someone had already put in the effort to separate all the code needed for the different floating-point types. In fact, this has been part of the CMake setup since its introduction in 2009. All that was left to do was to update the lists. Since there are no automated tests in place for selective builds, such a change would be prone to be broken again in the future.

weslleyspereira (approved):
Hi @christoph-conrads. Thanks for the improvements!
I think it is helpful to install `lapack.h` even when LAPACKE is off.

PR #640, opened 2021-12-11
cmake: forward the macOS deployment target to the TIME function test

mathstuf:
When using newer Xcode releases, older `gfortran` may be given a
deployment target that is not understood if the host is on a new enough
version. Forward the deployment target from the main project to the
testing project to check with similar settings as the main project.

**Checklist**


weslleyspereira (approved):
Thanks! The modification makes sense to me.

PR #647, opened 2022-01-30
Fix xORBDB6 performs numerically unadvisable operations

### LINKED ISSUE #634: xORBDB6 performs numerically unadvisable operations
**Description**

xORBDB5 and xORBDB6 compute vector norms differently causing disagreement on when a vector is numerically zero.

Given an isometric matrix `Q` and a vector `x`, xORBDB5 computes a vector `x'` that is orthogonal to the span of the columns of `Q`. (`Q` isometric means that the following two properties hold: `Q` has at most as many columns as rows and `Q^* Q = I`). Internally xORBDB5 calls xORBDB6. Given an isometric matrix `Q` and a vector `x`, xORBDB6 projects `x` onto the complement of the column span of `Q`. The code contains only the ominous comment that
> Kahan's "twice is enough" criterion

is applied. Probably this means that xORBDB6 uses at most two iterations of classical Gram-Schmidt orthogonalization to compute its results. This approach is known as _CGS2_, see [BarlowS2011](https://arxiv.org/abs/1108.4209v1) or [GiraudLR2002](https://www.cerfacs.fr/algor/reports/2002/TR_PA_02_33.pdf) and it matches the [matrix-vector multiplications inside this function](https://github.com/Reference-LAPACK/lapack/blob/5d4180cf8288ae6ad9a771d18793d15bd0c5643c/SRC/sorbdb6.f#L230-L239). The problem are the computation of the vector norm.

For a xORBDB5, the norm of the vector computed by xORBDB6 is computed with `SNRM2`:
https://github.com/Reference-LAPACK/lapack/blob/5d4180cf8288ae6ad9a771d18793d15bd0c5643c/SRC/sorbdb5.f#L223-L226

For xORBDB6, the vector norm is computed as follows:
https://github.com/Reference-LAPACK/lapack/blob/5d4180cf8288ae6ad9a771d18793d15bd0c5643c/SRC/sorbdb6.f#L241-L247

Consider the input to xORBDB6 below:
```
c := 2^-45
μ := c ε

    | 0  1 -μ |
Q = | 0  0 +μ |
    | 1  0  0 |
    | 0  μ  1 |
```
With `x = e_2` (i.e., the second standard basis vector), the xORBDB6 variable `NORMSQ2` is zero whereas `SNRM2` is nonzero causing xORBDB5 to return an incorrect vector.

**Checklist**

christoph-conrads:
**Description**

fixes #634

**Checklist**

### DIFF
--- a/SRC/zunbdb6.f
+++ b/SRC/zunbdb6.f
@@ -41,10 +41,16 @@
 *> with respect to the columns of
 *>      Q = [ Q1 ] .
 *>          [ Q2 ]
-*> The columns of Q must be orthonormal.
+*> The Euclidean norm of X must be one and the columns of Q must be
+*> orthonormal. The orthogonalized vector will be zero if and only if it
+*> lies entirely in the range of Q.
 *>
-*> If the projection is zero according to Kahan's "twice is enough"
-*> criterion, then the zero vector is returned.
+*> The projection is computed with at most two iterations of the
+*> classical Gram-Schmidt algorithm, see
+*> * L. Giraud, J. Langou, M. Rozložník. "On the round-off error
+*>   analysis of the Gram-Schmidt algorithm with reorthogonalization."
+*>   2002. CERFACS Technical Report No. TR/PA/02/33. URL:
+*>   https://www.cerfacs.fr/algor/reports/2002/TR_PA_02_33.pdf
 *>
 *>\endverbatim
 *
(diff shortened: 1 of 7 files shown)


christoph-conrads:
The fixes are only implemented for SORBDB6 so far.

langou (dismissed):

christoph-conrads:
@langou The changes from SORBDB6 were applied to DORBDB6, CUNBDB6, ZUNBDB6. The pull request should be complete.

SRC/zunbdb6.f
```diff
@@ -239,27 +248,31 @@ SUBROUTINE ZUNBDB6( M1, M2, N, X1, INCX1, X2, INCX2, Q1, LDQ1, Q2,
...
-      SCL1 = REALZERO
-      SSQ1 = REALONE
-      CALL ZLASSQ( M1, X1, INCX1, SCL1, SSQ1 )
-      SCL2 = REALZERO
-      SSQ2 = REALONE
-      CALL ZLASSQ( M2, X2, INCX2, SCL2, SSQ2 )
-      NORMSQ2 = SCL1**2*SSQ1 + SCL2**2*SSQ2
+      SCL = REALZERO
+      SSQ = REALZERO
+      CALL CLASSQ( M1, X1, INCX1, SCL, SSQ )
```
angsch, line 253:
```suggestion
      CALL ZLASSQ( M1, X1, INCX1, SCL, SSQ )
```
Mismatching data type also on the next line.

christoph-conrads: Thank you, fixed.

christoph-conrads:
The code changes in all four variants are horribly broken: the variable `NORMSQ2` is nowhere introduced, nowhere being assigned to but it is being read. gfortran 11 does not complain about this without `-Wextra` flag. I would love to know the rationale for this incredible behavior because this nonsense is not permissible even in C89.
```
      IF( NORMSQ2 .LE. N * EPS * NORM ) THEN
```

PR #658, opened 2022-04-03
Fix tests

angsch:
**Description**

Fix two bugs that have been introduced with the revision of the input parameter checks. For some reason the LAPACK tests did not rebuild and I did not notice that some tests broke. I am sorry about this.

The commits also add downscaled tests that I used to trace illegal memory accesses for tsqr/swlq.

**Checklist**


langou (approved):

weslleyspereira:
Thanks @angsch!

PR #663, opened 2022-04-27
Fix CBLAS/src/CMakeLists.txt

mtowara:
remove spurious ../ from target_include_directories path introduced by #616


weslleyspereira:
Thanks for the fix @mtowara !

PR #664, opened 2022-04-27
Adds a test related to the fix #625

weslleyspereira:
This PR adds a test related to the fix #625.

The test:
- Verifies if `xLARRV` returns `INFO=0` when `(N.EQ.0).OR.(M.EQ.0)`.
- Verifies if `xLARRV` modifies its outputs when `(N.EQ.0).OR.(M.EQ.0)`.

The problem related on #625 was the out-of-bounds access `IBLOCK( 0 )`. To try to stress this problem does not happen anymore, I set `IWORK(1) = 100` and use the array `IWORK(2)` as `IBLOCK`.

- [ ] For now, the test was implemented only `SLARRV`. If the test makes sense, implement it on `DLARRV`, `CLARRV` and `ZLARRV`.


TESTING/EIG/schkee.F
```diff
@@ -1884,6 +1884,52 @@ PROGRAM SCHKEE
...
+     $                A(3,1), A(4,1), A(5,1), IWORK(2),
+     $                IWORK(2), A(6,1), A(8,1), 5,
+     $                IWORK(2), A(9,1), IWORK(2), INFO )
+         IF( INFO.NE.0 ) THEN
+            WRITE( NOUT, FMT = 9959 )INFO, 'M'
+         ELSE
+            DO K = 1, 8
+               IF( A(K,1) .NE. WORK(K) ) THEN
+                  WRITE( NOUT, FMT = 9958 )'M'
+                  EXIT
```
weslleyspereira, line 1910: Need to remove this EXIT

PR #674, opened 2022-05-21
Update scaling routines to return immediately when scaling with one

angsch:
**Description**

This is a proposal to add a check for alpha being equal to one to the early return criterion of the scaling routines.

It appears prudent to add an early exit to the BLAS and the LAPACK scaling routines and not rely on the caller to precede every invocation of a scaling routine with a check for one. Other BLAS routines such as HER or AXPY have a quick return based on the value of alpha already.

### DIFF
--- a/BLAS/SRC/zdscal.f
+++ b/BLAS/SRC/zdscal.f
@@ -92,17 +92,20 @@
 *
 *     .. Local Scalars ..
       INTEGER I,NINCX
+*     .. Parameters ..
+      DOUBLE PRECISION ONE
+      PARAMETER (ONE=1.0D+0)
 *     ..
 *     .. Intrinsic Functions ..
-      INTRINSIC DCMPLX
+      INTRINSIC DBLE, DCMPLX, DIMAG
 *     ..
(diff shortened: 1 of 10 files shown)


BLAS/SRC/zdscal.f
```diff
@@ -92,25 +92,28 @@ SUBROUTINE ZDSCAL(N,DA,ZX,INCX)
...
 *     ..
-      IF (N.LE.0 .OR. INCX.LE.0) RETURN
+      IF (N.LE.0 .OR. INCX.LE.0 .OR. DA.EQ.ONE) RETURN
       IF (INCX.EQ.1) THEN
 *
 *        code for increment equal to 1
 *
          DO I = 1,N
-            ZX(I) = DCMPLX(DA,0.0d0)*ZX(I)
+            ZX(I) = DCMPLX(DA*DBLE(ZX(I)),DA*DIMAG(ZX(I)))
```
langou, line 108: Nice. Thanks. Much better like this indeed. The previous line of code was ``weird``.

langou (approved):
Thanks for the PR. Looks good to me.

PR #681, opened 2022-06-20
Improve FMA usage in laqr5

angsch:
**Description**

Rearrange the application of the Householder reflector to save one instruction per dot product if fused-multiply-add is available. The proposed compute pattern is already realized by dlahqr.

### DIFF
--- a/SRC/claqr5.f
+++ b/SRC/claqr5.f
@@ -279,7 +279,7 @@
       PARAMETER          ( RZERO = 0.0e0, RONE = 1.0e0 )
 *     ..
 *     .. Local Scalars ..
-      COMPLEX            ALPHA, BETA, CDUM, REFSUM
+      COMPLEX            ALPHA, BETA, CDUM, REFSUM, T1, T2, T3
       REAL               H11, H12, H21, H22, SAFMAX, SAFMIN, SCL,
      $                   SMLNUM, TST1, TST2, ULP
       INTEGER            I2, I4, INCOL, J, JBOT, JCOL, JLEN,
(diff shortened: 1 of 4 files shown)


SRC/claqr5.f
```diff
@@ -610,25 +611,28 @@ SUBROUTINE CLAQR5( WANTT, WANTZ, KACC22, N, KTOP, KBOT, NSHFTS, S,
...
 *              ==== Perform update from left for subsequent
 *              .    column. ====
 *
-               REFSUM =  CONJG( V( 1, M ) )*( H( K+1, K+1 )
-     $                  +CONJG( V( 2, M ) )*H( K+2, K+1 )
-     $                  +CONJG( V( 3, M ) )*H( K+3, K+1 ) )
-               H( K+1, K+1 ) = H( K+1, K+1 ) - REFSUM
-               H( K+2, K+1 ) = H( K+2, K+1 ) - REFSUM*V( 2, M )
-               H( K+3, K+1 ) = H( K+3, K+1 ) - REFSUM*V( 3, M )
+               T1 = CONJG( V( 1, M ) )
```
thijssteel, line 628: Since this is only a single column, this probably doesn't actually add anything performance wise, but I like it for consistency.

thijssteel (approved):
Nice PR. Wouldn't surprise me if these small changes result in a noticeable speedup.

SRC/claqr5.f
angsch: Good spot, Thijs, I missed that one. I will upload a revision in the few days. My experiments show an improvement for small problems; less flops are good for accuracy and performance. Let me use that opportunity to say thank you for your contribution with the optimal bulge packing. It's great work :-)

PR #682, opened 2022-06-20
Add numerical tests for trevc3

angsch:
**Description**

Add some numerical tests for the so far completely untested trevc3

### DIFF
--- a/SRC/claqr5.f
+++ b/SRC/claqr5.f
@@ -533,11 +533,13 @@
 *                 .    Mth bulge. Exploit fact that first two elements
 *                 .    of row are actually zero. ====
 *
-                  REFSUM = V( 1, M )*V( 3, M )*H( K+3, K+2 )
-                  H( K+3, K   ) = -REFSUM
-                  H( K+3, K+1 ) = -REFSUM*CONJG( V( 2, M ) )
-                  H( K+3, K+2 ) = H( K+3, K+2 ) -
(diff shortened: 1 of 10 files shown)


thijssteel:
If i'm not mistaken, trevc3 is tested because it is a part of geev, and geev is thoroughly tested, so I wouldn't say that trevc3 is completely untested (correct me if i'm wrong). However, separate tests for subroutines are always welcome.

Is this perhaps an indication that you will use the skills you used to write #651 to improve trevc3?

angsch:
@thijssteel Could you be so kind and have a look?
Your suggestion to write the application of the Householder reflectors in a consistent way should be complete now; and, while I was already on it, the change is applied to the generalized problem as well. As far as I see it, that's all occurrences.


>  trevc3 is tested because it is a part of geev, and geev is thoroughly tested

The tests cover at least one more case that is not covered by geev: The backtransform is done in the tests, no via the eigenvector computation routine. There are a few more improvements that can be done in trevc3, so more tests are indeed good. :-)

thijssteel (approved):
Looks good to me, sorry for the delay

langou (approved):

martin-frbg:
Just a late and general comment from me - wouldn't it make sense to update the PR title when a PR evolves to include a rewrite of the underlying algorithm in addition to the new tests ?

Issue #217, opened 2017-12-21
syswapr docs seem wrong


mgates3:
The docs for syswapr / heswapr seem to be largely a copy-paste from sytri / hetri. In particular, the description of A seems wrong:

```
*> \param[in,out] A
*> \verbatim
*>          A is DOUBLE PRECISION array, dimension (LDA,N)
*>          On entry, the NB diagonal matrix D and the multipliers
*>          used to obtain the factor U or L as computed by DSYTRF.
*>
*>          On exit, if INFO = 0, the (symmetric) inverse of the original
*>          matrix.  If UPLO = 'U', the upper triangular part of the
*>          inverse is formed and the part of A below the diagonal is not
*>          referenced; if UPLO = 'L' the lower triangular part of the
*>          inverse is formed and the part of A above the diagonal is
*>          not referenced.
*> \endverbatim
```

From the brief description of syswapr, it is just permuting rows and cols, not computing the complete inverse as described above.

PR #685, opened 2022-06-28
Fixes

angsch:
**Description**

1. Fix an out-of-bounds write in the tests
2. Fix documentation as reported in #217. For the real routines, it would make sense to describe A as _symmetric N-by-N matrix_. For complex data types, the routine appears sensible for both symmetric and Hermitian matrices. For that reason, the commit intentionally does not state what (symmetric, Hermitian) A is. Additionally, replace the loops with calls to the BLAS to be consistent. Note that `UPLO = 'Lower'` already had a call to strided `SWAP`.

**Checklist**

### DIFF
--- a/SRC/csyswapr.f
+++ b/SRC/csyswapr.f
@@ -58,15 +58,13 @@
 *> \param[in,out] A
 *> \verbatim
 *>          A is COMPLEX array, dimension (LDA,N)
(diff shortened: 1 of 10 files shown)


langou (dismissed):

angsch:
@langou Thanks for approving and sorry for the long delay with handling the CI failure.

PR #690, opened 2022-07-14
Fix the left-looking variant of geqrf following #638

### LINKED ISSUE #673: Bug in docs/geqrf
**Description**

Here follows the bug report from the Arm Performance Libraries:

"LAPACK holds two variants of GEQRF: The default variant is SRC/dgeqrf.f. The left-looking variant lives in SRC/VARIANTS/qr/LL/dgeqrf.f. The default variant has received an update of the handling of the case when n = 0 via https://github.com/Reference-LAPACK/lapack/pull/638/. The left-looking variant was not updated in the PR.
 
We noticed this difference when googling DGEQRF to get to the netlib web documentation. The first hit is the left-looking variant - the default variant is not even listed. Could you be so kind and update the left-looking QR factorisation? Would it be possible to update the documentation in spite of not being a release? After all, the default QR factorisation has the fix already and it is a subtle change that can be missed easily."

Thanks!

I am willing to propose a PR to solve the issue. 
Related to https://github.com/Reference-LAPACK/lapack/pull/638.

weslleyspereira:
This PR applies the fixes from #638 to the left-looking variant of GEQRF.  

Closes #673.


angsch (approved):
Thanks!

langou (approved):

PR #691, opened 2022-07-14
Fix workspace query for *syevd and *heevd routines

neil-lindquist:
**Description**

The optimal workspace for `dsytrd` is `n*nb` where `nb` is the blocksize provided by `ilaenv`.  However, when `dsyevd` is called with `lwork=-1`, the requested workspace size is only `2*n + nb` (`2*n` is needed to hold the tridiagonal matrix).
So, `dsyevd` will use the unblocked `dsytrd` unless the user knows to provide extra workspace, regardless of how well `ilaenv` is tuned.

The issue is also present in `ssyevd`, `cheevd`, and `zheevd`.

This PR corrects `dsyevd` et al. to request enough workspace to use a blocked tridiagonal reduction.

**Checklist**

### DIFF
--- a/SRC/cheevd.f
+++ b/SRC/cheevd.f
@@ -284,7 +284,7 @@
                LIWMIN = 1
             END IF
             LOPT = MAX( LWMIN, N +
-     $                  ILAENV( 1, 'CHETRD', UPLO, N, -1, -1, -1 ) )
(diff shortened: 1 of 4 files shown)


weslleyspereira (approved):
Thanks  @neil-lindquist!

PR #696, opened 2022-08-05
Time check flag propagation and default value.

ACSimon33:
**Description**

We had an issue where we need architecture flags for the Nagfor compiler to run on Apple Silicon. Those flags are defined from the outside build harness via:

`-DCMAKE_Fortran_FLAGS="-Wc,-target,arm64-apple-macos11"`
`-DCMAKE_EXE_LINKER_FLAGS="-Wl,-arch,arm64"`

The problem was that these flags weren't propagated to the time check subproject, resulting in all of the tests failing including `second_NONE.f`. That alone wouldn't be a problem if the default value for TIME_FUNC was set correctly, which wasn't the case.

I added the flag propagation as well as the correct default value for TIME_FUNC. I also added `-DCMAKE_VERBOSE_MAKEFILE=ON` which will add a bit more output to the log files. It won't add anything to stdout or stderr.

**Checklist**


weslleyspereira (approved):
Thanks!

ACSimon33:
Can we merge this or should I change something?

langou:
> Can we merge this or should I change something?

Thanks for ping.

PR #698, opened 2022-08-08
Use normwise criterion for inf eigenvalues in QZ

thijssteel:
As described in "A new deflation criterion for the QZ algorithm" T. Steel, R. Vandebril and J. Langou (https://arxiv.org/abs/2208.02057)
When testing for infinite eigenvalues in the QZ algorithm, the normwise criterion performs better than the elementwise one. 

This PR changes the infinite eigenvalue check to the normwise criterion, but keeps the elementwise check for the finite eigenvalues.

### DIFF
--- a/SRC/chgeqz.f
+++ b/SRC/chgeqz.f
@@ -523,9 +523,7 @@
             END IF
          END IF
 *
-         IF( ABS( T( ILAST, ILAST ) ).LE.MAX( SAFMIN, ULP*( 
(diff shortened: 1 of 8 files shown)


angsch:
Looks good to me :-)

With your changes `tempr` becomes an unused variable in `[cz]laqz0`, if you want to remove it.

langou (approved):

PR #700, opened 2022-08-08
Fixed format warnings in 64 bit integer builds.

ACSimon33:
**Description**

There were print format warnings when 64-bit integers are enabled. The `fprintf` statements were using "%d" or "%i" as integer conversion specifiers which are the ones for 32-bit integers. 64-bit are "%lld" (MacOS) or "%ld" (Linux):

```shell
/Users/simonm/git/lapack/CBLAS/src/cblas_xerbla.c:66:69: warning: format specifies type 'int' but the argument has type 'int64_t' (aka 'long long') [-Wformat]
      fprintf(stderr, "Parameter %d to routine %s was incorrect\n", info, rout);
                                 ~~                                 ^~~~
                                 %lld
1 warning generated.
```

I added the preprocessor defines CBLAS_IFMT and LAPACK_IFMT which use the correct specifier (platform independent via "inttypes.h"). We could also put `#ifdefs` around each `fprintf` statement which is a bit tedious IMHO.

**Checklist**

### DIFF
--- a/LAPACKE/include/lapack.h
+++ b/LAPACKE/include/lapack.h
@@ -84,7 +84,9 @@ extern "C" {
 #define lapack_int     int32_t
 #endif
 
+#ifndef LAPACK_IFMT
 #define LAPACK_IFMT    PRId32
+#endif
 
 #ifndef lapack_logical
 #define lapack_logical lapack_int
--- a/LAPACKE/include/lapack.h
+++ b/LAPACKE/include/lapack.h
@@ -81,15 +81,26 @@ extern "C" {
 
 /*----------------------------------------------------------------------------*/
 #ifndef lapack_int
-#define lapack_int     int32_t
+#if defined(LAPACK_ILP64)
+#define lapack_int        int64_t
+#else
+#define lapack_int        int32_t
+#endif
 #endif
 
+/*
+ * Integer format string
+ */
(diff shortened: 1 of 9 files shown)


LAPACKE/include/lapack.h
```diff
@@ -80,7 +81,11 @@ extern "C" {
 
 /*----------------------------------------------------------------------------*/
 #ifndef lapack_int
-#define lapack_int     int
+#define lapack_int     int32_t
```
weslleyspereira, line 84: I think we should use:

```C
#ifndef lapack_int
#if defined(LAPACK_ILP64)
#define lapack_int              int64_t
#else
#define lapack_int              int32_t
#endif
#endif
```

for `lapack_int` as in [LAPACKE/include/lapacke_config.h](https://github.com/Reference-LAPACK/lapack/pull/700/files#diff-7db5a114acf1e298d576cd7998665526f604cfd7d0b50ce8121dd321507abee1)

ACSimon33: Yes! That way users don't have to define HAVE_LAPACK_CONFIG_H anymore if they want a 64-bit integer build with the standard config.

weslleyspereira (approved):
Thanks!

Issue #695, opened 2022-08-02
Increased number of test failures in CI


angsch:
HEAD has an increased number of numerical failures compared to ca. 1/2 year ago. Note that the errors do not make CI mark the run as failed.

Now it is 
```
 		Processing LAPACK Testing output found in the TESTING directory
SUMMARY             	nb test run 	numerical error   	other error  
================   	===========	=================	================  
REAL             	1317267		12	(0.001%)	0	(0.000%)	
DOUBLE PRECISION	1318089		39	(0.003%)	0	(0.000%)	
COMPLEX          	777619		0	(0.000%)	0	(0.000%)	
COMPLEX16         	777594		1	(0.000%)	0	(0.000%)	
```

`Testing DOUBLE PRECISION CS-Decomposition-routines-dcsd.out  CSD:     39 out of    600 tests failed to pass the threshold
 failing to pass the threshold: 39`
`Testing REAL              CS-Decomposition-routines-scsd.out  CSD:     12 out of    600 tests failed to pass the threshold
 failing to pass the threshold: 12
`

It used to be a single failure for `COMPLEX16`.

martin-frbg:
Still bisecting, but it looks like all was good on Jun 30 when 686 (cmake nagfor support) got merged, and the 12 errors in REAL (but nothing else yet) crept in by Jul 5

martin-frbg:
bisected the 12 errors in REAL to 
commit bdcd890a185482119c52dd7acd9a702f0cad782a
Author: Christoph Conrads <gitlab@christoph-conrads.name>
Date:   Mon Oct 25 18:42:55 2021 +0000

    SORCSD: fix documentation on matrix dimensions

 SRC/sorbdb2.f    | 4 ++--
 SRC/sorbdb4.f    | 2 +-
 SRC/sorcsd2by1.f | 2 +-
 3 files changed, 4 insertions(+), 4 deletions(-)

the date looks weird, but this is actually a small part of PR #647 which got merged on July 15 and the change in file
SRC/sorcsd2by1.f actually affected code rather than documentation. It seems likely that at least some of the errors in DOUBLE PRECISION may have been caused by related changes in that PR, but I have not verified that yet (close to midnight here)

martin-frbg:
Reverting
```
-           CALL SCOPY( M-P, WORK(IORBDB+P), 1, U2, 1 )
+           CALL SCOPY( M-Q, WORK(IORBDB+P), 1, U2, 1 )
```
[...]

PR #702, opened 2022-08-09
Fix type precision in dorbdb6 and zunbdb6

mjacobse:
**Description**

Building shared libraries on current master https://github.com/Reference-LAPACK/lapack/commit/984dcc059ca82e33a4d1c5326f7cf3a8f07dfaf0 with `-flto -Wlto-type-mismatch` generates the warnings
```
lapack/SRC/dorbdb6.f:252: warning: type of ‘dlassq’ does not match original declaration [-Wlto-type-mismatch]
  252 |       CALL DLASSQ( M1, X1, INCX1, SCL, SSQ )
      | 
lapack/SRC/dlassq.f90:136: note: ‘dlassq’ was previously declared here
  136 | subroutine DLASSQ( n, x, incx, scl, sumsq )
      | 
lapack/SRC/dlassq.f90:136: note: code may be misoptimized unless ‘-fno-strict-aliasing’ is used
```
and
```
lapack/SRC/zunbdb6.f:224: warning: type of ‘dlamch’ does not match original declaration [-Wlto-type-mismatch]
  224 |       EPS = DLAMCH( 'Precision' )
      | 
lapack/INSTALL/dlamch.f:68: note: return value type mismatch
   68 |       DOUBLE PRECISION FUNCTION DLAMCH( CMACH )
      | 
lapack/INSTALL/dlamch.f:68: note: type ‘double’ should match type ‘float’
lapack/INSTALL/dlamch.f:68: note: ‘dlamch’ was previously declared here
lapack/INSTALL/dlamch.f:68: note: code may be misoptimized unless ‘-fno-strict-aliasing’ is used
```
showing that these calls to `DLASSQ` and `DLAMCH` are not correct; they are using wrong argument and return types respectively.

To me it looks like a simple oversight of not replacing `REAL` with `DOUBLE PRECISION` in two places. And that indeed gets rid of the type mismatches. Hence this pull request.

If the choice of `REAL` is intentional in these places, then I believe the calls should still be fixed, but in a different way. Perhaps casting to `DOUBLE PRECISION` before passing to `DLASSQ` and using `SLAMCH` instead of `DLAMCH`?

**Checklist**

### DIFF
--- a/SRC/dorbdb6.f
+++ b/SRC/dorbdb6.f
@@ -181,7 +181,7 @@
 *     ..
 *     .. Local Scalars ..
       INTEGER            I, IX
-      REAL               EPS, NORM, NORM_NEW, SCL, SSQ
+      DOUBLE PRECISION   EPS, NORM, NORM_NEW, SCL, SSQ
 *     ..
 *     .. External Functions ..
       DOUBLE PRECISION   DLAMCH
--- a/SRC/zunbdb6.f
+++ b/SRC/zunbdb6.f
@@ -185,7 +185,7 @@
       DOUBLE PRECISION   EPS, NORM, NORM_NEW, SCL, SSQ
 *     ..
 *     .. External Functions ..
-      REAL               DLAMCH
+      DOUBLE PRECISION   DLAMCH
 *     ..
 *     .. External Subroutines ..
       EXTERNAL           ZGEMV, ZLASSQ, XERBLA


weslleyspereira:
Thanks! Maybe this is related to https://github.com/Reference-LAPACK/lapack/issues/695.

weslleyspereira (approved):

mjacobse:
Maybe it could be worth to add a CI build with link time optimization and `-Werror=lto-type-mismatch` to avoid similar bugs in the future? Adding explicit interfaces instead of the `EXTERNAL` declarations for external functions to detect this would probably be even better, but also much more work. For LTO you simply pass the flag and can start detecting (some of?) these errors, albeit only at link time.

weslleyspereira:
Thanks for the suggestion, @mjacobse. I spend 1 hour to find all type mismatches and implicit type casts in the library. See #703.

PR #706, opened 2022-08-10
Fix type mismatches in function calls of testing code

mjacobse:
**Description**

Follow up on #702 which missed the type mismatch errors when compiling with `-flto -Wlto-type-mismatch` in the test code.

**Checklist**

- [N/A ] The documentation has been updated.


weslleyspereira (approved):
Looks great to me!

PR #713, opened 2022-08-22
Fixed the test dependencies of the test summary

ACSimon33:
**Description**

A small fix of the test dependencies. The `if(PYTHONINTERP_FOUND)` was from the time we used `find_package(PythonInterp 2.7 QUIET)` instead of the `find_program(PYTHON_EXECUTABLE python3)` (see a51f8ba).

I also changed the comment since it sounded like the test summary would run during the configuring stage.

**Checklist**


weslleyspereira (approved):
Looks great to me. Thanks!

Issue #473, opened 2021-01-12
Should there be CBLAS interfaces to CSROT and ZDROT ?


martin-frbg:
In https://github.com/xianyi/OpenBLAS/issues/3063 it was noticed that CBLAS interfaces for CSROT and ZDROT are/were missing from OpenBLAS' cblas.h . 
Given that the BLAS versions of these functions appear to have been in the reference implementation since at least 3.2.1 (the oldest I appear to have available in my lair, which also predates inclusion of the CBLAS), I wonder if this is an oversight here as well, or was it a conscious decision ? I do note that MKL and Accelerate both carry the CBLAS interfaces, and Atlas added them sometime in 2008 with a mailing list comment to the effect that they were non-standard.
EDIT: same for cblas_crotg and cblas_zrotg

PR #721, opened 2022-09-29
Fix 2 CBLAS issues

angsch:
**Description**

Part 1: Add the cblas interfaces related to rot that were reported as missing in #473 
Part 2: Fix a linker error coming from `cblas.h` defining `cblas_[ds]cabs1`, but not having an implementation.

**Checklist**

### DIFF
--- a/CBLAS/include/cblas.h
+++ b/CBLAS/include/cblas.h
@@ -143,19 +143,13 @@ void cblas_zaxpy(const CBLAS_INT N, const void *alpha, const void *X,
 /*
  * Routines with S and D prefix only
  */
-void cblas_srotg(float *a, float *b, float *c, float *s);
 void cblas_srotmg(float *d1, float *d2, float *b1, const float b2, float *P);
-void cblas_srot(const CBLAS_INT N, float *X, const CBLAS_INT incX,
-                float *Y, const CBLAS_INT incY, const float c, const float s);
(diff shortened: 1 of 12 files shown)


angsch:
@martin-frbg Could you be so kind and have a look at the cblas interfaces?

I am surprised that BLAS-1 routines such as `i[c,z]amax` call into `[d,s]cabs1` instead of following LAPACK's approach where `cabs1` is reimplemented as a statement function in every relevant file. In particular for `cblas_[d,s]cabs1`, it feels excessive to compute  `|Re(.)| + |Im(.)|` through 1 auxiliary subroutine and 1 function call. Is the replacement of function calls of cabs1 in favor of an inline definition something that should be addressed in the reference BLAS?

langou:
> I am surprised that BLAS-1 routines such as i[c,z]amax call into [d,s]cabs1 instead of following LAPACK's approach where cabs1 is reimplemented as a statement function in every relevant file. In particular for cblas_[d,s]cabs1, it feels excessive to compute |Re(.)| + |Im(.)| through 1 auxiliary subroutine and 1 function call. Is the replacement of function calls of cabs1 in favor of an inline definition something that should be addressed in the reference BLAS?

Good point, but that's kind of another thread. It would be good to have inline functions for these. I do not know whether there was a purposeful intent behind not using intrinsic.

weslleyspereira:
Thanks @angsch!!

PR #725, opened 2022-10-06
Implemented a LAPACKE_?langb interface

### LINKED ISSUE #722: Why is there no LAPACKE_?langb interface?
Hi all,

I wondered why there isn't a LAPACKE interface for `?langb`. If there is no reason against it I'm happy to implement it and submit a pull request.

ACSimon33:
**Description**

Following up on #722, I've implemented the  `LAPACKE_?langb` interfaces. I've used the same trick as in `LAPACKE_?lange` where you interpret the matrix as transposed in row-major layout. Although the matrix is in band layout here, it works because it is a square matrix (correct me if I'm wrong). `kl` and `ku` are swapped in the row-major case due to the implicit transposition.

Does the documentation update itself via oxygen?

Closes #722 

**Checklist**

### DIFF
--- a/LAPACKE/include/lapacke.h
+++ b/LAPACKE/include/lapacke.h
@@ -2313,6 +2313,19 @@ lapack_int LAPACKE_zlagge( int matrix_layout, lapack_int m, lapack_int n,
 float LAPACKE_slamch( char cmach );
 double LAPACKE_dlamch( char cmach );
 
+float LAPACKE_slangb( int matrix_layout, char norm, lapack_int n,
+                      lapack_int kl, lapack_int ku, const float* ab,
+                      lapack_int ldab );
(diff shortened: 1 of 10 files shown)


langou:
> Following up on https://github.com/Reference-LAPACK/lapack/issues/722, I've implemented the LAPACKE_?langb interfaces. I've used the same trick as in LAPACKE_?lange where you interpret the matrix as transposed in row-major layout. Although the matrix is in band layout here, it works because it is a square matrix (correct me if I'm wrong). kl and ku are swapped in the row-major case due to the implicit transposition.

Well done!

> Does the documentation update itself via [d]oxygen?

Good question. No, it does not update itself. When we release the new release (probably 3.11), then we (@julielangou) will update the documentation. We do not have any automagic way to update the doxygen documentation from the GitHub repo at each merge.

PR #728, opened 2022-10-12
Fix lower triangular band matrix generation

### LINKED ISSUE #343: TESTING: parameters set incorrectly for triangular band matrices in DLATB4
DLATB4 is called from DLATTB with IMAT negative for lower triangular matrices:
```
      IF( UPPER ) THEN
         CALL DLATB4( PATH, IMAT, N, N, TYPE, KL, KU, ANORM, MODE,
     $                CNDNUM, DIST )
....
      ELSE
         CALL DLATB4( PATH, -IMAT, N, N, TYPE, KL, KU, ANORM, MODE,
     $                CNDNUM, DIST )
...
      END IF
```
However, DLATB4 does not consider the sign if IMAT for triangular band matrices:
```
         IF( IMAT.EQ.2 .OR. IMAT.EQ.8 ) THEN
            CNDNUM = BADC1
         ELSE IF( IMAT.EQ.3 .OR. IMAT.EQ.9 ) THEN
            CNDNUM = BADC2
         ELSE
            CNDNUM = TWO
         END IF
*
         IF( IMAT.EQ.4 ) THEN
            ANORM = SMALL
         ELSE IF( IMAT.EQ.5 ) THEN
            ANORM = LARGE
         ELSE
            ANORM = ONE
         END IF
```
Consequently, lower triangular band matrices are tested only with `CNDNUM=2` and `ANORM=1`. This is in contrast to triangular (packed) matrices in DLATB4 where the corresponding code uses `MAT = ABS( IMAT )` instead of `IMAT`.

angsch:
Closes #343

**Description**

**Checklist**

- [X ] If the PR solves a specific issue, it is set to be closed on merge.

### DIFF
--- a/TESTING/LIN/clatb4.f
+++ b/TESTING/LIN/clatb4.f
@@ -517,17 +517,18 @@
 *
 *        Set the norm and condition number.
 *
-         IF( IMAT.EQ.2 .OR. IMAT.EQ.8 ) THEN
+         MAT = ABS( IMAT )
+         IF( MAT.EQ.2 .OR. MAT.EQ.8 ) THEN
             CNDNUM = BADC1
-         ELSE IF( IMAT.EQ.3 .OR. IMAT.EQ.9 ) THEN
+         ELSE IF( MAT.EQ.3 .OR. MAT.EQ.9 ) THEN
             CNDNUM = BADC2
          ELSE
             CNDNUM = TWO
          END IF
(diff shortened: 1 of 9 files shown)


weslleyspereira (approved):

weslleyspereira:
Fixes look great to me! Thanks!

PR #738, opened 2022-10-27
NaN check for trapezoidal routines.

### LINKED ISSUE #737: NaN check in LAPACKE_?lantr isn't correct for m != n
**Description**

The NaN check in `LAPACKE_?lantr` isn't checking the entire matrix for non-square matrices. The check is implemented like this:
```
if( LAPACKE_ztr_nancheck( matrix_layout, uplo, diag, MIN(m,n), a, lda ) ) {
    return -7;
}
```

There are two solutions to this. Either we add another check for the part which isn't checked - that could look like this for example:
```
if( LAPACKE_ztr_nancheck( matrix_layout, uplo, diag, MIN(m,n), a, lda ) ) {
    return -7;
}
if( LAPACKE_lsame( uplo, 'u' ) && n > m) {
    lapack_int offset = m * ((matrix_layout == LAPACK_COL_MAJOR) ? lda : 1);
    if( LAPACKE_zge_nancheck( matrix_layout, m, n - m, &a[offset], lda ) ) {
        return -7;
    }
}
if( LAPACKE_lsame( uplo, 'l' ) && m > n) {
    lapack_int offset = n * ((matrix_layout == LAPACK_ROW_MAJOR) ? lda : 1);
    if( LAPACKE_zge_nancheck( matrix_layout, m - n, n, &a[offset], lda ) ) {
        return -7;
    }
}
```

The cleaner solution in my opinion would be to change `LAPACKE_?tr_nancheck` to accept non-square matrices. That should be fairly simple by just adding a second size parameter and adjusting the loops. The NaN check in `LAPACKE_zlarfb` would benefit from this as well since there we have the same problem (which is already addressed via the first strategy).

The only downside of changing `LAPACKE_?tr_nancheck` is that there would be a lot of files that need fixes due to the interface change (64 files).

I'd like to know which strategy you would prefer.

**Checklist**

ACSimon33:
**Description**

Following up on #737 I've implemented a NaN check for trapezoidal matrices. I've added a`direct` argument to support a whole bunch of trapezoidal matrices (e.g. those which occur in `xLARFB`):

```
Diagonals 'F' (front / forward) and 'B' (back / backward):

        A = ( F       )           A = ( F     B       )
            (    F    )               (    F     B    )
            ( B     F )               (       F     B )
            (    B    )
            (       B )

      direct = 'F', uplo = 'L':

        A = ( *       )           A = ( *             )
            ( *  *    )               ( *  *          )
            ( *  *  * )               ( *  *  *       )
            ( *  *  * )
            ( *  *  * )

      direct = 'F', uplo = 'U':

        A = ( *  *  * )           A = ( *  *  *  *  * )
            (    *  * )               (    *  *  *  * )
            (       * )               (       *  *  * )
            (         )
            (         )

      direct = 'B', uplo = 'L':

        A = (         )           A = ( *  *  *       )
            (         )               ( *  *  *  *    )
            ( *       )               ( *  *  *  *  * )
            ( *  *    )
            ( *  *  * )

      direct = 'B', uplo = 'U':

        A = ( *  *  * )           A = (       *  *  * )
            ( *  *  * )               (          *  * )
            ( *  *  * )               (             * )
            (    *  * )
            (       * )
```

For now, I've only applied the new NaN check to  `LAPACKE_?lantr`. But the NaN check is designed to also work for `LAPACKE_?larfb` and potential other routines. @langou I'll investigate those later since I want first confirm with you that the interface that I've chosen is alright.

Additionally, I fixed some missing symbols in the Makefile which I forgot to add in my last PR for #722.

**Checklist**

- [x] Fixed missing symbols for `LAPACKE_?langb` in Makefile.
- [x] Implemented NaN check for trapezoidal matrices.
- [x] Fixed NaN check in `LAPACKE_?lantr`.
- [ ] Apply trapezoidal NaN check in `LAPACKE_?larfb` and maybe other routines. 
- [x] Closes #737

### DIFF
--- a/LAPACKE/include/lapacke_utils.h
+++ b/LAPACKE/include/lapacke_utils.h
@@ -376,6 +376,10 @@ lapack_logical LAPACKE_ctr_nancheck( int matrix_layout, char uplo, char diag,
                                       lapack_int n,
                                       const lapack_complex_float *a,
                                       lapack_int lda );
+lapack_logical LAPACKE_ctz_nancheck( int matrix_layout, char direct, char uplo,
+                                     char diag, lapack_int m, lapack_int n,
+                                     const lapack_complex_float *a,
+                                     lapack_int lda );
 
 lapack_logical LAPACKE_dgb_nancheck( int matrix_layout, lapack_int m,
                                       lapack_int n, lapack_int kl,
@@ -440,6 +444,9 @@ lapack_logical LAPACKE_dtr_nancheck( int matrix_layout, char uplo, char diag,
                                       lapack_int n,
                                       const double *a,
                                       lapack_int lda );
+lapack_logical LAPACKE_dtz_nancheck( int matrix_layout, char direct, char uplo,
+                                     char diag, lapack_int m, lapack_int n,
+                                     const double *a, lapack_int lda );
 
 lapack_logical LAPACKE_sgb_nancheck( int matrix_layout, lapack_int m,
                                       lapack_int n, lapack_int kl,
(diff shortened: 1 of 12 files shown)


langou (approved):

langou:
Thanks @ACSimon33. This is great. I like the pictures of the matrices occurring in LARFB, very nicely done.

PR #739, opened 2022-10-29
added new routines [S,D,C,Z]GELST and test code for the least squares

scr2016:
**Description**
added new routines [S,D,C,Z]GELST and test code for the least squares
squares algorithm that solves overdetermined or underdetermined systems for GE matrices using QR or LQ factorization with compact WY representation of Q inside the routine.

*GELST is similar to *GELS, but unlike *GELS does not compute triangular blocks twice, i.e. runs faster.

modified:   SRC/Makefile
new file:   SRC/cgelst.f
new file:   SRC/dgelst.f
new file:   SRC/sgelst.f
new file:   SRC/zgelst.f
modified:   TESTING/LIN/alahd.f
modified:   TESTING/LIN/cdrvls.f
modified:   TESTING/LIN/cerrls.f
modified:   TESTING/LIN/ddrvls.f
modified:   TESTING/LIN/derrls.f
modified:   TESTING/LIN/sdrvls.f
modified:   TESTING/LIN/serrls.f
modified:   TESTING/LIN/zdrvls.f

### DIFF
--- a/SRC/CMakeLists.txt
+++ b/SRC/CMakeLists.txt
@@ -86,7 +86,7 @@ set(SLASRC
    sgbsvx.f sgbtf2.f sgbtrf.f sgbtrs.f sgebak.f sgebal.f sgebd2.f
    sgebrd.f sgecon.f sgeequ.f sgees.f  sgeesx.f sgeev.f  sgeevx.f
    sgehd2.f sgehrd.f sgelq2.f sgelqf.f
-   sgels.f  sgelsd.f sgelss.f sgelsy.f sgeql2.f sgeqlf.f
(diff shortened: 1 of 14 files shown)


angsch:
@scr2016 Is there a particular reason why your changes are a separate routine and not an upgrade of `GELS`? Your changes appear fully compatible with `GELS`and could be a great modernization that could be widely picked up by, say, octave or Julia, which likely call `GELS` as the de facto standard LS solver. 

Do you have some performance results of how your algorithm behaves with `PBLAS` compared to the old `GELS` solver? The recursive QR factorization produces quite a few degenerated matrices.

dbielich:
I see edits to all `TESTING/LIN/Xerrls.f` files except `TESTING/LIN/zserrls.f`. 

So there is no way to check any errors from the double complex code currently.

langou (approved):

PR #742, opened 2022-10-30
Lapacke x larfb nancheck

ACSimon33:
**Description**

Following up on #738 I've applied the trapezoidal NaN check to `LAPACKE_?larfb`. I've also implemented transposition routines for trapezoidal routines and applied them similarly to `LAPACKE_?larfb_work`.

There was also a bug in the NaN check that I implemented earlier where I checked if `uplo == 'b'` which is obviously wrong. That's fixed as well.

### DIFF
--- a/LAPACKE/src/lapacke_clarfb.c
+++ b/LAPACKE/src/lapacke_clarfb.c
@@ -42,7 +42,7 @@ lapack_int LAPACKE_clarfb( int matrix_layout, char side, char trans, char direct
     lapack_int info = 0;
     lapack_int ldwork;
     lapack_complex_float* work = NULL;
-    lapack_int ncols_v, nrows_v;
+    lapack_int nrows_v, ncols_v;
     lapack_logical left, col, forward;
     char uplo;
     if( matrix_layout != LAPACK_COL_MAJOR && matrix_layout != LAPACK_ROW_MAJOR ) {
@@ -56,8 +56,8 @@ lapack_int LAPACKE_clarfb( int matrix_layout, char side, char trans, char direct
         col = LAPACKE_lsame( storev, 'c' );
         forward = LAPACKE_lsame( direct, 'f' );
 
-        ncols_v = ( !col && left ) ? m : ( ( !col && !left ) ? n : ( col ? k : 1 ) );
         nrows_v = ( col && left ) ? m : ( ( col && !left ) ? n : ( !col ? k : 1) );
+        ncols_v = ( !col && left ) ? m : ( ( !col && !left ) ? n : ( col ? k : 1 ) );
         uplo = ( ( left && col ) || !( left || col ) ) ? 'l' : 'u';
 
         if( !forward && ( col && k > nrows_v ) || ( !col && k > ncols_v )) {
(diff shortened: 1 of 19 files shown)


langou (approved):

LAPACKE/src/lapacke_clarfb.c
```diff
@@ -42,67 +42,37 @@ lapack_int LAPACKE_clarfb( int matrix_layout, char side, char trans, char direct
...
-        nrows_v =   ( LAPACKE_lsame( storev, 'c' ) && LAPACKE_lsame( side, 'l' ) ) ? m :
-                  ( ( LAPACKE_lsame( storev, 'c' ) && LAPACKE_lsame( side, 'r' ) ) ? n :
-                    ( LAPACKE_lsame( storev, 'r' ) ? k : 1) );
-        if( LAPACKE_cge_nancheck( matrix_layout, m, n, c, ldc ) ) {
-            return -13;
+        nrows_v = ( col && left ) ? m : ( ( col && !left ) ? n : ( !col ? k : 1) );
+        ncols_v = ( !col && left ) ? m : ( ( !col && !left ) ? n : ( col ? k : 1 ) );
+        uplo = ( ( left && col ) || !( left || col ) ) ? 'l' : 'u';
+
+        if( !forward && ( col && k > nrows_v ) || ( !col && k > ncols_v )) {
```
weslleyspereira, line 63: Is `if( !forward && ( col && k > nrows_v ) || ( !col && k > ncols_v ))` supposed to be `if(( col && k > nrows_v ) || ( !col && k > ncols_v ))` ? I didn't get the test `!forward`.

In any case, I arrived to this line of code because we should at least add parenthesis to ensure correct behavior. The conditional currently looks like `if( A && B || C )`.

weslleyspereira: I am marking @ACSimon33 and @langou just to make sure you both see my comment.

ACSimon33: Looking at the code before my PR, there was the same behavior, with a nested if statement:

```
       } else if( LAPACKE_lsame( storev, 'c' ) && LAPACKE_lsame( direct, 'b' ) ) {
            if( k > nrows_v ) {
                LAPACKE_xerbla( "LAPACKE_clarfb", -8 );
                return -8;
            }
            if( LAPACKE_ctr_nancheck( matrix_layout, 'u', 'u', k,
                                      &v[(nrows_v-k)*lrv], ldv ) )
                return -9;
            if( LAPACKE_cge_nancheck( matrix_layout, nrows_v-k, ncols_v, v, ldv ) )
                return -9;
        }
```

The test was there to prevent segfaults when accessing `&v[(nrows_v-k)*lrv]`. These tests were only in the if branches with `direct == b` (`!forward`). So in order to mirror the old behavior we should put parenthesis around the OR statement. The question remains if there is any reason to allow `k > nrows_v` (or `k > ncols_v`) if `direct == f`.

I think we probably can remove the `!forward` test.

langou: Related to this is #780. Now I am reading what @ACSimon33 wrote and it is written "n order to mirror the old behavior we should put parenthesis around the OR statement." Shoo. We just did the reverse and put parenthesis around the AND statement.

In any case, the comment "The question remains if there is any reason to allow `k > nrows_v` (or `k > ncols_v`) if `direct == f`. I think we probably can remove the `!forward` test.

We'll work on this and update the thread. The [<T>LAPACK](https://github.com/tlapack/tlapack) team is having a look at larfb these days so hopefully we can also get some input from them.

PR #749, opened 2022-11-07
Reciprocal scaling with less accumulation errors

weslleyspereira:
Reciprocal scaling of a vector is currently implemented in LAPACK as `(1/r)*x` instead of `x/r`.

1. In general, the former formula will have higher accumulation error. This is because it needs 2 floating-point operations per entry.
2. The former needs scaling procedure to handle small and large `r`. The latter doesn't.
3. One could argue that the former algorithm relies on BLAS SCAL to obtain higher performance. However, some vendors (MKL) already implement RSCL. I think we should have the more accurate algorithm in LAPACK for this specific case.

If so, this is the PR that implements the reciprocal scaling using a simple division.

### DIFF
--- a/SRC/csrscl.f
+++ b/SRC/csrscl.f
@@ -18,14 +18,14 @@
 *  Definition:
 *  ===========
 *
-*       SUBROUTINE CSRSCL( N, SA, SX, INCX )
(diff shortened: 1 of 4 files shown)


weslleyspereira:
Related to https://github.com/tlapack/tlapack/pull/178.

PR #755, opened 2022-11-17
Fix bug in new test case for LAPACK 3.11 SLATRS3

dklyuchinskiy:
**Fixed bug in new test case for LAPACK 3.11 SLATRS3**

1. In new test case 10 from `xlintsts_stest` wrong indexation in array B leads to failure of the test.
2. Also added use of LSAME in function iparam2stage 

**Checklist**


martin-frbg (approved):

weslleyspereira:
Thanks you, @dklyuchinskiy !!

PR #760, opened 2022-11-18
Windows build fix

ACSimon33:
**Description**

I encountered a build failure on Windows due to a missing symbol `DGESVQ` which doesn't exist. It was a typo in the test framework and should be `DGESVDQ`.

Also, I removed some unnecessary return statements in the trapezoidal transposition functions which I introduced by mistake in #742.


langou:
Thanks @ACSimon33. Good that the windows compiled is more picky on this EXTERNAL than others. Thanks for the two fixes.

langou (approved):

PR #764, opened 2022-11-22
Fix SLATRS3 and CLATRS3 tests

bartoldeman:
Some SLATRS3 tests were failing with OpenBLAS where the RHS has a BIGNUM (=Infinity in this test) component. OpenBLAS SSCAL, when multiplying with 0, will turn this component into 0, unlike reference BLAS which turns it into NaN. While NaN looks more correct it still did not explain why DLATRS3 tests were succeeding where DSCAL has the same difference between reference and OpenBLAS.

Diffing the test code unmasked a few bugs, which makes the test succeed and also unmasked a bug in the CLATRS3 test.

The ZLATRS3 and DLATRS3 tests look correct to me.

@angsch can you please review as test author (#651)?

### DIFF
--- a/TESTING/LIN/cchktr.f
+++ b/TESTING/LIN/cchktr.f
@@ -541,7 +541,7 @@
 *
                   SRNAMT = 'CLATRS3'
                   CALL CCOPY( N, X, 1, B, 1 )
-                  CALL CCOPY( N, X, 1, B, 1 )
+                  CALL CCOPY( N, X, 1, B( N+1 ), 1 )
                   CALL CSCAL( N, BIGNUM, B( N+1 ), 1 )
                   CALL CLATRS3( UPLO, TRANS, DIAG, 'N', N, 2, A, LDA,
      $                          B, MAX(1, N), SCALE3, RWORK, WORK, NMAX,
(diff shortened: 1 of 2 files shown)


bartoldeman:
I just see #755 overlaps this with one chunk for the SLATRS3 test, however misses the other fixes here (mostly for CLATRS3)

TESTING/LIN/cchktr.f
```diff
@@ -541,7 +541,7 @@ SUBROUTINE CCHKTR( DOTYPE, NN, NVAL, NNB, NBVAL, NNS, NSVAL,
 *
                   SRNAMT = 'CLATRS3'
                   CALL CCOPY( N, X, 1, B, 1 )
-                  CALL CCOPY( N, X, 1, B, 1 )
+                  CALL CCOPY( N, X, 1, B( N+1 ), 1 )
```
angsch, line 544: Great spot!

angsch:
Thanks, looks good. :+1:

bartoldeman:
> The changes look good to me too. Thanks, @bartoldeman !

@weslleyspereira Just wondering why this has been approved but not merged? Is there an issue somewhere?

PR #768, opened 2022-11-28
fix segfault problem when executing LAPACKE_*geqrt on row major #766

### LINKED ISSUE #766: The matrix size of LAPACKE_sge_trans in lapacke_sgeqrt_work.c seems to be not correct
**Description**

When LAPACKE_sgeqrt is executed with row major, the following code is executed to convert the result obtained with column major to row major.

https://github.com/Reference-LAPACK/lapack/blob/77fac7e8358b9e350281ea2c4a2b6ea92ad2a602/LAPACKE/src/lapacke_sgeqrt_work.c#L82-L83

Here, `ldt * MIN(m,n)` should be the number of elements in `t_t` (= the number of elements in `t`), but this expression takes more values when ~`ldt > nb`~ `nb > n`.

This causes a segmentation fault when `LAPACKE_sgeqrt` is executed with row major and `ldt > n`.
Here is an example that causes the problem:
```c
#include <stdio.h>
#include <time.h>
#include <lapacke.h>

int main(int argc, char *argv[])
{
    srandom(42);
    const int N = 16;
    const int NB = 4;
    const int LDA = N;
    const int LDT = 100 * N; // set the value so that LDT > N holds.

    // allocate A and T
    float *A = (float *)malloc(N * LDA * sizeof(float));
    float *T = (float *)malloc(N * LDT * sizeof(float));

    // initialize A
    for (int i = 0; i < N; i++)
    {
        for (int j = 0; j < N; j++)
        {
            A[i * LDA + j] = (float)random() / (float)RAND_MAX;
        }
    }

    // call sgeqrt
    LAPACKE_sgeqrt(LAPACK_ROW_MAJOR, N, N, NB, A, LDA, T, LDT);

    // print results
    for (int i = 0; i < N; i++)
    {
        for (int j = 0; j < N; j++)
        {
            printf("% 8f ", A[i * LDA + j]);
        }
        printf("\n");
    }
    printf("\n");
    for (int i = 0; i < NB; i++)
    {
        for (int j = 0; j < N; j++)
        {
            printf("% 8f ", T[i * LDT + j]);
        }
        printf("\n");
    }

    // free memory
    free(A);
    free(T);
    return 0;
}
```

Shouldn't this line look like this?
```c
LAPACKE_sge_trans(LAPACK_COL_MAJOR, nb, MIN(m,n), t_t, ldt_t, t, ldt );
```

I'm thinking that other `LAPACKE_*geqrt` functions may contain similar problems.

**Checklist**

sbite0138:
**Description**
This pull request fixes a segmentation fault that occurred when LAPACKE_*geqrt was executed on row major.
close #766.

**Checklist**

- [x] The documentation has been updated.(N/A)

### DIFF
--- a/LAPACKE/src/lapacke_cgeqrt_work.c
+++ b/LAPACKE/src/lapacke_cgeqrt_work.c
@@ -46,7 +46,7 @@ lapack_int LAPACKE_cgeqrt_work( int matrix_layout, lapack_int m, lapack_int n,
         }
     } else if( matrix_layout == LAPACK_ROW_MAJOR ) {
         lapack_int lda_t = MAX(1,m);
-        lapack_int ldt_t = MAX(1,ldt);
+        lapack_int ldt_t = MAX(1,nb);
         lapack_complex_float* a_t = NULL;
         lapack_complex_float* t_t = NULL;
         /* Check leading dimension(s) */
@@ -76,6 +76,7 @@ lapack_int LAPACKE_cgeqrt_work( int matrix_layout, lapack_int m, lapack_int n,
         }
         /* Transpose input matrices */
         LAPACKE_cge_trans( matrix_layout, m, n, a, lda, a_t, lda_t );
+        LAPACKE_cge_trans( matrix_layout, nb, MIN(m,n), t, ldt, t_t, ldt_t );
         /* Call LAPACK function and adjust info */
         LAPACK_cgeqrt( &m, &n, &nb, a_t, &lda_t, t_t, &ldt_t, work, &info );
         if( info < 0 ) {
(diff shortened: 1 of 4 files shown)


LAPACKE/src/lapacke_cgeqrt_work.c
```diff
@@ -76,14 +76,15 @@ lapack_int LAPACKE_cgeqrt_work( int matrix_layout, lapack_int m, lapack_int n,
         }
         /* Transpose input matrices */
         LAPACKE_cge_trans( matrix_layout, m, n, a, lda, a_t, lda_t );
+        LAPACKE_cge_trans( matrix_layout, nb, MIN(m,n), t, ldt, t_t, ldt_t );
```
weslleyspereira, line 79: `t_t` is output only, right? If so, there is no need for this line. The same happens in the other 3 files.

weslleyspereira: @sbite0138, if you agree with me, I can make the change. Thanks

weslleyspereira (approved):

PR #769, opened 2022-11-29
Fix uninitialized M when quick return in DLARRD and SLARRD

akobotov:
**Description**
Quick return in DLARRD and SLARRD for `N .LE. 0` happens too early, the output parameter `M` is left unset. This may lead to crash in ScaLAPACK's DLARRE2 and SLARRE2 respectfully, where the value returned in the `M` parameter is used to run a loop.

The fix:
1. Moves the zero initialization of `M` before deciding to quick return.
2. Removes duplicate initialization of `INFO` and duplicate check for quick return

**Note**

Most likely there are similar issues in other ?LARR? functions introduced on 16db9c24ea.

### DIFF
--- a/SRC/dlarrd.f
+++ b/SRC/dlarrd.f
@@ -381,6 +381,7 @@
 *     .. Executable Statements ..
 *
       INFO = 0
+      M = 0
 *
 *     Quick return if possible
 *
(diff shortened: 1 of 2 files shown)


langou (approved):
Excellent contribution. Thanks. Yes, M should be set to 0 if N is 0. Great. Good fix. Thanks.

langou:
I opened #770 so that we remember
> Most likely there are similar issues in other ?LARR? functions introduced on 16db9c24ea

Any help is welcome.

PR #787, opened 2023-01-22
Fix Y misscale in GEMV for some cases

bE554357:
**Description**

According to the description, $M$ and $N$ could be zero. It is expected that the vector $\mathbf{y}$ will be correctly scaled by scalar for such cases. 

For example, let GEMV computes
$\mathbf{y} := 0 \cdot \mathbf{y} + \mathbf{A}\mathbf{x}$
and $\mathbf{x}$ is zero length vector.  It is expected, that $\mathbf{y} = 0$ after GEMV, but it is not scaled due to statement in implementation:
```
- 	 IF ((M.EQ.0) .OR. (N.EQ.0) .OR.
-     +    ((ALPHA.EQ.ZERO).AND. (BETA.EQ.ONE))) RETURN
```

This pull request is objected to fix the behavior in such cases.

### DIFF
--- a/BLAS/SRC/cgemv.f
+++ b/BLAS/SRC/cgemv.f
@@ -215,11 +215,6 @@
           RETURN
       END IF
 *
-*     Quick return if possible.
-*
-      IF ((M.EQ.0) .OR. (N.EQ.0) .OR.
-     +    ((ALPHA.EQ.ZERO).AND. (BETA.EQ.ONE))) RETURN
-*
       NOCONJ = LSAME(TRANS,'T')
 *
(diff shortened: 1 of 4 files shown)


langou:
Hi, 

Note that the code follows the convention as written in the article

> J. J. Dongarra, J. Du Croz, S. Hammarling, and R. J. Hanson. An extended set of FORTRAN Basic Linear Algebra Subprograms. ACM Trans. Math. Software, 14:1–32, 399, 1988.

I quote (page 5, section 3 - naming convention)

> Note that it is permissible to call the routines with M or N = 0, in which case the routines exit immediately without referencing their vector or matrix arguments.

Cheers, 
Julien.

Issue #791, opened 2023-02-07
Issues, related to generalized symmetric-definite banded eigenproblem (S|H)BGV/GVD


dklyuchinskiy:
Hello! 

Please take a look into presented checklist.

**Checklist**

- [x] Redundant space is added after the function name in XERBLA call. Related to (S|H)BGV source files only. I'd be willing to make a PR to solve this issue
```
     IF( INFO.NE.0 ) THEN
         CALL XERBLA( 'DGBSV ', -INFO )
         RETURN
      END IF
```
- [ ] Wrong documentation (or memory requirements) for miminal workspace required
CHBGVD:
```
      IF( N.LE.1 ) THEN
         LWMIN = 1+N
         LRWMIN = 1+N
         LIWMIN = 1
```
but documentations says 
```
          LRWORK is INTEGER
          The dimension of array RWORK.
          If N <= 1,               LRWORK >= 1.
```
Also in SSBGVD for `N` case
```
      ELSE
         liwmin = 1
         lwmin = 2*n
      END IF
```
but documentations says:
```
If JOBZ = 'N' and N > 1, LWORK >= 3*N.
```
- [ ] Problems from previous point are not caught due to missed error check in tests.
This branch
```
1940     ELSE IF( LSAMEN( 3, C3, 'DSG' ) ) THEN
```
 from `TESTING/EIG/dchkee.F` does not contain error check.
- [ ] Strange commented code in the same branch  `TESTING/EIG/dchkee.F`

![image](https://user-images.githubusercontent.com/30545679/217171747-c7d4d070-69cc-4dea-9174-d1703e9b7308.png)

Should it be deleted or is there a plan for future implementation?

Best regards,
Dmitry

PR #792, opened 2023-02-08
remove redundant space from xerbla call sbgv/hbgv

dklyuchinskiy:
Part of issue #791.

### DIFF
--- a/SRC/chbgv.f
+++ b/SRC/chbgv.f
@@ -235,7 +235,7 @@
          INFO = -12
       END IF
       IF( INFO.NE.0 ) THEN
-         CALL XERBLA( 'CHBGV ', -INFO )
+         CALL XERBLA( 'CHBGV', -INFO )
          RETURN
       END IF
 *
--- a/SRC/dsbgv.f
+++ b/SRC/dsbgv.f
@@ -228,7 +228,7 @@
          INFO = -12
       END IF
       IF( INFO.NE.0 ) THEN
-         CALL XERBLA( 'DSBGV ', -INFO )
+         CALL XERBLA( 'DSBGV', -INFO )
          RETURN
       END IF
 *
--- a/SRC/ssbgv.f
+++ b/SRC/ssbgv.f
@@ -228,7 +228,7 @@
          INFO = -12
       END IF
       IF( INFO.NE.0 ) THEN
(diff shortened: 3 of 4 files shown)


langou (approved):

langou:
Thanks. I do not have a strong opinion about these white spaces. Either way is fine. Thanks for the PR. That's great.

dklyuchinskiy:
> Thanks. I do not have a strong opinion about these white spaces. Either way is fine. Thanks for the PR. That's great.

I just wrote a small unit test to see the error message, which xerbla returns, and it is failed due to additional space in naming. Now behaviour is alinged with other LAPACK functions.

Thanks for the merge!

langou:
Ah! OK. Thanks for letting me know. Perfect then.

martin-frbg:
Actually many (most?) five-letter functions that use XERBLA have this trailing space, maybe it is the unit test that needs to be more flexible ?

dklyuchinskiy:
> Actually many (most?) five-letter functions that use XERBLA have this trailing space, maybe it is the unit test that needs to be more flexible ?

OMG, indeed. I did not check other 5 letter functions... Is it done for special purpose? LAPACK namings can be more than 6 letters.

Yeap, unit test can be changed, but it should be? Unfortunately, trailing space led to misundersanding from my side. It could be reverted if there exists LAPACK standard in such case. But for my point of few it is not aligned between functions.

langou:
> Is it done for special purpose? 

No, this is a remnant of FORTRAN 77. Six-letter is not needed any longer. The space in the parameter passed to XERBLA is not needed any longer.

The code with a space `CHBGV ` is OK. The code without a space is OK too.  I am fine either way. 

XERBLA should work with a space in the name of the subroutine. So it does bother me that the code with a space is a problem on your end. That being said, we can remove the space. I do not mind.

LAPACK namings can be more than 6 letters.

I agree with @martin-frbg that we want to be consistent across 5-letter-name subroutines.

dklyuchinskiy:
@langou thank you for comprehensive explanation!

Of course, it would be nice to have the same style for all 5-letter-name subroutines. Indeed, not all are aligned. For example, `GEQR`, `GELQ`, and now `SBGV` :sweat_smile: If it is not a big problem at the moment, it can be kept as is.

Issue #96, opened 2016-12-08
xLABAD


langou:
This is related to the  subroutines xLABAD.

> SLABAD takes as input the values computed by SLAMCH for underflow and
> overflow, and returns the square root of each of these values if the
> log of LARGE is sufficiently large.  This subroutine is intended to
> identify machines with a large exponent range, such as the Crays, and
> redefine the underflow and overflow limits to be the square roots of
> the values computed by SLAMCH.  This subroutine is needed because
> SLAMCH does not compensate for poor arithmetic in the upper half of
> the exponent range, as is found on a Cray.

Is that still relevant on current Cray machines? 

I would assume all machines are using IEEE and such a routine is obsolete by now.

jwdemmel:
Indeed, slabad should be a no-op on current computers, unless "double"
gets compiled into IEEE754 binary quad precision or decimal double precision.
So ok to omit.

weslleyspereira:
This issue was closed with #805.

PR #805, opened 2023-03-21
Remove legacy function LABAD

angsch:
**Description**

LABAD is a legacy function that used to make LAPACK usable on machines that did not comply with IEEE arithmetic. This PR removes the function calls to LABAD and turns LABAD itself into a no-op. See also #96.

If you agree that the removal (rather than turning the routine only into a no-op) is the right approach, I would go over the other routines (testing etc.), too.

**Checklist**


weslleyspereira:
Great PR! Thanks a lot for working on that.

I think we have at least 3 other people in favor of removing LABAD from LAPACK. Please, see https://github.com/Reference-LAPACK/lapack/discussions/727.

langou (dismissed):

martin-frbg:
How about adding a (non-fatal !) build-time test just in case there is something more modern around than a late-1990s Cray with the same limitations ?

langou:
Thanks @martin-frbg. 

I think I understand what you are saying but (1) I do not think LABAD translate to any modern architecture, and ever will, (2) it was a dirty fix back then, it is still a dirty fix and we should remove this, if we want to do this again, we should do something differently.


The way I am reading LABAD is as follows. 

(1) We can only use half of the range of numbers with confidence on old Cray's machine. So we want to restrict the range on an old Cray's by doing something like:
https://github.com/Reference-LAPACK/lapack/blob/cfaa5aefabb0dc3292c9c355d17c228f6351d7d8/SRC/dlabad.f#L94-L95

(2) How do we know we are on an old Cray? They have a huge range of number, LARGE is greater than +10^2000, so we can know we are on an old Cray by checking LARGE:
https://github.com/Reference-LAPACK/lapack/blob/cfaa5aefabb0dc3292c9c355d17c228f6351d7d8/SRC/dlabad.f#L93

The whole thing feels bogus to me. This is nice historical piece of software but I think we can safely move on from this routine.

Julien.

angsch:
LABAD is now fully removed from the numerical functions and tests; LABAD itself is no-op.

I intentionally did not move LABAD to the deprecated folder. This is to avoid an undefined symbol if someone calls LABAD in a custom external legacy routine. Recall that the deprecated routines are not built by default.

langou (approved):

PR #808, opened 2023-03-27
Refactor xGEBAL

eprovst:
**Description**

This PR should improve the readability of xGEBAL, without changing it functionally. In my testing, this does give identical results to the original, however a second pair of eyes is needed to verify. Execution times also seem to be identical, if not ever so slightly better.

The first commit contains the refactor. The second is a slight change where the NaN check is taken outside the loop. As far as I know, multiplying and dividing by 2 should not introduce NaNs, else the latter change is of course incorrect.

I also noticed that although the 2-norm is used, the stopping criterion seems to be for the 1-norm (cf. James, R., Langou, J., & Lowery, B. R. (2014). On matrix balancing and eigenvector computation. arXiv:1401.5766.)? I'll leave that one to the experts.

(As some background: balancing seemed relevant to an issue I was facing, so I wanted to be able to tinker with DGEBAL, in the end it was irrelevant to my work, but maybe this refactor could be useful to someone else later.)

**Checklist**

### DIFF
--- a/SRC/cgebal.f
+++ b/SRC/cgebal.f
@@ -85,6 +85,7 @@
 *> \verbatim
 *>          ILO is INTEGER
 *> \endverbatim
+*>
 *> \param[out] IHI
 *> \verbatim
 *>          IHI is INTEGER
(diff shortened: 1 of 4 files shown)


eprovst:
Now suddenly tests _are_ failing on my side, so changing this to draft...

langou (approved):

eprovst:
I seem to have missed a part of the code when translating my changes from the real to the complex case. Fixed as of now, sorry about missing that before opening the PR.

thijssteel (approved):
The original code is quite hard to read, so it's also hard to verify that this code does the same. That being said, I trust the tests for standard eigenvalue problems to be thorough enough. The scaling routine makes much more sense now. Good job @eprovst

PR #832, opened 2023-05-24
Removes unused SMIN from bdsqr

### LINKED ISSUE #243: Unused value?
https://github.com/Reference-LAPACK/lapack/blob/4f7f2aab27814a920613f27f4b0d5055a421dcaf/SRC/dbdsqr.f#L456-L468
At the end of the loop, `SMIN` is the absolute of the digonal minimum value, but its value is never used, I wonder why.

weslleyspereira:
Closes #243.

xBDSQR seems to be using SMINL to test if shifting would ruin relative accuracy. Therefore, SMIN is not necessary at:

https://github.com/Reference-LAPACK/lapack/blob/dfad0d5639d669736afc71d57e24d95001279577/SRC/dbdsqr.f#L448-L465

This was noted by @TarcioV at #243. Thanks!

### DIFF
--- a/SRC/cbdsqr.f
+++ b/SRC/cbdsqr.f
@@ -259,7 +259,7 @@
      $                   NM12, NM13, OLDLL, OLDM
       REAL               ABSE, ABSS, COSL, COSR, CS, EPS, F, G, H, MU,
      $                   OLDCS, OLDSN, R, SHIFT, SIGMN, SIGMX, SINL,
-     $                   SINR, SLL, SMAX, SMIN, SMINL, SMINOA,
+     $                   SINR, SLL, SMAX, SMIN, SMINOA,
      $                   SN, THRESH, TOL, TOLMUL, UNFL
 *     ..
 *     .. External Functions ..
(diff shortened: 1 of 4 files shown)


angsch:
> xBDSQR seems to be using SMINL to test if shifting would ruin relative accuracy. 

I agree, `smin` does not make any sense. My understanding of the algorithm is that `smax` and and `sminl` are over- and underestimators of the singular values. In view of the comment
https://github.com/Reference-LAPACK/lapack/blob/bd82aa0dfc851a6d470f3aada9731d9f6215fae9/SRC/dbdsqr.f#L385
I wonder if it makes sense to rename `sminl` to `smin` to have a matching name pair. I suspect that the name `sminl` was an emergency solution since `smin` was already taken.

In any case, it may be good to remove the variable that becomes unused from the list of "local scalars".

weslleyspereira:
Thanks @angsch! Your comments make complete sense to me. I went ahead and applied your suggestions regarding SMIN and SMINL.

langou (approved):

PR #834, opened 2023-05-24
Compilation of LAPACKE possibly independent from Fortran

### LINKED ISSUE #273: Add LAPACK option and support building only LAPACKE without requiring fortran
@langou  : @phcerdan and I will start working on the following

* [ ] Do not require `Fortran`. See https://github.com/Reference-LAPACK/lapack/blob/9fb29d8ea47d643ad44a67d3da5de36e280143f5/CMakeLists.txt#L3

* [ ] Add option `LAPACK` enabled by default (similar to the other option named `LAPACKE` and `CBLAS`). Enable Fortran only with `LAPACK` is on. In a nutshell, `enable_language(Fortran)` will be called only if `LAPACK` is `ON`

* [x] Exclude [this](https://github.com/Reference-LAPACK/lapack/blob/9fb29d8ea47d643ad44a67d3da5de36e280143f5/CMakeLists.txt#L27-L31) code if `CMAKE_CONFIGURATION_TYPES` is defined

* [ ] Update `appveyor.yml` to test building with `LAPACK:BOOL=OFF` and `LAPACKE:BOOL=ON` using generators `-G "Visual Studio 14 2015 Win64` (will add VS 2017 later)

weslleyspereira:
Attempt to close #273.

#273 has an interesting idea in my view. If one needs only LAPACKE (for instance, if LAPACK is already installed), Fortran compiler may not be necessary. This PR is my first step towards this idea.

Still needs testing, so I am putting this as a Draft.


weslleyspereira:
Rebased with current master.

weslleyspereira:
I put a test in the CI, so it should be now good to merge

langou (approved):

drhpc:
Oh, in #972 I added fixed Fortran language use in order to introduce optional C++ use. Might want to adapt that to fix both issues properly. My gripe was that suddenly CMake requires a working C++ compiler even without LAPACK++ or BLAS++. The pkgsrc build environment is picky and only provides the compilers a package claims to need.

PR #843, opened 2023-05-26
Updates the documentation of xGEMV and xGBMV related to when M=0 and N=0

### LINKED ISSUE #248: gbmv leaves result uninitialized for zero column matrix
SGBMV(TRANS='N',M=5,N=0,KL=0,KU=0,ALPHA=1,A,LDA=1,X,INCX=1,BETA=0,Y,INCY=1)

This should fill Y with zeros but actually it leaves it uninitialized. Encountered in BLAS-3.6.0.

### LINKED ISSUE #788: Y is misscaled in ?GEMV for some cases
**Description**

Description and possible solution is presented in #787

weslleyspereira:
Closes #248.
Closes #788.

The documentation of GEMV (and GBMV) misses a boundary case, namely when M=0 or N=0. There are two possibilities for the solution of this issue:

1. The behavior of GEMV (and GBMV) does not seem to match the behavior of GEMM, SYRK, SYR2K, HERK and HER2K regarding matrices with zero input sizes. In GEMM, for instance, C gets updated even if the internal dimension K is zero. In GEMV (and GBMV), if the internal dimension (M or N) is zero, Y does not get updated. The first solution is, therefore, to change the condition in GEMV (and GBMV) to be compatible to the lv3 BLAS.

2. The behavior of GEMV (and GBMV) is aligned to its original proposal (https://dl.acm.org/doi/10.1145/42288.42291) that says:
`
Note that it is permissible to call the routines with M or N = 0, in which case the routines exit immediately without referencing their vector or matrix arguments.
`
    Moreover, tests in ([testBLAS](https://github.com/tlapack/testblas)) confirm that several BLAS implementations conform with this rule.
    
This PR proposes a fix for the issue using (2), which preserves the current behavior and is compatible to several BLAS implementations.

### DIFF
--- a/BLAS/SRC/cgbmv.f
+++ b/BLAS/SRC/cgbmv.f
@@ -148,6 +148,7 @@
 *>           ( 1 + ( n - 1 )*abs( INCY ) ) otherwise.
 *>           Before entry, the incremented array Y must contain the
 *>           vector y. On exit, Y is overwritten by the updated vector y.
+*>           If either m or n is zero, then Y not referenced.
 *> \endverbatim
 *>
 *> \param[in] INCY
--- a/BLAS/SRC/cgbmv.f
+++ b/BLAS/SRC/cgbmv.f
@@ -148,7 +148,8 @@
 *>           ( 1 + ( n - 1 )*abs( INCY ) ) otherwise.
 *>           Before entry, the incremented array Y must contain the
 *>           vector y. On exit, Y is overwritten by the updated vector y.
-*>           If either m or n is zero, then Y not referenced.
+*>           If either m or n is zero, then Y not referenced and the function
+*>           performs a quick return.
 *> \endverbatim
 *>
 *> \param[in] INCY
--- a/BLAS/SRC/cgemv.f
+++ b/BLAS/SRC/cgemv.f
@@ -119,6 +119,7 @@
 *>           Before entry with BETA non-zero, the incremented array Y
 *>           must contain the vector y. On exit, Y is overwritten by the
 *>           updated vector y.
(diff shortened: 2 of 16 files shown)


angsch:
Is this behaviour checked/enforced by some test? It is a subtle change that could slip through when vendors upgrade their LAPACK version and their implementation does not follow reference-LAPACK.

weslleyspereira:
The most recent commit also updates `*la_gbamv.f` and `*la_geamv.f` using strategy (2).

We have:
1. routines with m-by-n A and y: `*gemv.f, *gbmv.f, *la_gbamv.f, *la_geamv.f`. (8 + 4 + 4 + 4 =) 20 files. __Documentation was updated.__
2. routines with n-by-n A and y: `*symv.f, *spmv.f, *sbmv.f, *hemv.f, *hpmv.f, *hbmv.f, *la_syamv.f, *la_heamv.f`. (4 + 4 + 6 + 2 + 2 + 2 + 4 + 2 =) 26 files. __No need to update documentation.__
3. routines with m-by-n A and no y: `*tbmv.f, *trmv.f, *tpmv.f`. (4 + 4 + 4 =) 12 files. __No need to update documentation.__

weslleyspereira:
> Is this behaviour checked/enforced by some test? It is a subtle change that could slip through when vendors upgrade their LAPACK version and their implementation does not follow reference-LAPACK

Thanks for asking! Yes, we currently check this behavior in ([testBLAS](https://github.com/tlapack/testblas)). In fact, multiple BLAS implementations (all libraries I tested so far in testBLAS) satisfy this behavior. We do not have such checks in the LAPACK test suite currently.

martin-frbg (approved):

PR #847, opened 2023-06-02
Improve variants of Cholesky and QR

angsch:
- Fix precision-related typos in the documentation
- Remove the file `sceil.f`, which provides the same functionality as the Fortran intrinsic `CEILING`.  `sceil` was only used in the LL variant of QR. All occurrences have been replaced with calls to `CEILING`, which is already used by routines in `SRC`.


langou:
Oh my. Thanks @angsch. 

I am thankful we are removing the SRC/VARIANTS/qr/LL/sceil.f file. Good idea.

langou (approved):

PR #849, opened 2023-06-05
fixes #848

### LINKED ISSUE #848: "minor" typo: docs globally (?) misuse the term "minor"
I am not aware of a text that defines a minor to be a matrix.  Hence I was bit surprised to read "the leading minor of order i is not positive definite", here:

https://github.com/Reference-LAPACK/lapack/blob/cea4a6315c5361a8b9074654ccdcc2def0bb1ea1/SRC/dpptrf.f#L79-L84

Actually, `grep -n SRC/*.f -e "minor"` reveals many more instances ...

AFAIK, a minor is a determinant and not a matrix (see, e.g., Aitken, 1944, p. 39; Horn and Johnson, 2013, p. 17; ...), and it is therefore nonsense to consider whether a minor is positive _definite_.  Should the text be globally amended to "the leading principal minor of order i is not positive"?

jaganmn:
AFAICT, in all cases the text was really describing Sylvester's criterion requiring positive leading principal minors, rather than some other criterion requiring positive definite leading principal submatrices.

### DIFF
--- a/SRC/VARIANTS/cholesky/RL/cpotrf.f
+++ b/SRC/VARIANTS/cholesky/RL/cpotrf.f
@@ -79,8 +79,8 @@ C> \verbatim
 C>          INFO is INTEGER
 C>          = 0:  successful exit
 C>          < 0:  if INFO = -i, the i-th argument had an illegal value
-C>          > 0:  if INFO = i, the leading minor of order i is not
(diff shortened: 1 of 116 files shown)


langou:
Thanks, this looks great.

langou (approved):

PR #852, opened 2023-06-11
Remove legacy warning for non IEEE 754 compliant machines

angsch:
Following the removal of `LABAD`, it looks reasonable to remove the comment warning about non IEEE 754 compliant machines. The machines mentioned in the comment date back to the 80s.

Besides deleting the comments in `LAED3`, also the call to `LAMC3` is removed. The purpose of `LAMC3` is to add 2 real numbers without compiler optimization. `LAMC3` appears to me like a outdated function that should be fully removed from LAPACK. Besides `LAED3`, there are a bunch of other routines such as `SLALS0` where `LAMC3` is called. None of these explain explicitly why `LAMC3` is required. I think that these occurrences serve the same reason as in `LAED3`, ie avoid accuracy problems on non-IEEE compliant machines. Do you agree?


langou (approved):

langou (dismissed):

langou:
So based on Jim's comments, it looks like all the changes of Angelika are great. 👍 

The changes to (S/D)LAED3 are good.

Then I think to be consistent, we might, in addition, want to do the "same" changes to (S/D)LAED9 and (S/D)LASD3.

This would leave the use of (S/D)LAMC3 in (S/D)LALS0 and (S/D)LASD8. Which is what, I think, we want. 

This agrees with Jim's email.

So here are a few more suggested changes.

**(S/D)LAED9:**
Remove these three lines of code:
https://github.com/Reference-LAPACK/lapack/blob/a83d8d28214f1a3760943d56e49c58a4e55ef13d/SRC/dlaed9.f#L233-L235
And then LAMDA (sic.) would be INPUT (as opposed to IN/OUT), but it is already declared as INPUT only (by mistake), so nothing to change then:
https://github.com/Reference-LAPACK/lapack/blob/a83d8d28214f1a3760943d56e49c58a4e55ef13d/SRC/dlaed9.f#L99

**(S/D)LASD3:**
Remove these three lines of code:
https://github.com/Reference-LAPACK/lapack/blob/a83d8d28214f1a3760943d56e49c58a4e55ef13d/SRC/dlasd3.f#L330-L332
And then LAMDA (sic.) would be INPUT (as opposed to IN/OUT), so need to edit this line of comment:
https://github.com/Reference-LAPACK/lapack/blob/a83d8d28214f1a3760943d56e49c58a4e55ef13d/SRC/dlasd3.f#L106

langou:
If possible, since we are editing (S/D)LAED3 and the like, maybe we can change `DLAMDA` to `DLAMBDA`? 

A `grep` of LAMDA reveals that we have LAMDA in the following ten files:

- (C/D/S/Z)LAED8
- (D/S)LAED2
- (D/S)LAED3
- (D/S)LAED9

langou:
@angsch write 

> Besides `LAED3`, there are a bunch of other routines such as `SLALS0` where `LAMC3` is called. None of these explain explicitly why `LAMC3` is required.

The fact that we are lacking comments is a good point. 

Maybe, in (S/D)LALS0 and (S/D)LASD8, just before calling LACM3, we add something like:
> We use LACM3 to perform (x+y)+z to try to enforce the correct parenthesis. The goal is to try to avoid the compiler doing x+(y+z).

And, in (S/D)LAMC3, we write
> This routine is currently used by (S/D)LALS0 and (S/D)LASD8. This routine is used to to try to enforce the correct parenthesis in the expression (x+y)+z. The goal is to try to avoid the compiler doing x+(y+z).

angsch:
Thank you, Julien, for following up and checking with Prof Demmel :+1:.   It really makes sense to fix `LAED9` and `LASD3`, too. I will add a commit with the changes you suggest tomorrow.

angsch:
Summary of changes:

- The initial commit missed some comments on non-IEEE 743 compliant machines.
- Rename lamba to lambda, change formatting to meet the line length limit when needed
- Remove calls to `LAMC3` when used as a workaround for Cray machines
- Update the documentation when the removal of `LAMC3` leaves the eigenvalues as `in` rather than `inout`
- Add comment on the purpose of `LAMC3` where it used to enforce the execution order

angsch:
We are getting a bunch of errors with CMake due to missing executables. At least some appear to be related to the CBLAS suffix change.

```
 80/123 Test  #28: example2_64_CBLAS ................***Not Run   0.00 sec
        Start  63: LAPACK-xeigtstd_dsg_in
Unable to find executable: /home/runner/work/lapack/lapack/build/bin/xexample2_64_CBLAS
```

langou:
Thanks. I reverted #856. Can you try again to run the test?

angsch:
Rebasing solved the CI problems. :+1:

Issue #548, opened 2021-05-08
Stackoverflow for `xlintstz_ztest_in` under 3.9.1 on windows 


h-vetinari:
While trying to build lapack 3.9.1 for conda-forge (see [here](https://github.com/conda-forge/lapack-feedstock/pull/51)), there's a new error for `ctest --output-on-failure -E "x*cblat*"` after building:
```
 81/108 Test  #81: LAPACK-xlintstz_ztest_in .........***Failed    0.06 sec
Running: D:/bld/blas-split_1620498704464/work/build/bin/xlintstz.exe
ARGS= OUTPUT_FILE;D:/bld/blas-split_1620498704464/work/build/TESTING/ztest.out;ERROR_FILE;D:/bld/blas-split_1620498704464/work/build/TESTING/ztest.out.err;INPUT_FILE;D:/bld/blas-split_1620498704464/work/TESTING/ztest.in
Test OUTPUT:

Test ERROR:

CMake Error at D:/bld/blas-split_1620498704464/work/TESTING/runtest.cmake:36 (message):
  Test D:/bld/blas-split_1620498704464/work/build/bin/xlintstz.exe returned
  Stack overflow
```

The build is done with minGW, roughly as follows (more details [here](https://github.com/conda-forge/lapack-feedstock/blob/master/recipe/bld.bat)):
```
set "CC=gcc.exe"
set "CXX=g++.exe"
set "FC=gfortran.exe"

cmake -G "MinGW Makefiles" ^
    -DCMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% ^
    -DBUILD_SHARED_LIBS=yes ^
    -DCMAKE_BUILD_TYPE=Release ^
    -DBUILD_TESTING=ON ^
    -DLAPACKE=ON ^
    -DCBLAS=ON ^
    -DBUILD_DEPRECATED=ON ^
    -Wno-dev ..

mingw32-make -j%CPU_COUNT%
mingw32-make install
if %ERRORLEVEL% NEQ 0 exit 1
```

martin-frbg:
Interesting as 3.9.1 is expected to have considerably _reduced_ the stack requirements for xlintstz compared to earlier versions.(And it definitely has on unixoid platforms)

martin-frbg:
sorry, the changes I was thinking of (#492) affected xeigtstz rather than xlintstz

weslleyspereira:
[...]

PR #854, opened 2023-06-17
Attempt to fix Appveyor on Windows

angsch:
The test log says that the failures in Appveyor are due to a stack overflow. In the past, this was fixed by moving to dynamic allocation, see https://github.com/Reference-LAPACK/lapack/issues/548. Let's give it a try and see if the errors vanish.


angsch:
Appveyor passed! I find it strange that the stack overflow problem did not surface when I added the syl01 tests, but at a later point in time.

martin-frbg (approved):

PR #860, opened 2023-06-26
[cz]hetrd_hb2st.F, [sd]sytrd_sb2st.F OpenMP fixes

chrwarm:
TID is an unmodified OpenMP private variable,
whose value is undefined in this code path. Since
it's within 'OMP MASTER', safe to assume its
value should be zero.

**Description**

**Checklist**


langou (approved):

weslleyspereira:
Thanks @chrwarm !

PR #885, opened 2023-07-14
Fix actual arguments in some procedure calls

mjacobse:
**Description**

Some internal and some testing calls did not match the actual procedure definitions:

* Rank mismatch in call to `xLARFG` from `xGELQT3`: `xLARFG` expects scalars for `ALPHA` and `TAU`,  but arrays were passed
* Rank mismatch in call to `xLANGE` from `xLATRS`: `xLANGE` expects array for `WORK`, but scalar was passed. Note that `WORK` is not referenced with `NORM /= 'I'`, so the returned value does not need to be used.
* Rank mismatch in call to `xLASCL` from `xTRSYL3`: `xLASCL` expects scalar for `INFO`, but array was passed
* Too many arguments in calls to `xSYTD2`/`xHETD2` from test `xERRST`
* Rank mismatch in call to `xLATRS3` from test `xERRTR`: `xLATRS3` expects array for `SCALE`, but scalar was passed
* Type mismatch in call to `CSCAL` from test `CCHKTR`: `CSCAL` expects `COMPLEX`, but `REAL` was passed. Since `ZCHKTR` calls `ZDSCAL` in that place, this should probably be a call to `CSSCAL` instead, for which the types fit.
* Type mismatch in call to `xLATRS3` from test `xCHKTR`: `xLATRS3` expects real array for `WORK`, but complex array was passed. Passing a real array with the previous number of elements `LWORK=NMAX` leads to invalid memory reference inside `xLATRS3`. Passing a real array of the same size `LWORK=2*NMAX` as before fixes that. Though I am not quite sure what the intention was here; the requirement for `LWORK` is not quite trivial and `NMAX` just seems to be an opaque test function parameter?

Found these by replacing external procedure declarations with full interface definitions using a [script](https://gist.github.com/mjacobse/98689c21cdd18a5835db9049798a9a86).

**Checklist**

~~- [ ] The documentation has been updated.~~
~~- [ ] If the PR solves a specific issue, it is set to be closed on merge.~~

### DIFF
--- a/TESTING/LIN/cerrtr.f
+++ b/TESTING/LIN/cerrtr.f
@@ -70,7 +70,7 @@
 *     .. Local Scalars ..
       CHARACTER*2        C2
       INTEGER            INFO
-      REAL               RCOND, SCALE
+      REAL               RCOND, SCALE, SCALES(0)
 *     ..
 *     .. Local Arrays ..
       REAL               R1( NMAX ), R2( NMAX ), RW( NMAX )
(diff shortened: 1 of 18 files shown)


langou (approved):
These changes looks good to me. Thanks a lot!
It would be good to have @angsch's opinion too.

angsch:
Thank you for raising these. Finding the mismatching argument count, and the real/complex mismatches are great!

My understanding is that arguments are passed as pointers, no matter if it is a scalar or an array. So in my view a mismatch between a caller passing a scalar and the receiver expecting an array is not a bug. @mjacobse Do you know about about this? Is this an issue with some compilers/future standards/... ?

kugelrund:
In practice I think you are right that compilers implement passing scalars and arrays the same way with a pointer. That is why this happened to work in practice. However I do not think that this is defined behaviour by the Fortran standard, but really non-conforming code. Though admittedly I have not gone through any Fortran standards document to get a clear reference to this, I just took **gfortran** rejecting the code (once an explicit interface is visible) at face value.

More so, a simple test case
```fortran
program main
    real :: array(1) = 0, scalar = 0

    call test_array(array)
    call test_scalar(scalar)

    call test_array(scalar)
    call test_scalar(array)
contains
    subroutine test_array(array)
        real :: array(1)
    end
    subroutine test_scalar(scalar)
        real :: scalar
    end
end program main
```
fails to compile with any of the compilers available on compiler-explorer: https://godbolt.org/z/nn7WM1qT3

* **gfortran**:
>/app/example.f90:7:20: [...] Error: Rank mismatch in argument 'array' at (1) (rank-1 and scalar)
>/app/example.f90:8:21: [...] Error: Rank mismatch in argument 'scalar' at (1) (scalar and rank-1)
* **ifort**/**ifx**:
>/app/example.f90(7): error #8284: If the actual argument is scalar, the dummy argument shall be scalar unless the actual argument is of type character or is an element of an array that is not assumed shape, pointer, or polymorphic.
>/app/example.f90(8): error #6634: The shape matching rules of actual arguments and dummy arguments have been violated.
* **flang**:
>/app/example.f90:7:21: error: Whole scalar actual argument may not be associated with a dummy argument 'array=' array

Testing locally, using `.f` file extension and using fixed-form does not change anything for me. And for this test case, no compiler arguments seem to relax this error either.

In a slightly different test case without `program` block (see https://godbolt.org/z/8zPfb1rGa), I am able to get a rank mismatch to compile under some circumstances. For **gfortran** since version 10.1, `-fallow-argument-mismatch` or `-std=legacy` has to be passed, which degrades the error to a warning. But passing any of `-Wpedantic`, `-pedantic` or `-pedantic-errors` to request strict standard compliance promotes it back to an error again. Indeed, from the [description of `-fallow-argument-mismatch`](https://gcc.gnu.org/onlinedocs/gfortran/Fortran-Dialect-Options.html#index-fallow-argument-mismatch):
> Some code contains calls to external procedures with mismatches between the calls and the procedure definition, or with mismatches between different calls. Such code is non-conforming, and will usually be flagged with an error. This options degrades the error to a warning [...]

**ifort** and **ifx** accept the program by default. But enabling `-warn all` (or just `-warn interface` is enough), the same errors as above are thrown.

So all of this makes me think that this is indeed invalid code according to the Fortran standard (and as far as I can tell for any version), even if this may happen to (accidentally) work in practice due to everything being passed as pointers.

mjacobse:
In practice I think you are right that compilers implement passing scalars and arrays the same way with a pointer. That is why this happened to work in practice. However I do not think that this is defined behaviour by the Fortran standard, but really non-conforming code. Though admittedly I have not gone through any Fortran standards document to get a clear reference to this, I just took **gfortran** rejecting the code (once an explicit interface is visible) at face value.

More so, a simple test case
```fortran
program main
    real :: array(1) = 0, scalar = 0

    call test_array(array)
    call test_scalar(scalar)

    call test_array(scalar)
    call test_scalar(array)
contains
    subroutine test_array(array)
        real :: array(1)
    end
    subroutine test_scalar(scalar)
        real :: scalar
    end
end program main
```
fails to compile with any of the compilers available on compiler-explorer: https://godbolt.org/z/nn7WM1qT3

* **gfortran**:
>/app/example.f90:7:20: [...] Error: Rank mismatch in argument 'array' at (1) (rank-1 and scalar)
>/app/example.f90:8:21: [...] Error: Rank mismatch in argument 'scalar' at (1) (scalar and rank-1)
* **ifort**/**ifx**:
>/app/example.f90(7): error #8284: If the actual argument is scalar, the dummy argument shall be scalar unless the actual argument is of type character or is an element of an array that is not assumed shape, pointer, or polymorphic.
>/app/example.f90(8): error #6634: The shape matching rules of actual arguments and dummy arguments have been violated.
* **flang**:
>/app/example.f90:7:21: error: Whole scalar actual argument may not be associated with a dummy argument 'array=' array

Testing locally, using `.f` file extension and using fixed-form does not change anything for me. And for this test case, no compiler arguments seem to relax this error either.

In a slightly different test case without `program` block (see https://godbolt.org/z/8zPfb1rGa), I am able to get a rank mismatch to compile under some circumstances. For **gfortran** since version 10.1, `-fallow-argument-mismatch` or `-std=legacy` has to be passed, which degrades the error to a warning. But passing any of `-Wpedantic`, `-pedantic` or `-pedantic-errors` to request strict standard compliance promotes it back to an error again. Indeed, from the [description of `-fallow-argument-mismatch`](https://gcc.gnu.org/onlinedocs/gfortran/Fortran-Dialect-Options.html#index-fallow-argument-mismatch):
> Some code contains calls to external procedures with mismatches between the calls and the procedure definition, or with mismatches between different calls. Such code is non-conforming, and will usually be flagged with an error. This options degrades the error to a warning [...]

**ifort** and **ifx** accept the program by default. But enabling `-warn all` (or just `-warn interface` is enough), the same errors as above are thrown.

So all of this makes me think that this is indeed invalid code according to the Fortran standard (and as far as I can tell for any version), even if this may happen to (accidentally) work in practice due to everything being passed as pointers.

angsch:
I did not know about this - perhaps the flags should be added to the (CI) Makefile/CMake?

mjacobse:
Hm, I'm not quite sure I understand, which flag(s) do you mean in particular? I mostly just referenced them to give reasons for why I think that the current implementation is not standard conforming Fortran. It may happen to work (accidentally), but I believe it is not at all guaranteed to work correctly on all possible platforms. Hence the proposed changes in this PR.

That said I would be very much in favor of introducing flags like `-Wpedantic` for gcc and `-warn all` for Intel, but to me that would be outside the scope of this PR.

TESTING/LIN/cerrtr.f
```diff
@@ -70,7 +70,7 @@ SUBROUTINE CERRTR( PATH, NUNIT )
 *     .. Local Scalars ..
       CHARACTER*2        C2
       INTEGER            INFO
-      REAL               RCOND, SCALE
+      REAL               RCOND, SCALE, SCALES(0)
```
martin-frbg, line 73: Pardon my ignorance - is the zero-size dimension (here and in the other ?errtr files) intentional ? This is documented (and used) as an "output" parameter of the respective ?clatrs3 function in which call it is occurs, and AFAIK assigning to a zero-sized array is undefined behavior at best...

martin-frbg: Sorry, I misread - NRHS is guaranteed to be at most zero in all cases here so ?latrs3 will do an early exit without writing to scale(s).

mjacobse: Yeah, that was my thinking. All the calls are with `NRHS <= 0` and the documentation says `SCALE` is of dimension `NRHS` (without an extra restriction like `MAX(1, NRHS)` that other functions have), so `0` seemed like the most natural dimension for the argument to me. Happy for it to be changed though if you think other dimensions would be more intuitive. I suppose a dimension of `1` would have been closer to before the change.

PR #889, opened 2023-08-09
Improve Cholesky variants

angsch:
Make the Cholesky variants a bit more competitive by replacing the scalar `potf2` with the recursive `potrf2`. The latter is already used in the primary Cholesky factorization (`SRC/dpotrf.f` etc.) .


langou (approved):

langou:
Thanks @angsch. There is also the question of the variant of the "main" POTRF in LAPACK. Right now I think LAPACK/SRC is using the left-looking variant. This is fine with me, but maybe using the plain recursive variant or the right-looking one would be better. (Preference for recursive.) I am comfortable if we leave things as they are.

PR #893, opened 2023-08-12
allow absolute GNU install dirs

chayleaf:
**Description**

`CMAKE_INSTALL_LIBDIR` is allowed to be an absolute path as per [CMake docs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html). This patch changes the build script to use `CMAKE_INSTALL_FULL_LIBDIR` instead, and gets rid of the prefix detection code (many other libraries, such as LLVM, have gotten rid of it already as it's quite finicky in practice due to absolute dirs being allowed).

Absolute `CMAKE_INSTALL_LIBDIR`s are used at least on NixOS. This is because NixOS packages are allowed to have multiple outputs (e.g. put docs in one dir, headers and other files for development to another, the binaries to a third one), and NixOS isn't FHS-compliant (There's no `/usr`, each package output has its own directory instead).

GNU Guix is a similar distro, but I'm not sure how CMake packages are built there. Either way, it's good to fully conform to the spec.

**Checklist**


weslleyspereira:
Hello @chayleaf. Would you be able to merge to/rebase with the current master branch? The script for the Github Actions needed to be updated in a recent commit. Thanks!

weslleyspereira (approved):

PR #900, opened 2023-08-27
Remove auxiliary GEQRS, GELQS

### LINKED ISSUE #709: GELQS, GEQRS outdated
Please consider deprecating `GELQS` and `GEQRS`. 

- There is no easy way to query the workspace to actually use the routines
- The documentation has errors (especially `GELQS`)
- The triangular system could do with a check of the diagonal for zeros, i.e., be solved with `TRTRS`, and not `TRSM`

`GELS` is a better and safer alternative. `GELS` wraps the matrix factorization (which `GE[LQ,QR]` require as input), the right-hand side transform and the solution of the triangular system in a coherent form. It solves a strict superset of what `GE[LQ,QR]S` can solve.

If you prefer to fix the routines rather than deprecating them, I am happy to create a PR.

angsch:
`GEQRS` and `GELQS` were likely computational routines once, (they have an input argument check, which auxiliary/test routines typically do not have) but have been superseded by `GELS` a long time ago. The only remaining occurrences are in the tests `CHKQR` and `CHKLQ`, where they are auxiliary routines to widen the test coverage.

* Replace `GEQRS` and `GELQS` by calls to GELS in tests.
* Remove the functions from the error exit tests.
* Delete `GEQRS` and `GELQS`.

This closes #709

An alternative to deleting the routines is to move them to `SRC/DEPRECATED/`. Any opinions?


martin-frbg:
My preference would be to move them to DEPRECATED -  they must have been in LAPACK for decades and one never knows what old but perfectly working code out there would be broken by their removal.

langou:
Thanks Angelika.

WRT: remove or move to DEPRECATED. I am fine either way. Either (*) we remove or (*) we move to DEPRECATED. I have a slight preference for 'remove'. I agree with Martin that it is likely that some codes out there use GEQRS or GELQS. However 'removing' or 'moving to DEPRECATED' will imply an error at build time for these codes either way. The idea of DEPRECATED was that it would be easier for folks to find the routines that we removed from LAPACK. But I think that, with the usage of GitHub, we do not really need DEPRECATED any longer. It should be feasible for folks to find these routines in the git repository of GitHub by searching the history of the code.

langou (dismissed):

martin-frbg:
> DEPRECATED was that it would be easier for folks to find the routines that we removed from LAPACK. But I think that, with the usage of GitHub, we do not really need DEPRECATED any longer. It should be feasible for folks to find these routines in the git repository of GitHub by searching the history of the code.

That viewpoint is certainly fine when assuming that the end user builds everything from source and will be only slightly inconvenienced by a build-breaking change. But I assume it will be a bigger problem for binary distributions (or their users), which most likely define BUILD_DEPRECATED already for maximum compatibility and would need to patch the deleted routines back in .

langou:
That's a good point. I missed this point. Very well then. This makes me change my opinion. 

Based on Martin's rationale, I changed my opinion and I think it is better to move the routine to DEPRECATED as opposed to remove them.

langou (approved):

PR #903, opened 2023-08-30
Fix BLAS++ and LAPACK++ build

weslleyspereira:
In the current LAPACK master, building with `BLAS++` flag `ON` causes the following error in CMake:

```
[build] [ 47%] Building CXX object test/CMakeFiles/tester.dir/test.cc.o
[build] In file included from /home/weslleyp/storage/lapack/build_RelWithDebInfo/blaspp-prefix/src/blaspp/test/test.cc:12:
[build] /home/weslleyp/storage/lapack/build_RelWithDebInfo/blaspp-prefix/src/blaspp/test/test.hh:9:10: fatal error: testsweeper.hh: No such file or directory
[build]     9 | #include "testsweeper.hh"
[build]       |          ^~~~~~~~~~~~~~~~
[build] compilation terminated.
[build] make[2]: *** [test/CMakeFiles/tester.dir/build.make:76: test/CMakeFiles/tester.dir/test.cc.o] Error 1
[build] make[1]: *** [CMakeFiles/Makefile2:154: test/CMakeFiles/tester.dir/all] Error 2
[build] make: *** [Makefile:136: all] Error 2
[build] ninja: build stopped: subcommand failed.
[proc] The command: /usr/local/bin/cmake --build /home/weslleyp/storage/lapack/build_RelWithDebInfo --config Release --target all -- exited with code: 1
```

There is some issue in BLAS++ (https://bitbucket.org/icl/lapackpp/downloads/lapackpp-2020.10.02.tar.gz) so that it can't find `testsweeper.hh`. I can give more details on my system configuration.

The point is: we actually don't need to build the BLAS++ tests to obtain `libblaspp`. Thus, this PR solves the issue by adding `-Dbuild_tests=OFF` to the configuration step of BLAS++ and LAPACK++ in LAPACK.


weslleyspereira:
I have two other concerns related to my last message:
1. We are using an old version of BLAS++ and LAPACK++ (2010.10.02). Is that intentional?
2. We have no tests in `cmake.yml` with `BLAS++=ON` and `LAPACK++=ON`.

mgates3:
>     1. We are using an old version of BLAS++ and LAPACK++ (2010.10.02). Is that intentional?

2020.10.02. You had me worried for a minute.

I think that was just what was available at the time. Perhaps we could use a `release` branch or tag to refer to the latest release. Is there a standard convention for that in Git? I just added the `release` branch, but we can do whatever seems best.

We can update the release version, I'm just not excited about updating LAPACK every time that BLAS++ and LAPACK++ are updated.

weslleyspereira:
Sorry for the "2010", @mgates3. Not intentional. I fixed my message above.

So, I have an idea on how we could make it easy to 

> I think that was just what was available at the time. Perhaps we could use a `release` branch or tag to refer to the latest release. Is there a standard convention for that in Git? I just added the `release` branch, but we can do whatever seems best.
> 
> We can update the release version, I'm just not excited about updating LAPACK every time that BLAS++ and LAPACK++ are updated.

Ok. Thanks. We could at least try to update these dependencies whenever we release LAPACK, or whenever BLAS++ and LAPACK++ have a major update regarding the wrappers. I am also not too excited to update LAPACK every time that BLAS++ and LAPACK++ are updated. I was just surprised it was using a release from almost 3 years ago.

weslleyspereira:
Now, about my second bullet: `We have no tests in cmake.yml with BLAS++=ON and LAPACK++=ON`. This is more worrisome to me. Yesterday, I was trying to use those flags and end up finding out that sometimes BLAS++ links with a BLAS library outside the repository, and I think that was not the plan when the flags were added. I will create an issue with more details.

weslleyspereira:
I updated this PR so that it solves #905. It should be ready for review.

Note that the changes do not add tests for the flag `LAPACK++` because of https://github.com/icl-utk-edu/lapackpp/issues/44.

CMakeLists.txt
```diff
@@ -472,38 +486,137 @@ function(_display_cpp_implementation_msg name)
...
+        -D BLAS_LIBRARIES="${BLAS_LIBRARIES}"
+        -D LAPACK_LIBRARIES="$<TARGET_FILE:${LAPACKLIB}>${BLAS_Fortran_LIB}" )
+    else()
+      add_custom_command( OUTPUT blaspp-cmd APPEND
+        COMMAND ${CMAKE_COMMAND}
+        -B "${blaspp_BINARY_DIR}"
+        -D BLAS_LIBRARIES="${BLAS_LIBRARIES}"
+        -D LAPACK_LIBRARIES="${LAPACK_LIBRARIES}" )
+    endif()
+  endif()
```
mgates3, line 549: These 4 `add_custom_command` blocks are quite redundant. May I suggest setting some variables to have one 1 instance? See my top-level review comment for suggestion.

CMakeLists.txt
```diff
@@ -613,45 +726,35 @@ install(FILES
...
   install(
-    FILES "${CMAKE_CURRENT_BINARY_DIR}/lib/lapackpp/lapackppConfig.cmake"
-          "${CMAKE_CURRENT_BINARY_DIR}/lib/lapackpp/lapackppConfigVersion.cmake"
+    FILES "${CMAKE_CURRENT_BINARY_DIR}/lib/cmake/lapackpp/lapackppConfig.cmake"
+          "${CMAKE_CURRENT_BINARY_DIR}/lib/cmake/lapackpp/lapackppConfigVersion.cmake"
     DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/"
   )
 
 endif()
 if (BLAS++)
```
mgates3, line 745: LAPACK++ requires BLAS++, so as above, I think this should be `if (BLAS++ OR LAPACK++)`.

mgates3 (changes requested):
I coded up [some changes in this commit](https://github.com/mgates3/lapack/commit/ecf6f41a57e2021244ed2b43fc40c6da160004e5). These work on my Mac with this configuration:
```
cmake -D BLAS++=true -D LAPACK++=true \
      -D USE_OPTIMIZED_BLAS=true -D BLA_VENDOR=Apple \
      -D CMAKE_INSTALL_PREFIX=${HOME}/install-lapack -B build
cd build
make
make install
```
Haven't tried other configurations or elsewhere yet.

CMakeLists.txt
```diff
@@ -472,38 +486,137 @@ function(_display_cpp_implementation_msg name)
...
+  endif()
+
+  # Setup remaining configuration options and installation
+  add_custom_command( OUTPUT blaspp-cmd APPEND
+    COMMAND ${CMAKE_COMMAND}
+    -B "${blaspp_BINARY_DIR}"
+    -D CMAKE_INSTALL_PREFIX="${blaspp_BINARY_DIR}"
+    -D CMAKE_INSTALL_LIBDIR="${PROJECT_BINARY_DIR}/lib"
+    -D blas_libraries_cached=""
+    -D lapack_libraries_cached=""
```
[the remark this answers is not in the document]
weslleyspereira: I don't think I need to set `blas_libraries_cached` and `lapack_libraries_cached` even if we configure `cmake` 2x. I needed it at some point of writing the PR, but I can't reproduce it anymore.

CMakeLists.txt
weslleyspereira: The redundancy is because we cannot use CMake generator expressions (e.g., `$<TARGET_FILE:${LAPACKLIB}>`) to set CMake variables at configuration time. See https://stackoverflow.com/questions/51353110/how-do-i-output-the-result-of-a-generator-expression-in-cmake. There is certainly a better way to do this, but I couldn't figure it out.

weslleyspereira:
Even with the changes from [this commit](https://github.com/mgates3/lapack/commit/ecf6f41a57e2021244ed2b43fc40c6da160004e5). I am still having trouble to use the flag LAPACK++. See https://github.com/icl-utk-edu/lapackpp/issues/50

weslleyspereira:
@mgates3, could you review the current stage of this PR. I think this is ready to merge. Thanks!

CMakeLists.txt
mgates3: With `cmake -D LAPACK++=on ..`, this will build BLAS++ and LAPACK++, but doesn't install `libblaspp.a`. Changing to `if (BLAS++ or LAPACK++)` fixes the issue. Can be merged with `if` block below (line 650).

CMakeLists.txt
```diff
@@ -551,47 +630,26 @@ install(FILES
...
-  write_basic_package_version_file(
-    "lapackppConfigVersion.cmake"
-    VERSION 2020.10.02
-    COMPATIBILITY AnyNewerVersion
-  )
-  install(
-    FILES "${CMAKE_CURRENT_BINARY_DIR}/lib/lapackpp/lapackppConfig.cmake"
-          "${CMAKE_CURRENT_BINARY_DIR}/lib/lapackpp/lapackppConfigVersion.cmake"
-    DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/"
+    DIRECTORY "${LAPACK_BINARY_DIR}/lib/"
```
mgates3, line 636: Same issue as below; use `${CMAKE_INSTALL_LIBDIR}`.

mgates3 (changes requested):
Here's the patch that worked for me:
```
sh leconte lapack> git diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 17609c05d..89db6a866 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -633,21 +633,19 @@ install(FILES
 
 if (LAPACK++)
   install(
-    DIRECTORY "${LAPACK_BINARY_DIR}/lib/"
+    DIRECTORY "${LAPACK_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}/"
     DESTINATION ${CMAKE_INSTALL_LIBDIR}
     FILES_MATCHING REGEX "lapackpp"
   )
 endif()
 
-if (BLAS++)
+if (BLAS++ OR LAPACK++)
   install(
-    DIRECTORY "${LAPACK_BINARY_DIR}/lib/"
+    DIRECTORY "${LAPACK_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}/"
     DESTINATION ${CMAKE_INSTALL_LIBDIR}
     FILES_MATCHING REGEX "blaspp"
   )
-endif()
 
-if (BLAS++ OR LAPACK++)
   install(
     DIRECTORY "${LAPACK_BINARY_DIR}/include/"
     DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
```
Otherwise, looks good. I tested on macOS with GNU compilers and OpenBLAS, and on Linux (Redhat) with GNU compilers, Intel MKL, and CUDA.

One minor issue, on Linux it said
```
-- Build files have been written to: /home/mgates/repos/lapack/build2/_deps/lapackpp-build
gmake[3]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
```
and then run `make` in serial instead of parallel, so it was rather slow. Not a big deal.

mgates3 (dismissed):

mgates3:
@weslleyspereira Any plans to fix this branch up and merge?
2024.10.26 is the latest BLAS++ and LAPACK++ version.
https://github.com/icl-utk-edu/blaspp/releases/tag/v2024.10.26
https://github.com/icl-utk-edu/lapackpp/releases/tag/v2024.10.26

weslleyspereira:
> @weslleyspereira Any plans to fix this branch up and merge? 2024.10.26 is the latest BLAS++ and LAPACK++ version. https://github.com/icl-utk-edu/blaspp/releases/tag/v2024.10.26 https://github.com/icl-utk-edu/lapackpp/releases/tag/v2024.10.26

I can do that. Should I wait for https://github.com/icl-utk-edu/blaspp/pull/91 and https://github.com/icl-utk-edu/lapackpp/pull/71?

weslleyspereira:
Couldn't make BLAS++ and LAPACK++ link with the correct LAPACK library when using shared libs. See:

- https://github.com/Reference-LAPACK/lapack/actions/runs/14112347619/job/39534062300 (end of build phase)
- https://github.com/Reference-LAPACK/lapack/actions/runs/14112347619/job/39534062301 (end of build phase)

Interestingly, static libraries are correctly linked. See:

https://github.com/Reference-LAPACK/lapack/actions/runs/14112347619/job/39534062287 (end of build phase)

PR #904, opened 2023-08-31
Apply ROUNDUP_LWORK function in lapack

kleineLi:
**Apply ROUNDUP_LWORK function in lapack** 

This work is continuation of PR https://github.com/Reference-LAPACK/lapack/pull/605. Applied only for s|c precisions. Do we need these changes to be applied for d|z precisions or it is redundant?

Please review! 
/cc @weslleyspereira


weslleyspereira:
Great that you took the time to work on that. Thanks!

I believe we should not think to much and just apply the round ups for single and double precision. If it is easy to do, I would include it in this PR. It adds minimal overhead to the workspace calculation, and the benefits can be big.

weslleyspereira:
I'll ping @mgates3 here as well, which might have something to add to the discussion.

mgates3:
Fine to add it to double-precision files as well. Keeps single & double consistent and avoids issues if single is generated from double somehow.

`double` has 53 bits of precision, so lwork is highly unlikely to exceed that. If lwork = 4n^2, it would need n > 47,453,132 = (2^53 / 4)^0.5 to encounter an issue, right?

In MAGMA, the equivalent to `roundup_lwork` is applied in all files due to precision generation, but it does nothing for double or double-complex, just returns the original value.
Cf. https://github.com/icl-utk-edu/magma/blob/606169ca033e50fb1853ceec0bc60250617ea6a7/control/magma_zauxiliary.cpp#L44

AlexanderZotkevich:
Let me suggest to apply only for s|c precisions. As it was mentioned by @mgates3 , 64 PiB of memory before there was an issue for double.

kleineLi:
Thank you @mgates3 for the comprehensive comment.

I think that adding functionality for d|z precision unnecessary at the moment, since the probability of working with matrix of size 47mn is extremely small.

But for single precision I've met the rounding issue with matrix size 9000.

SRC/sgetsls.f
```diff
@@ -294,6 +294,7 @@ SUBROUTINE SGETSLS( TRANS, M, N, NRHS, A, LDA, B, LDB,
 *
        SMLNUM = SLAMCH( 'S' ) / SLAMCH( 'P' )
        BIGNUM = ONE / SMLNUM
+       CALL SLABAD( SMLNUM, BIGNUM )
```
kleineLi, line 297: remove this

weslleyspereira:
Hi @kleineLi. Would you be able to merge to/rebase with the current master? The script for the Github Actions needed to be updated in a recent commit. Thanks

kleineLi:
Hi @weslleyspereira. I have rebased my branch on the current master

SRC/clahef_aa.f
```diff
@@ -169,7 +169,8 @@ SUBROUTINE CLAHEF_AA( UPLO, J1, M, NB, A, LDA, IPIV,
 *     .. External Functions ..
       LOGICAL      LSAME
       INTEGER      ICAMAX, ILAENV
-      EXTERNAL     LSAME, ILAENV, ICAMAX
```
kleineLi, line 172: revert

SRC/ssytrd.f
```diff
@@ -222,8 +222,8 @@ SUBROUTINE SSYTRD( UPLO, N, A, LDA, D, E, TAU, WORK, LWORK, INFO )
 *     ..
 *     .. External Functions ..
       LOGICAL            LSAME
-      INTEGER            ILAENV
-      EXTERNAL           LSAME, ILAENV
+      INTEGER            ILAENV, SROUNDUP_LWORK
```
kleineLi, line 225:
```suggestion
      INTEGER            ILAENV
      REAL                  SROUNDUP_LWORK
```

SRC/sgghd3.f
```diff
@@ -259,15 +259,15 @@ SUBROUTINE SGGHD3( COMPQ, COMPZ, N, ILO, IHI, A, LDA, B, LDB, Q,
 *     ..
 *     .. External Functions ..
       LOGICAL            LSAME
-      INTEGER            ILAENV
-      EXTERNAL           ILAENV, LSAME
+      INTEGER            ILAENV, SROUNDUP_LWORK
```
kleineLi, line 262:
```suggestion
      INTEGER            ILAENV
      REAL               SROUNDUP_LWORK
```

angsch:
@kleineLi Could you be so kind and apply the fix also to the alternative implementation of the QR decomposition (`SRC/VARIANTS/qr/LL`)?

kleineLi:
> @kleineLi Could you be so kind and apply the fix also to the alternative implementation of the QR decomposition (`SRC/VARIANTS/qr/LL`)?

Done

weslleyspereira:
We can probably merge this one. I don't think the failure in AppVeyor has anything to do with this PR.

weslleyspereira (approved):

PR #909, opened 2023-09-14
Fix issue #908 related to accumulation in xLASSQ

### LINKED ISSUE #908: Problem with the accumulation of the existing sum in xLASSQ
While running tests in https://github.com/tlapack/tlapack/pull/276, I found an undesirable behavior in the accumulation of the existing sum in xLASSQ which may lead to wrong Frobenius norm computation by xLANGE and others. I explain better below. 

In https://github.com/Reference-LAPACK/lapack/pull/570/commits/e7d572c5234e2b4789cde63dbeb82019ba4b19bb, I introduced the following assumption to xLASSQ:
```f90
!> If scale * sqrt( sumsq ) > tbig then
!>    we require:   scale >= sqrt( TINY*EPS ) / sbig   on entry,
!> and if 0 < scale * sqrt( sumsq ) < tsml then
!>    we require:   scale <= sqrt( HUGE ) / ssml       on entry,
!> where
!>    tbig -- upper threshold for values whose square is representable;
!>    sbig -- scaling constant for big numbers; \see la_constants.f90
!>    tsml -- lower threshold for values whose square is representable;
!>    ssml -- scaling constant for small numbers; \see la_constants.f90
!> and
!>    TINY*EPS -- tiniest representable number;
!>    HUGE     -- biggest representable number.
```
Unfortunately, these restrictions are not well fitted for the norm routines like xLANGE.

In fact, the following code returns a wrong Frobenius norm with LAPACK 3.10.1 and LAPACK 3.11.0:
```c
#include <float.h>
#include <lapack.h>
#include <math.h>
#include <stdio.h>

int main() {
  lapack_int const n = 2;
  float const B =
      powf(FLT_RADIX, floor(0.5f * (FLT_MAX_EXP - FLT_MANT_DIG + 1)));
  printf("B = %f\n", B);

  float a[n * n];
  for (lapack_int i = 0; i < n; ++i)
    for (lapack_int j = 0; j < n; ++j)
      a[i * n + j] = B;

  printf("Frobenius norm: %f\n",
         LAPACK_slange("Frobenius", &n, &n, a, &n, NULL));
  printf("Exact norm: %f\n", n * B);

  return 0;
}
```
In this example, we have a 2-by-2 matrix with entries of moderate size (B = 2^52 = 4503599627370496).

I will propose a fix in a separate PR.

weslleyspereira:
This solves the issue by removing the assumptions on `scale` and `sumsq` on input.

Closes #908.

### DIFF
--- a/SRC/dlassq.f90
+++ b/SRC/dlassq.f90
@@ -122,7 +122,7 @@
 !  =====================================================================
 subroutine DLASSQ( n, x, incx, scale, sumsq )
    use LA_CONSTANTS, &
-      only: wp=>dp, zero=>dzero, one=>done, safmin=>dsafmin, &
+      only: wp=>dp, zero=>dzero, one=>done, &
             sbig=>dsbig, ssml=>dssml, tbig=>dtbig, tsml=>dtsml
    use LA_XISNAN
 !
@@ -140,11 +140,7 @@ subroutine DLASSQ( n, x, incx, scale, sumsq )
 !  .. Local Scalars ..
    integer :: i, ix
    logical :: notbig
-   real(wp) :: abig, amed, asml, ax, ymax, ymin, sqrtmin, sqrtmax
-!  ..
-!  .. Set constants ..
-   sqrtmin = sqrt(safmin)
-   sqrtmax = one / sqrtmin
+   real(wp) :: abig, amed, asml, ax, ymax, ymin
 !  ..
 !
 !  Quick return if possible
@@ -192,14 +188,8 @@ subroutine DLASSQ( n, x, incx, scale, sumsq )
       ax = scale*sqrt( sumsq )
       if (ax > tbig) then
          if (scale > one) then
-            scale = scale * sbig  ! sbig < scale <= sbig * max
-            if (scale > sqrtmin) then
-               ! sqrtmin < scale < sqrtmax, so it is safe to square scale
-               abig = abig + (scale * scale) * sumsq
-            else
-               ! Do not square scale, as it may underflow
-               abig = abig + scale * (scale * sumsq)
-            end if
+            scale = scale * sbig
+            abig = abig + scale * (scale * sumsq)
          else
             ! sumsq > tbig^2 => (sbig * (sbig * sumsq)) is representable
             abig = abig + scale * (scale * (sbig * (sbig * sumsq)))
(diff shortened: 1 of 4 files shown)


SRC/dlassq.f90
```diff
@@ -200,13 +191,43 @@ subroutine DLASSQ( n, x, incx, scale, sumsq )
...
-         abig = abig + (scale*sbig)**2 * sumsq
+         if (scale > one) then
+            scale = scale * sbig  ! sbig < scale <= sbig * max
+            if (scale > sqrtmin) then
+               ! sqrtmin < scale < sqrtmax, so it is safe to square scale
+               abig = abig + (scale * scale) * sumsq
+            else
+               ! Do not square scale, as it may underflow
+               abig = abig + scale * (scale * sumsq)
+            end if
```
angsch, line 202: Is is possible to remove the if branch altogether and always take the accumulation order of the else branch, ie always accumulate `scale * (scale * sumsq)` ?

weslleyspereira: I think so. Thanks a lot! I will try to write some notes with the upper and lower bounds on those quantities. I will provide the improvement in the next commit.

weslleyspereira: Let me write the reasoning behind the changes. It simpler than I first thought.

Here is the new code: 

```f90
   if( sumsq > zero ) then
      ax = scale*sqrt( sumsq )
      if (ax > tbig) then
         if (scale > one) then
            scale = scale * sbig
            abig = abig + scale * (scale * sumsq)
         else
            ! sumsq > tbig^2 => (sbig * (sbig * sumsq)) is representable
            abig = abig + scale * (scale * (sbig * (sbig * sumsq)))
         end if
      else if (ax < tsml) then
         if (notbig) then
            if (scale < one) then
               scale = scale * ssml
               asml = asml + scale * (scale * sumsq)
            else
               ! sumsq < tsml^2 => (ssml * (ssml * sumsq)) is representable
               asml = asml + scale * (scale * (ssml * (ssml * sumsq)))
            end if
         end if
      else
         amed = amed + scale * (scale * sumsq)
      end if
   end if
```

Cases:
1. `ax > tbig` and `scale > one`:
    - we can safely compute the update `scale_updated = scale * sbig` because `sbig < one` and `scale > one`.
    - `scale_updated * sumsq` does not underflow. In fact, if `scale_updated * sumsq` were to underflow, then we would know that `scale_updated < one` and `sumsq < one`. In this case, `scale_updated * scale_updated * sumsq` would also underflow independently on how we organize the operations. This would contradict the hypothesis that `(ax * sbig) * (ax * sbig) = scale_updated * scale_updated * sumsq` is representable. So, indeed, `scale_updated * sumsq` does not underflow.
    - `scale_updated * sumsq` does not overflow using an argument similar to the above.
2. `ax > tbig` and `scale <= one`:
    - In this case, we do not know if `scale * sbig` is representable. However, we know that `sumsq > tbig * tbig` and, thus, `(sbig * (sbig * sumsq))` is representable.
3. `ax < tsml` and `scale < one`: Reasoning is analogous to (1).
4. `ax < tsml` and `scale >= one`: Reasoning is analogous to (2).
5. `tsml <= ax <= tbig`: Reasoning is analogous to the second bullet in (1).

langou (approved):

PR #910, opened 2023-09-18
Fix build for CMake >= 3.27

scivision:
CMake >= 3.27 [warns](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings) if `cmake_minimum_required(VERSION )` is less than 3.5. 

CMake for Lapack was already using CMake 3.6 features.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.6)
 
 project(LAPACK Fortran C)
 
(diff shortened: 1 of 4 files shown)


weslleyspereira (dismissed):
I like the change. The CMake build system in LAPACK may actually be using functionality from more recent versions of CMake anyway.

ACSimon33:
@weslleyspereira We need at least CMake 3.6 since we are using `list(FILTER ..)`: 

[https://cmake.org/cmake/help/latest/command/list.html#filter](https://cmake.org/cmake/help/latest/command/list.html#filter)

So maybe it makes sense to set the minimum version to 3.6

weslleyspereira (approved):

PR #913, opened 2023-10-10
Try testBLAS with LAPACK

weslleyspereira:
This PR propose the inclusion of tests covering:
1. Inf and NaN propagation.
2. Corner cases like: zero and negative sizes on inputs; invalid arguments.

About the tests:
- They are written in C++.
- The test library is here: https://github.com/tlapack/testblas.
- Tests follows what was proposed in https://arxiv.org/abs/2207.09281.
- Tests should also work when `USE_OPTIMIZED_BLAS` or `USE_OPTIMIZED_LAPACK` is ON.
- The changes only target the CMake build system by adding a new target: `run-testBLAS`.

This PR is aligned to what is proposed in #903.


weslleyspereira:
Related to #788.

PR #920, opened 2023-10-31
Always use C as linker language for CBLAS

weslleyspereira:
@mkrainiuk, could please you take a look at this change? With this, we can install CBLAS without a Fortran compiler if we already have an optimized BLAS library installed.


mkrainiuk (approved):
Good idea, thank you for improving it!

Issue #915, opened 2023-10-17
Intel ifx and cmake build failing on linux after PR#898


colleeneb:
**Description**

It looks like the merge of PR https://github.com/Reference-LAPACK/lapack/pull/898 broke my Intel ifx build. The PR itself looks good (I guess it fixed the Intel branch to getting the right flags!) but the Intel flags for linux (https://github.com/Reference-LAPACK/lapack/blob/4174d8d080199d30c0c834a959752bc361b91cbe/CMAKE/CheckLAPACKCompilerFlags.cmake#L23) have a space between the `integer-size` and `64` and this is breaking the build. It now fails at cmake time on trying to compile with the error of

```
...
-- Detecting Fortran/C Interface
Failed to compile
...
 ifx  -.../lapack/build/CMakeFiles/FortranCInterface/VerifyC -recursive -fp-model strict -integer-size;64 -O2 -c /soft/packaging/spack/gnu-ldpath/build/linux-sles15-x86_64/gcc-11.2.0/cmake-3.26.3-vnn7ncxwjjekhm5ehnwhk4w6tyoygc4p/share/cmake-3.26/Modules/FortranCInterface/Verify/VerifyFortran.f -o CMakeFiles/VerifyFortran.dir/VerifyFortran.f.o
    ifx: command line warning #10006: ignoring unknown option '-integer-size'
    ifx: command line error: no files specified; for help type "ifx -help"
    /bin/sh: 64: command not found
```
The flag is passed as `-integer-size;64` so it's not getting passed correctly somehow. Instead of the desired `-integer-size 64` it's `-integer-size;64`. I can adjust CheckLAPACKCompilerFlags.cmake to replace
```
set(FOPT_ILP64 -integer-size 64)
```
with
``` 
set(FOPT_ILP64 "-integer-size 64")
```
as a workaround though. I can make a PR with this if it's agreed that this is the right solution.
```

To reproduce, run the cmake step with ifx with:
```
blas_root=$PWD
    git clone https://github.com/Reference-LAPACK/lapack.git
    cd lapack/
    mkdir build
    cd build/
    cmake -DLAPACKE=ON -DBUILD_INDEX64=ON -DBUILD_SHARED_LIBS=ON -DCBLAS=ON -DCMAKE_INSTALL_LIBDIR=${blas_root} -DCMAKE_INSTALL_PREFIX=${blas_root} ..
``` 


**Checklist**

- [ x] I've included a minimal example to reproduce the issue
[...]

PR #922, opened 2023-10-31
Adding in quotes in cmake for Intel ifx and linux build 

colleeneb:
This is following up on https://github.com/Reference-LAPACK/lapack/issues/915. 

This adds in quotes around a flag passed to the Intel compiler. Otherwise we see a cmake error, as shown in https://github.com/Reference-LAPACK/lapack/issues/915.

**Description**

**Checklist**


weslleyspereira (approved):
Thanks!

PR #924, opened 2023-11-01
Clarify documentation of singularity/rank-deficiency checks

### LINKED ISSUE #918: Documentation for *TRTRS should note it only detects exact singularity
**Description**
The current description of DTRTRS reads claims that `A check is made to verify that A is nonsingular.`. Initially I assumed that nonsingular meant "nonsingular within acceptable numerical precision", but after reading the code itself it has become evident that only exact singularity (as in `0.0d+0`) triggers an error.

I know it is too late to change this behavior, but was this a good design decision? I would think that at the very least, any subnormally small (< 2^−1022 for FP64) diagonal element is ought to trip the error exit. Perhaps something like `tiny(0.0d+0)` would have been more appropriate.
This does not only affect *TRTRS, but many of the callers of it, which rely on *TRTRS for their rank-deficiency/singularity checking. For example, DGELS only errors out if the least squares matrix has an exact rank deficiency, even though it is reasonable to assume that being only a subnormal number away from rank deficiency will result in a loss of numerical precision in the computed linear coefficients. 

Admittedly all of the above is purely academic. Even if I am correct, it would not be nice to the users if the behaviour of the function was changed, so I am proposing to change the documentation instead, making it immediately clear to the reader that these checks only detect _exact_ singularity/rank-deficiency, and that near-singular cases are not caught.

**Checklist**

- [N/A] I've included a minimal example to reproduce the issue

TiborGY:
### Description
This PR fixes and closes #918, by changing the relevant parts of the docs.

### *TRTRS, *TBTRS and *TPTRS
- [X] In *TRTRS, *TBTRS and *TPTRS, replace `A check is made to verify that A is nonsingular.` with the following:
```
This subroutine verifies that A is nonsingular, but callers should note that only exact
singularity is detected. It is conceivable for one or more diagonal elements of A to be
subnormally tiny numbers without this subroutine signalling an error.

If a possible loss of numerical precision due to near-singular matrices is a concern, the
caller should verify that A is nonsingular within some tolerance before calling this subroutine.
```
- [X] In *TRTRS, *TBTRS and *TPTRS, replace `> 0:  if INFO = i, the i-th diagonal element of A is zero,` with `> 0:  if INFO = i, the i-th diagonal element of A is exactly zero,`

### Callers of *TRTRS: *GELS, *GELST, *GETSLS, *GGLSE, *GGGLM 
- [X] In *GELS, *GELST, *GETSLS replace `It is assumed that A has full rank.` with the following:
```
It is assumed that A has full rank, and only a rudimentary protection
against rank-deficient matrices is provided. This subroutine only detects
exact rank-deficiency, where a diagonal element of the triangular factor
of A is exactly zero.

It is conceivable for one (or more) of the diagonal elements of the triangular
factor of A to be subnormally tiny numbers without this subroutine signalling
an error. The solutions computed for such almost-rank-deficient matrices may
be less accurate due to a loss of numerical precision.
```
- [X] In *GELS, *GELST, *GETSLS replace `> 0:  if INFO =  i, the i-th diagonal element of the triangular factor of A is zero` with `> 0:  if INFO =  i, the i-th diagonal element of the triangular factor of A is exactly zero`
- [x] In *GGLSE, add the following:
```
Callers of this subroutine should note that the singularity/rank-deficiency checks
implemented in this subroutine are rudimentary. The <S,D,C,Z>TRTRS subroutine called by this
subroutine only signals a failure due to singularity if the problem is exactly singular.

It is conceivable for one (or more) of the factors involved in the generalized RQ
factorization of the pair (B, A) to be subnormally close to singularity without this
subroutine signalling an error. The solutions computed for such almost-rank-deficient
problems may be less accurate due to a loss of numerical precision.
```
- [x] In *GGLSE, replace `RQ factorization of the pair (B, A) is singular` with `RQ factorization of the pair (B, A) is exactly singular`
- [x] In *GGGLM, add the following:
```
Callers of this subroutine should note that the singularity/rank-deficiency checks
implemented in this subroutine are rudimentary. The <S,D,C,Z>TRTRS subroutine called by this
subroutine only signals a failure due to singularity if the problem is exactly singular.

It is conceivable for one (or more) of the factors involved in the generalized QR
factorization of the pair (A, B) to be subnormally close to singularity without this
subroutine signalling an error. The solutions computed for such almost-rank-deficient
problems may be less accurate due to a loss of numerical precision.
```
- [x] In *GGGLM, replace `QR factorization of the pair (A, B) is singular` with `QR factorization of the pair (A, B) is exactly singular`

### Callers of *TBTRS: 
- [x] Apart from the testing subroutines, there appear to be no internal callers of *TBTRS in LAPACK.

### Callers of *TPTRS: 
- [x] Apart from the testing subroutines, there appear to be no internal callers of *TPTRS in LAPACK.

**Checklist**

### DIFF
--- a/SRC/cgels.f
+++ b/SRC/cgels.f
@@ -37,7 +37,17 @@
 *>
 *> CGELS solves overdetermined or underdetermined complex linear systems
 *> involving an M-by-N matrix A, or its conjugate-transpose, using a QR
-*> or LQ factorization of A.  It is assumed that A has full rank.
+*> or LQ factorization of A.
+*>
+*> It is assumed that A has full rank, and only a rudimentary protection
+*> against rank-deficient matrices is provided. This subroutine only detects
+*> exact rank-deficiency, where a diagonal element of the triangular factor
+*> of A is exactly zero.
+*>
+*> It is conceivable for one (or more) of the diagonal elements of the triangular
+*> factor of A to be subnormally tiny numbers without this subroutine signalling
+*> an error. The solutions computed for such almost-rank-deficient matrices may
+*> be less accurate due to a loss of numerical precision.
 *>
 *> The following options are provided:
 *>
@@ -161,7 +171,7 @@
 *>          = 0:  successful exit
 *>          < 0:  if INFO = -i, the i-th argument had an illegal value
 *>          > 0:  if INFO =  i, the i-th diagonal element of the
-*>                triangular factor of A is zero, so that A does not have
+*>                triangular factor of A is exactly zero, so that A does not have
 *>                full rank; the least squares solution could not be
 *>                computed.
 *> \endverbatim
--- a/SRC/cgelst.f
+++ b/SRC/cgelst.f
@@ -38,7 +38,16 @@
 *> CGELST solves overdetermined or underdetermined real linear systems
 *> involving an M-by-N matrix A, or its conjugate-transpose, using a QR
 *> or LQ factorization of A with compact WY representation of Q.
-*> It is assumed that A has full rank.
+*>
+*> It is assumed that A has full rank, and only a rudimentary protection
+*> against rank-deficient matrices is provided. This subroutine only detects
+*> exact rank-deficiency, where a diagonal element of the triangular factor
+*> of A is exactly zero.
+*>
+*> It is conceivable for one (or more) of the diagonal elements of the triangular
+*> factor of A to be subnormally tiny numbers without this subroutine signalling
+*> an error. The solutions computed for such almost-rank-deficient matrices may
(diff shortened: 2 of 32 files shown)


langou (approved):

TiborGY:
Please note that I am still working on this PR, it is not ready for merge yet. Hence the draft status.

langou (dismissed):

TiborGY:
This is now ready for review and merge.

TiborGY:
@langou If there are no objections, this is ready for merge when convenient.

langou (approved):

PR #925, opened 2023-11-02
Add Anderson's tests of NRM2

angsch:
The existing tests of `NRM2` cover only the branch where squaring does not require scaling. This commit adds Edward Anderson's tests in slightly modified form. His tests include arrays with extreme numbers such that the branches that require scaling are tested. The pass/fail threshold is the same as in Anderson's tests.

The original tests are included in the archive provided alongside with

> Anderson E. (2017)
> Algorithm 978: Safe Scaling in the Level 1 BLAS
> ACM Trans Math Softw 44:1--28
> https://doi.org/10.1145/3061665

This commit modifies the tests such that they could be integrated into the existing BLAS-1 tests. It hardcodes the constants rather than computing them at compile time because some of the required intrinsics are not included in `f2c`. The original idea was to make it easy for dependent projects to integrate the tests, either in this form or in C. However, I could not find a way to include `Inf` and `NaN` tests without breaking `f2c`, more below.

In the current form, test coverage improves as follows.

|| [D,S]NRM2    | [DZ,SC]NRM2 |
| -------- | -------- | ------- |
|lines| 59.1% -> 100.0%  | 67.0% -> 100.0%    |
|branches| 30.0% ->  90.0% | 41.7% ->  91.7%    |

**Limitations:**
- ~The case `incx = 0` is included in Anderson's original test, but is **not** included in this commit. Adding non-positive incx to the loop over `incx` introduces compiler errors (computed indices in arrays < 1) in the existing tests. I have not looked into this.~
- This commit breaks `f2c` because of the handling of `Inf` and `NaN`. It uses the workaround proposed by Anderson to avoid the compiler from complaining about overflow and division by zero when `Inf` and `NaN` are (intentionally) generated.  `HUGE` is a function that is not permitted in `f2c`. If the tests of `Inf` and `NaN` are removed, `f2c` works. 

Since `Inf` and `NaN` are covered by the TLAPACK tests, I would be happy to remove these two tests and keep the xBLAT1 tests `f2c`able. Further, is there any opinion about testing `incx = 0`? It could of course be added as a special case only for `NRM2`.

### DIFF
--- a/BLAS/TESTING/cblat1.f
+++ b/BLAS/TESTING/cblat1.f
@@ -121,7 +121,8 @@
       SUBROUTINE CHECK1(SFAC)
 *     .. Parameters ..
       INTEGER           NOUT
-      PARAMETER         (NOUT=6)
+      REAL              THRESH
+      PARAMETER         (NOUT=6, THRESH=10.0E0)
 *     .. Scalar Arguments ..
       REAL              SFAC
 *     .. Scalars in Common ..
@@ -141,7 +142,7 @@
       INTEGER           ICAMAX
       EXTERNAL          SCASUM, SCNRM2, ICAMAX
 *     .. External Subroutines ..
-      EXTERNAL          CSCAL, CSSCAL, CTEST, ITEST1, STEST1
+      EXTERNAL          CB1NRM2, CSCAL, CSSCAL, CTEST, ITEST1, STEST1
 *     .. Intrinsic Functions ..
       INTRINSIC         MAX
 *     .. Common blocks ..
@@ -256,6 +257,10 @@
    20       CONTINUE
             IF (ICASE.EQ.6) THEN
 *              .. SCNRM2 ..
+*              Test scaling when some entries are tiny or huge
+               CALL CB1NRM2(N,(INCX-2)*2,THRESH)
+               CALL CB1NRM2(N,INCX,THRESH)
+*              Test with hardcoded mid range entries
                CALL STEST1(SCNRM2(N,CX,INCX),STRUE2(NP1),STRUE2(NP1),
      +                     SFAC)
             ELSE IF (ICASE.EQ.7) THEN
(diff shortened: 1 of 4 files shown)


langou:
Thanks @angsch. This is great and very impressive. Great to check a larger range of input. Impressive to improve test coverage from 41.7% to 91.7%.

I do not have a strong opinion on either (a) compatibility with F2C and (b) testing INCX = 0.

(a) WRT F2C, on the one hand, it bothers me to maintain compatibility with a software (F2C) that, I think, is not maintained anymore. On the other hand, we do not have to purposely go out of our ways to break compatibility. And it is true that [testBLAS](https://github.com/tlapack/testBLAS) tests for NaNs and Infs. I'd rather be self-content (and break F2C), but we do not have to do this now. 

Note: One reason some NaNs and Infs checks were moved to testBLAS (as opposed to the BLAS test suite) was that the behavior for some BLAS subroutine was not defined in the presence of Infs and Nans, and so we were leery to move this to the BLAS test suite. But it feels like for NRM2, the expected behavior with NaNs and Infs is clear. (There were other reasons for testBLAS, namely, for example, a more modern test suite.)

(b) WRT INCX = 0, it feels like if this is a valid input this should be tested. I think that [testBLAS](https://github.com/tlapack/testBLAS) tests for INCX=0. So maybe this can wait too.

angsch:
Makes sense. I will fiddle in `incx=0` and let's see if someone has a strong opinion on `f2c`.

angsch:
- `incx=0` is added to the tests
- I've tested with gfortran and ifort that that the tests trigger if the NRM2 implementation is flawed 

My only remaining concern is if the NaN check, currently inlined as `a.ne.a` should be done in a separate routine. LAPACK's `xISNAN` does so to avoid overly aggressive compiler optimization according to the comment. In my tests the inlined check worked.

langou:
> incx=0 is added to the tests. I've tested with gfortran and ifort that that the tests trigger if the NRM2 implementation is flawed

Thanks!

> My only remaining concern is if the NaN check, currently inlined as a.ne.a should be done in a separate routine. LAPACK's xISNAN does so to avoid overly aggressive compiler optimization according to the comment. In my tests the inlined check worked.

Right. Maybe we just add a comment in the TESTING then to explain that this can be improved. I think, for now, it is enough to leave at that. That is leave it at "inline a.ne.a" for checking for NaNs. 

Note that testing for NaNs in BLAS is also done by the project https://github.com/tlapack/testBLAS

langou (dismissed):

weslleyspereira:
> Note that testing for NaNs in BLAS is also done by the project https://github.com/tlapack/testBLAS

And we are trying to get the tests working with LAPACK's continuous testing framework, see #913.

angsch:
Based on my testing, everything is fine if the tests are compiled even with `-O3`. Some NaN checks appear to be dropped if the tests are compiled with `-ffast-math`, which is documented as unsafe optimizations. The code now has a comment that the NaN checks rely on a not too aggressive compiler. I guess it is recommended to compile the tests at a moderate optimization level (`-O2`) and the code to be tested with whatever flags are suited.

langou (approved):

langou:
Thanks @angsch!

Issue #917, opened 2023-10-26
SORCSD2BY1 computes inaccurate result


christoph-conrads:
**Description**

Given a 2-by-1 block matrix `Q` with orthonormal columns, SORCSD2BY1 computes matrices `U1`, `U2`, `Σ1`, `Σ2`, and `V` such that
```
|Q11|   |U1   | |Σ1|
|   | = |     | |  | V^T
|Q21|   |   U2| |Σ2|
```

In commit 4174d8d080199d30c0c834a959752bc361b91cbe, SORCSD2BY1 computes an inaccurate result with a backward error of `||Q - U Σ V^T|| ≈ 10^-5` when it should be at most `ε ||Q|| ≈ 10^-7`. Input with `m = 3`, `n = 2`, `p = 2` (Q11 ∈ ℝ(p, n), Q21 ∈ ℝ(m - p, n)):
```
    [-2.0392263e-01 -9.7898704e-01]
Q = [ 1.1427624e-08  9.2925374e-09]
    [ 9.7898704e-01 -2.0392257e-01]
```

**Checklist**

- [x] I've included a minimal example to reproduce the issue
- [x] I'd be willing to make a PR to solve this issue

This issue was found while working on #406.

christoph-conrads:
The Python code below demonstrates the problem. It requires Python 3, NumPy, and a shared LAPACK library.

Call it with `python3 lapack-issue-917.py` and optionally pass the path to the LAPACK shared library. Example (the first execution uses the system's LAPACK installation, the second execution uses my LAPACK build of the latest master branch code):
```console
$ python3 lapack-issue-917.py 
name of LAPACK library: liblapack.so.3
tolerance: 1.69e-07
actual value: 2.97e-05 (176.0 times tolerance)
input matrix:
[[-2.0392263e-01 -9.7898704e-01]
 [ 1.1427624e-08  9.2925374e-09]
 [ 9.7898704e-01 -2.0392257e-01]]
reassembled matrix:
[[-2.0392254e-01 -9.7898668e-01]
 [ 6.0102630e-06  2.9068287e-05]
 [ 9.7898704e-01 -2.0392255e-01]]
$ python3 lapack-issue-917.py /tmp/tmp.y7q9ZOrOZA/lib/liblapack.so.3
name of LAPACK library: /tmp/tmp.y7q9ZOrOZA/lib/liblapack.so.3
tolerance: 1.69e-07
[...]

PR #928, opened 2023-11-06
xORBDB6/xUNBDB6: fix a constant

christoph-conrads:
**Description**

The commits listed below updated xORBDB6/xUNBDB6 with the aim of improving numerical stability by avoiding the use of squared values. With the exception of SORBDB6 these commits did not properly up a magic constant called `ALPHA` in the code. SORBDB6 was initially updated correctly before commit a015b21445ad6f43746ced7615438fe6c82a1b66 inserted an incorrect value again.

The incorrect magic constant may lead to early returns with only partially orthogonalized output vectors.

Relevant commits:
* 54b3964b0f6879b22d1587f821a50f512c2dd10c
* 6479e0f53116987e4898aa5dc98a842ffbeae38d
* 94419e8bfae9212e44ac5acf24031b83d84f9251

**Checklist**

### DIFF
--- a/SRC/cunbdb6.f
+++ b/SRC/cunbdb6.f
@@ -174,7 +174,7 @@
 *
 *     .. Parameters ..
       REAL               ALPHA, REALONE, REALZERO
-      PARAMETER          ( ALPHA = 0.01E0, REALONE = 1.0E0,
+      PARAMETER          ( ALPHA = 0.1E0, REALONE = 1.0E0,
      $                     REALZERO = 0.0E0 )
       COMPLEX            NEGONE, ONE, ZERO
       PARAMETER          ( NEGONE = (-1.0E0,0.0E0), ONE = (1.0E0,0.0E0),
--- a/SRC/dorbdb6.f
+++ b/SRC/dorbdb6.f
@@ -174,7 +174,7 @@
 *
 *     .. Parameters ..
       DOUBLE PRECISION   ALPHA, REALONE, REALZERO
-      PARAMETER          ( ALPHA = 0.01D0, REALONE = 1.0D0,
+      PARAMETER          ( ALPHA = 0.1D0, REALONE = 1.0D0,
      $                     REALZERO = 0.0D0 )
       DOUBLE PRECISION   NEGONE, ONE, ZERO
       PARAMETER          ( NEGONE = -1.0D0, ONE = 1.0D0, ZERO = 0.0D0 )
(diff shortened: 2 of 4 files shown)


christoph-conrads:
Input triggering the bug fixed in this PR can be found in the linked post: https://github.com/Reference-LAPACK/lapack/issues/917#issuecomment-1792441290.

langou (approved):

langou:
Thanks @christoph-conrads. ALPHA = 1 / 100 is certainly safer than ALPHA = 1/10.

christoph-conrads:
> Thanks @christoph-conrads. ALPHA = 1 / 100 is certainly safer than ALPHA = 1/10.

`ALPHA` is now one tenth, that is `ALPHA` is larger. The constant is used in two places:
* to check if an early return is possible (see [line 261ff. in CUNBDB6](https://github.com/Reference-LAPACK/lapack/blob/master/SRC/cunbdb6.f#L261-L263))
* to check if the vector `x` is practically zero after the second Gram-Schmidt iteration (see [line 307ff. in CUNBDB6](https://github.com/Reference-LAPACK/lapack/blob/master/SRC/cunbdb6.f#L307-L314))

Moreover the value was not really changed. In the past it was used in comparisons involving squares of norms (e.g., as in `ALPHASQ = 0.01; ALPHASQ ≤ ∥x∥²`); the current code avoids the use of squares (e.g., as in `ALPHA = 0.1; ALPHA ≤ ∥x∥`).

langou:
https://github.com/Reference-LAPACK/lapack/blob/8ecaaf9869b6bb95ed280f6f54ae5ae061eeeded/SRC/dorbdb6.f#L255-L263

Thanks for correcting me. 

ALPHA = 1/100 leads to a loss of accuracy of about 100x. ALPHA = 1/10 is better. Sure. Good. We can continue stir ALPHA down. A standard value, in the Gram-Schmidt context, is ALPHA = 1/sqrt(2). (But we are not in the Gram-Schmidt context.) All in all, I am not against setting ALPHA to ZERO so as to force reorthogonalization.

Since, this projection (maybe with reorthogonalization) is ``one and done``, as opposed to repeatedly project columns one after the other as in the case in Gram-Schmidt, the value of ALPHA is a little different. 

Taking an ALPHA ``small`` such as 1/10 should not be too much of an issue in this case. (Only 10x loss of accuracy at most.)

All in all though, I am not against setting ALPHA to ZERO so as to force reorthogonalization.

PR #936, opened 2023-11-11
make the DMD comments in the header Doxygen compliant

langou:
This should make the header comments of the eight DMD subroutines doxygen complaints.


weslleyspereira (approved):
Thanks!

Issue #546, opened 2021-05-08
xORM{QR,LQ,QL, xUNMQR optimal workspace size is always > 4096


christoph-conrads:
**Description**

The functions multiplying elementary reflectors with matrices called
* {S,D}ORMQR,
* {S,D}ORMRQ,
* {S,D}ORMLQ,
* {S,D}ORMQL,
* {C,Z}UNMQR,
* {C,Z}UNMRQ,
* {C,Z}UNMLQ,
* {C,Z}UNMQL

return at least 4096 for the optimal workspace size even for dimensions as tiny as zero or one. My expectation is that the optimal workspace size is less than 4096 for tiny numbers of columns and rows. The behavior occurs for SORMQR because the optimal workspace size is computed as
```fortran
LWKOPT = NW*NB + TSIZE
```
where `TSIZE = LDT * NBMAX = 64 * 65 = 4160`.

Relevant initialization code:
https://github.com/Reference-LAPACK/lapack/blob/2dafa3d2756a7825c23a8c8456781561e36668ae/SRC/sormqr.f#L185-L187

Workspace computation:
https://github.com/Reference-LAPACK/lapack/blob/2dafa3d2756a7825c23a8c8456781561e36668ae/SRC/sormqr.f#L241-L249

Demo program in C++11 (compile with `c++  -Wextra -Wall -pedantic -std=c++11 sormqr-workspace.cpp -llapack -L/path/to/lapack-build/lib`, run with `env LD_LIBRARY_PATH=/tmp/path/to/lapack-build/lib ./a.out`):
```c++
#include <cassert>
#include <complex>
#include <cstddef>
#include <cstdint>
#include <cstdio>
#include <cstdlib>

#include <algorithm>
#include <limits>

#define lapack_complex_float float _Complex
#define lapack_complex_double double _Complex

extern "C" {

typedef int32_t lapack_int;

void dormqr_(char const* side, char const* trans, lapack_int const* m,
    lapack_int const* n, lapack_int const* k, double const* A,
    lapack_int const* lda, double const* tau, double* C, lapack_int const* ldc,
    double* work, lapack_int const* lwork, lapack_int* info, size_t side_len,
    size_t trans_len);

void sormqr_(char const* side, char const* trans, lapack_int const* m,
    lapack_int const* n, lapack_int const* k, float const* A,
    lapack_int const* lda, float const* tau, float* C, lapack_int const* ldc,
[...]

Issue #257, opened 2018-06-18
GEQRF: document corner case of LWORK


amigalemming:
I expected that a workspace query would always return a valid LWORK. This is not the case, a workspace query can yield LWORK=0. Should a warning be added to the documentation of the workspace query?

ilayn:
Can you also add which query leads to LWORK=0?

amigalemming:
But `INFO` is set to `-7`.

ilayn:
Ah that can be too. Netlib site is down so I can't look at the source code but either it will raise an error or quick exit due to N=0, depends on which one is checked first. In both cases WORK is not addressed if I remember correctly.

amigalemming:
Even on quick return the first element of `WORK` is written:
https://github.com/Reference-LAPACK/lapack/blob/master/SRC/sgeqrf.f

ilayn:
There are two cases in the first one `LWORK=-1` and `N = 0`. In this case, WORK(1) will be set to 0 and calling SGEQRF with this `N=0` and `LWORK = 0` will lead to `INFO=-7`. That is the expected outcome since `LWORK < MAX(1, N)`

In the second case assume we used an arbitrary number `X > 1` but `N=0` still then in the following block 

```
*
*     Quick return if possible
*
      K = MIN( M, N )
      IF( K.EQ.0 ) THEN
         WORK( 1 ) = 1
         RETURN
     END IF
```
This will cause the quick exit. That will be the result because A has a shape `(LDA, 0)` hence nothing is needed to be modified and `TAU` has the shape 0 again nothing is needed. Initially work size is unspecified `( * )` but its specification is that it will have the shape `(MAX(1,LWORK))` hence first element is written so I don't see the problem of being corner case.

weslleyspereira:
I believe this issue was fixed by #638. Maybe we want to let this open to remember that the problem still persists for several other routines, e.g., gelqf, gebrd, hetrd and gehrd.

dklyuchinskiy:
> I believe this issue was fixed by #638. Maybe we want to let this open to remember that the problem still persists for several other routines, e.g., gelqf, gebrd, hetrd and gehrd.

This is actual, indeed. I have recently met this problem during `gebrd/sytrd` usage inside other computational routine. In this case you need to remember, that `lwork` can be `0` for some corner cases, which will lead to memory problems. And user should always set `lwork = max( 1, dummy[0] );` which is inconvenient.

PR #942, opened 2023-11-21
handle and document corner cases of lwork in lapack

dklyuchinskiy:
**Handle and document corner cases of lwork in lapack**

This PR contains fixes related to the following issues:
1. #257 - parameter LWORK can't be zero on any case, info checker should know about this
2. #257, #546 - parameter LWORK should be small as possible for corner cases M = 0 or N  = 0.  Final LWORK value should be aligned with quick exit and info checker.
3. Some typos and code style fixes, changes in documentation.

Actually, for many functions in LAPACK these issues are already solved or did not exist at all. This PR handles remaining ones for general, symmetric and generalized problems (double precision only at the moment). Please review. If there are some questions, let's discuss!

### DIFF
--- a/SRC/cgebrd.f
+++ b/SRC/cgebrd.f
@@ -252,7 +252,7 @@
       ELSE
          LWKMIN = MAX( M, N )
          NB = MAX( 1, ILAENV( 1, 'CGEBRD', ' ', M, N, -1, -1 ) )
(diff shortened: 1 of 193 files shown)


langou (approved):

dklyuchinskiy:
Extension to all precisions for this scope of functions is in progress.

PR #959, opened 2023-12-12
fix some DMD issues

dklyuchinskiy:
**fix some DMD issues**

1. add DMD tests into CMakeLists
2. init variable SSUM before call of LASSQ to prevent error exit with `info = -8` or `info = -10`

Related to #960.


langou (approved):

langou:
Thanks @dklyuchinskiy. Good catch!

Issue #963, opened 2023-12-20
[v3.12.0] sblat1.f (SNRM2) test fail on i386 architectures


cdluminate:
**Description**

The `sblat1.f` test fails on i386 architectures, including i386 (effectively i686), and hurd-i386:
https://buildd.debian.org/status/package.php?p=lapack&suite=experimental
https://buildd.debian.org/status/fetch.php?pkg=lapack&arch=i386&ver=3.12.0-1&stamp=1702247469&raw=0
https://buildd.debian.org/status/fetch.php?pkg=lapack&arch=hurd-i386&ver=3.12.0-1&stamp=1702271309&raw=0

The typical failure is the `SNRM2` call, which look like:
```
 Test of subprogram number  7            SNRM2 
                                       FAIL
  SNRM2: N=     4, INCX=  -2, IV= 9, IW= 9, test= 0.20971520E+07
                                       FAIL
  SNRM2: N=     4, INCX=   1, IV= 9, IW= 9, test= 0.20971520E+07
                                    ----- PASS -----
```

I'm confused about why this failure does not occur on other 32-bit architectures such as armhf and m68k. Any idea?
It does not seem like a compiler issue neither. Both gfortran-12 and gfortran-13 lead to the same results.

The minimal example to reproduce this issue is to build this package on a clean Debian unstable (sid) chroot.

**Checklist**

- [x] I've included a minimal example to reproduce the issue
- [x] I'd be willing to make a PR to solve this issue

angsch:
There is a typo in the test that on most systems apparently yields a zero, but not on the two systems above.

@cdluminate  I sent you an email.

cdluminate:
BTW, the following is the default build flags for Debian (distribution-wide):
```
CFLAGS=-g -O2 -ffile-prefix-map=/home/lumin=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection
CPPFLAGS=-Wdate-time -D_FORTIFY_SOURCE=2
FFLAGS=-g -O2 -ffile-prefix-map=/home/lumin=. -fstack-protector-strong -fstack-clash-protection -fcf-protection
LDFLAGS=-Wl,-z,relro
```
[...]

PR #964, opened 2023-12-21
Fix the wrong implementation of the new tests for xblat1.f

cdluminate:
**Description**

Fixes: #963

Since this is a regression after upgrading from v3.11.0 to v3.12.0, we can narrow down the range of the bug into the newly added SB1NRM2 subroutine. According to the buildlog and the documentation in the code, the VALUES(9), calculated as SXVALS(XX,2) should be infty. But the current code is returning a zero (or randomly) initialized variable YY, which does not make sense.

In fact, if you go back to the reference implementation, namely the supplementary material of this paper
  https://dl.acm.org/doi/abs/10.1145/3061665
You can find a similar implementation of the SXVALS function in the `la_xxvals.F90` file. This patch corrests the test following the reference code.

**Checklist**

### DIFF
--- a/BLAS/TESTING/cblat1.f
+++ b/BLAS/TESTING/cblat1.f
@@ -994,13 +994,17 @@
 *     .. Scalar Arguments ..
       REAL              XX
       INTEGER           K
+*     .. Parameters ..
+      REAL              ZERO
+      PARAMETER         (ZERO=0.0E+0)
 *     .. Local Scalars ..
-      REAL              X, Y, YY, Z
+      REAL              X, Y, Z
 *     .. Intrinsic Functions ..
       INTRINSIC         HUGE
 *     .. Executable Statements ..
+      X = ZERO
       Y = HUGE(XX)
-      Z = YY
+      Z = Y*Y
       IF (K.EQ.1) THEN
          X = -Z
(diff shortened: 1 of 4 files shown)


cdluminate:
looks good? @angsch

BLAS/TESTING/cblat1.f
```diff
@@ -995,12 +995,13 @@ REAL FUNCTION SXVALS(XX,K)
       REAL              XX
       INTEGER           K
 *     .. Local Scalars ..
-      REAL              X, Y, YY, Z
+      REAL              X, Y, Z
 *     .. Intrinsic Functions ..
       INTRINSIC         HUGE
```
angsch, line 1000: Add declaration of `ZERO`
```
*     .. Parameters ..
      REAL              ZERO
      PARAMETER         (ZERO=0.0E+0)
```

angsch:
Thank you for the fix. :+1:

BLAS/TESTING/cblat1.f
cdluminate: fixed.

cdluminate:
I have addressed the comments and force-pushed.

langou (approved):

langou (approved):

PR #971, opened 2024-01-08
Re-sync Makefile with CMake

turboencabulator:
The list of source files listed in the Makefile vs. CMakeLists.txt has gotten out-of-sync.  Doing a selective build using the Makefiles will give different results than CMake.  (Just to be clear, building all precisions should be OK.)  This set of changes fixes most of the differences.

Comparing the two was challenging due to how the lists of source files are formatted/organized, and I wanted to make this process easier in case we need to check it again later.  The first commit changes the formatting to one source file per line; there shouldn't be any change to the order of files (see the last commit for that).  I'm doing a side-by-side diff after stripping off the .f/.o file suffixes to compare the two.

All of the differences I found (other than differences in the order of files) were introduced in #637.  That PR didn't update the Makefile to match CMake, so I've done it here with a few fix-ups.  That change was to fix missing dependencies when building only one precision, basically rearranging some needed files to different source groups so they'll get built.

(That PR also adds some differences to the BLAS build scripts which I'm not going to touch in this PR.  From what I can tell, LAPACK built for one precision might need BLAS routines from another precision, so it's now building at least the needed files from BLAS.  So now if you want to build only BLAS and only specific precisions of it, you might get more than you asked for.  I think the proper fix is that BLAS may need to be built for all precisions even if LAPACK is built with only one.)

### DIFF
--- a/SRC/CMakeLists.txt
+++ b/SRC/CMakeLists.txt
@@ -47,7 +47,10 @@ set(ALLAUX
    ilauplo.f
    iladiag.f
    chla_transtype.f
+   la_constants.f90
    la_xisnan.F90
+   sisnan.f
+   slaisnan.f
    ../INSTALL/ilaver.f
    ../INSTALL/lsame.f
    xerbla.f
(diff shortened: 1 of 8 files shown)


langou (approved):

langou:
Thanks. I agree that this will be better like this moving forward. Thanks for your work

PR #972, opened 2024-01-08
CMake build: Add -DLAPACK option, un-break CBLAS+BLAS build, do not require C++ if unnecessary

drhpc:
This fixes some build issues I face when packaging for pkgsrc using the CMake build.

- Adds option to build without LAPACK, so that any of BLAS, LAPACK, CBLAS, LAPACKE can be built as individual library from source.
- Fixes CBLAS build missing blas-targets for CMake by adding that. Note that I do not use those cmake files and scrub them from the install. Would like an option to avoid that machinery, as we use pkg-config also for CMake to locate BLAS deps (BLA_PREFER_PKGCONFIG in FindBLAS).
- Avoids requiring a C++ compiler when not needed.
- Also fixes macro usage for BLAS++ / LAPACK++.

I hope that can be merged in one go. Though, I have to admit that I am no CMake expert and someone might find further errors. But this works for me at least.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.6)
 
-project(LAPACK)
+project(LAPACK Fortran C)
 
 set(LAPACK_MAJOR_VERSION 3)
 set(LAPACK_MINOR_VERSION 12)
@@ -424,6 +424,7 @@ function(_display_cpp_implementation_msg name)
   message(STATUS "----------------")
 endfunction()
 if (BLAS++)
+  enable_language(CXX)
   _display_cpp_implementation_msg("BLAS")
   include(ExternalProject)
   ExternalProject_Add(blaspp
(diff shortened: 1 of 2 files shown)


drhpc:
Does https://github.com/Reference-LAPACK/lapack/pull/834 mean that LAPACKE may be built without Fortran … but only LAPACKE? This PR might need some adaptation to only enable Fortran for BLAS, LAPACK … and CBLAS?

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.6)
 
-project(LAPACK)
+project(LAPACK Fortran C)
```
quant, line 3: With the project language specified here, statements `enable_language(Fortran)` at lines 226 and 301 are not needed anymore.

CMakeLists.txt
```diff
@@ -278,6 +282,10 @@ endif()
 
 option(USE_OPTIMIZED_LAPACK "Whether or not to use an optimized LAPACK library instead of included netlib LAPACK" OFF)
 
+option(LAPACK "Whether to build or use LAPACK (to enable a BLAS-only build)")
```
quant, line 285: Please be explicit, do not omit `OFF`, even though it is the default.

PR #989, opened 2024-02-25
Two small fixes of bdsqr

angsch:
Two small fixes of bdsqr:

1) In https://github.com/Reference-LAPACK/lapack/pull/234 only the documentation of `dbdsqr` was updated. This  updates the documentation of the remaining precisions.

2) `bdsqr` can return `-ZERO` as a singular value. For example, `[d,z]bdsqr` on

```
D = [ -4.0, -3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0, 4.0, 5.0]
E = [ 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0]
```
returns `D(10) = -ZERO`. This adds a branch enforcing that all singular values follow the definition and are non-negative, i.e., `-ZERO` is converted into `+ZERO`. Note that this is already enforced in `lasq1` (used for the singular-values-only path).

### DIFF
--- a/SRC/cbdsqr.f
+++ b/SRC/cbdsqr.f
@@ -166,7 +166,7 @@
 *>
 *> \param[out] RWORK
 *> \verbatim
-*>          RWORK is REAL array, dimension (4*N)
+*>          RWORK is REAL array, dimension (4*(N-1))
 *> \endverbatim
 *>
 *> \param[out] INFO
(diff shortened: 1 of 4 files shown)


thijssteel:
I approve of this change. However, @langou and I were recently discussing something similar about -ZERO and ZERO. Since we cannot normally distinguish between the two, for example:
- ``` -0. == 0.``` evaluates to true.

However, ```sign(1., -0.) == -1.0``` in gfortran, which could definitely cause problems so I support this change.

Issue #988, opened 2024-02-23
Should we allow C++?


thijssteel:
I have been working on a pet project. What is would like to do is introduce some C++ code in LAPACK, with a matrix class to represent the matrices, not just a pointer. This is not meant as a way for people to have a nice API to call LAPACK (there are plenty of libraries that provide nice wrappers), but to make things easier for developers in the future. It would allow for comprehensive bounds checks that even work for subarray (via asserts of course, we should disable those in a production build).

The way I want to achieve this is by mixing C++ and Fortran, so that we don't have to suffer through the immense task it would be to translate all of LAPACK at once. We could just stick to C++ for new algorithms or for routines that need to be reworked/debugged (for example, dgesdd, see #672)

A proof of concept implementation is available at https://github.com/thijssteel/lapackv4

I appreciate all feedback.

martin-frbg:
I suspect this will create portability problems for some - finding a Fortran compiler for some platforms is hard enough, but having both Fortran and C++ as dependencies in a project just from the reference implementation of a standard API may be a bit much (and by the time everything got rewritten, users may have moved on to newer APIs that drop the Fortran naming constraints)

thijssteel:
Thanks for the feedback

> finding a Fortran compiler for some platforms is hard enough

While this is true, I kind of assumed that most platforms that have a Fortran compiler also have a C++ compiler, maybe not a cutting edge one with C++20 support, but we could restrict which features we use. Do you have examples? I assume you know more about those issues than me from your work on OpenBLAS.

ilayn:
I would never thought I would see this day 😃 and I am very confused about this. I do feel that indeed getting out of F77 is an absolute win. However going into C++ scares me and smells like yet another trap for 10-years-later ourselves regretting it. 
Newer Fortran90 is there but I think you folks already considered that option anyways and not sure if that solves the issue you already mentioned in the issue so I'll skip that.

Definitely not want to go into a language flamewar at all but C++ does not feel like an archival language like Fortran or C is and it is getting larger and weirder with each standard update. However there are many many industry grade C++ code hence I would refrain from dismissing it. Having said that, in the same breath, I would also agree that C++ is not an array friendly number crunching language.

[...]

PR #991, opened 2024-02-29
Introduce cpp (and also adds lasr3, which applies a sequence of rotations efficiently)

thijssteel:
related to issue #988

This PR introduces some C++ code to LAPACK.

The general design for the C++ code is:
- C wrappers on top of existing fortran code
- C++ wrappers on top of those C wrappers
- C++ code for the new algorithm and its tests (note, those tests are not automated, so definitely need improving)
- C wrappers around the new C++ code
- Fortran wrappers around those C wrappers so current users can still use a familiar API.

A lot of things are introduced in this PR:
- Vector and Matrix classes, these are thin wrappers around a pointer and a size, allowing things like verbose selection of submatrices, columns, ..., all with bounds checks for every step.
- LASR3, an algorithm which efficiently applies a sequence of plane rotations to a matrix. (Note, this applies multiple sequences, making is more a level 3 routine than lasr, which is more a level 2 routine, hence the name) This algorithm is described in the paper: "Restructuring the Tridiagonal and Bidiagonal QR Algorithms for Performance" F. G. Van Zee et al. I have made a modification of this algorithm which packs the matrix A into an aligned workspace for efficiency. This means we also achieve high performance for matrices that are not aligned and even for access patterns that are normally not efficient (applying a sequence from the left on a column major matrix).

I expect (or hope) a lot of people will have comments on whether we should use C++ and on my implementation. To keep the discussions easy to follow, I suggest that conceptual comments on whether or not we should use C++ are direction to issue #998, comments on the specific way I introduced C++ and my implementation of the algorithm should be posted here.

### DIFF
--- a/LAPACK_CPP/src/lapack_cpp/rot_cpp.cpp
+++ b/LAPACK_CPP/src/lapack_cpp/rot_cpp.cpp
@@ -1,0 +1,67 @@
+#ifdef USE_FORTRAN_BLAS
+    #include <cassert>
+    #include <complex>
+    #include <type_traits>
+
+    #include "lapack_c.h"
+    #include "lapack_cpp/base.hpp"
+    #include "lapack_cpp/blas/rot.hpp"
+
+namespace lapack_cpp {
+
+/**
+ * Templated wrapper around c functions, will be instantiated for each type.
+ *
+ * @tparam T
+ * @param A
+ * @param B
+ * @param C
+ */
+template <typename T, typename TC, typename TS, typename idx_t>
+inline void rot_c_wrapper(const Vector<T, idx_t>& x,
+                          const Vector<T, idx_t>& y,
+                          TC c,
+                          TS s)
+{
+    assert(x.size() == y.size());
+
+    if constexpr (std::is_same<T, double>::value) {
+        lapack_c_drot(x.size(), x.ptr(), x.stride(), y.ptr(), y.stride(), c, s);
+    }
+    else if constexpr (std::is_same<T, float>::value) {
+        lapack_c_srot(x.size(), x.ptr(), x.stride(), y.ptr(), y.stride(), c, s);
+    }
+    else if constexpr (std::is_same<T, std::complex<float>>::value) {
+        lapack_c_crot(x.size(), x.ptr(), x.stride(), y.ptr(), y.stride(), c, s);
+    }
+    else if constexpr (std::is_same<T, std::complex<double>>::value) {
+        lapack_c_zrot(x.size(), x.ptr(), x.stride(), y.ptr(), y.stride(), c, s);
+    }
+    else {
+        assert(false);
+    }
+}
+
+    // We have a lot of types to instantiate for, so we use a macro to avoid
+    // repetition.
+    #define INSTANTIATE_ROT(T, TC, idx_t)                                    \
+        template <>                                                          \
+        void rot(const Vector<T, idx_t>& x, const Vector<T, idx_t>& y, TC c, \
+                 T s)                                                        \
+        {                                                                    \
+            rot_c_wrapper(x, y, c, s);                                       \
+        }
+
+INSTANTIATE_ROT(float, float, lapack_idx_t)
+INSTANTIATE_ROT(double, double, lapack_idx_t)
+INSTANTIATE_ROT(std::complex<float>, float, lapack_idx_t)
+INSTANTIATE_ROT(std::complex<double>, double, lapack_idx_t)
+INSTANTIATE_ROT(float, float, int)
+INSTANTIATE_ROT(double, double, int)
+INSTANTIATE_ROT(std::complex<float>, float, int)
+INSTANTIATE_ROT(std::complex<double>, double, int)
+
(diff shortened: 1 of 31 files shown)


LAPACK_CPP/src/lapack_cpp/rot_cpp.cpp
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Templated wrapper around c functions, will be instantiated for each type.
+ *
+ * @tparam T
+ * @param A
+ * @param B
+ * @param C
+ */
+template <typename T, typename TC, typename TS, typename idx_t>
+inline void rot_c_wrapper(const Vector<T, idx_t>& x,
```
christoph-conrads, line 21: Have you considered overloading? The advantage of overloading at this point would be that the user can add his own overloads:
```c++
inline void rot_c_wrapper(const Vector<float, idx_t>& x, const Vector<float, idx_t>& y, TC c, TS s) {
    assert(x.size() == y.size());
    return lapack_c_srot(x.size(), x.ptr(), x.stride(), y.ptr(), y.stride(), c, s);
}
inline void rot_c_wrapper(const Vector<double, idx_t>& x, const Vector<double, idx_t>& y, TC c, TS s) {
    assert(x.size() == y.size());
    return lapack_c_drot(x.size(), x.ptr(), x.stride(), y.ptr(), y.stride(), c, s);
}
...
```

LAPACK_CPP/src/lapack_cpp/rot_cpp.cpp
```diff
@@ -0,0 +1,67 @@
...
+        lapack_c_srot(x.size(), x.ptr(), x.stride(), y.ptr(), y.stride(), c, s);
+    }
+    else if constexpr (std::is_same<T, std::complex<float>>::value) {
+        lapack_c_crot(x.size(), x.ptr(), x.stride(), y.ptr(), y.stride(), c, s);
+    }
+    else if constexpr (std::is_same<T, std::complex<double>>::value) {
+        lapack_c_zrot(x.size(), x.ptr(), x.stride(), y.ptr(), y.stride(), c, s);
+    }
+    else {
+        assert(false);
```
christoph-conrads, line 41: The code below fails at compile time:
```c++
static_assert(!std::is_same<T, T>::value, "unknown data type T")
```

References:
* [_How to assert that a constexpr if else clause never happen?_](https://stackoverflow.com/questions/53945490/how-to-assert-that-a-constexpr-if-else-clause-never-happen)

LAPACK_CPP/src/lapack_cpp/rot_cpp.cpp
thijssteel: `rot_c_wrapper` is not meant to be user callable. It is just a utility so we can call it when instantiating the function we actually want to overload: `rot`. It is a pattern I designed for gemm, where the logic is more complicated. I guess having this function at all does not make that much sense here.

christoph-conrads:
> comments on the specific way I introduced C++ and my implementation of the algorithm should be posted here.

C++11 introduced the `auto` keyword. Throughout the code you use the C++98 pattern though, e.g.,
```c++
const idx_t n_timings = 100;
```
In C++11 style I would write:
```c++
const auto n_timings = idx_t{100};
```
The C++11 style avoids all coercions and non-representable values cause compiler errors. With C++98 style non-representable value are silently truncated/rounded/etc; GCC 12 with `-Wextra -Wall` prints no warnings whereas Clang 14 prints warnings when this happens. For this reason and due to the more uniform syntax in combination with the `using` keyword (`using UC = unsigned char;`) I prefer the C++11 style. [The C++11 style does not work though when the type name contains a space like `unsigned char`.]

What are your thoughts about this?

christoph-conrads:
> comments on the specific way I introduced C++ and my implementation of the algorithm should be posted here.

I can'tquite wrap my head around the file names, specifically the repeated `cpp` in `src/lapack_cpp/rot_cpp.cpp` and the mixed language indicator in `src/lapack_c/rot_c.f90`.

I drafted a CMake build in christoph-conrads/introduce-cpp that you may be interested in.

thijssteel:
> I can'tquite wrap my head around the file names, specifically the repeated cpp in src/lapack_cpp/rot_cpp.cpp and the mixed language indicator in src/lapack_c/rot_c.f90

The file names are mostly to make them unique, so that someone exporting all the .o files to a single obj folder does not run into issues. 

The lapack_c defines all the c interfaces, it just happens to be a lot easier to define those interfaces in fortran than in c. But perhaps the name can be improved.

thijssteel:
> C++11 introduced the auto keyword. Throughout the code you use the C++98 pattern though

I was not familiar with that style. My opinion on it is mixed. I can try to start using it, but i worry about enforcing language specific things on that level may waste a lot of time. I also feel that way about move vs refernce. On the other hand, if we start from a good example, it may not be a large cost

thijssteel:
I've found a pretty significant issue with this PR. I rely on template instantiation in the cpp files, but that may fail if the original function was inlined. I am not entirely sure how to get around that.

Issue #244, opened 2018-03-05
DROTMG returns unexpected values


vladimir-ch:
In some cases, DROTMG returns values that, when used in DROTM, do not zero out the second component of the given vector.

This can be observed for example with the following input values for DROTMG:
```
D1 = 1600000000
D2 = 800000000
X1 = 8
Y1 = 7
```
After DROTM, Y1 is currently mapped to -24052.672896090524.

The issue we see is in the rescaling part. The setup phase sets the non-unit elements of H. When rescaling is done, the unit elements are set explicitly before the scaling. Rescaling for D1 scales H11,H12 and sets FLAG=-1 but then rescaling for D2 overwrites H12 with 1 which is wrong.

In Gonum we fixed the issue by adopting in https://github.com/gonum/gonum/pull/420 the implementation proposed in:
```
Tim Hopkins. 1997. Restructuring the BLAS level 1 routine for computing the modified
givens transformation. SIGNUM Newsl. 32, 4 (October 1997), 2-14.
DOI=http://dx.doi.org/10.1145/289251.289253
```
OpenBLAS later fixed their rotmg implementation with the same approach in https://github.com/xianyi/OpenBLAS/pull/1480.
There was also some discussion in https://github.com/xianyi/OpenBLAS/issues/1452 where the initial report about strange values returned by DROTMG was done.

Also, as I comment in that thread, we would like to clarify whether the value of `GAM` could be increased and to what value. From what I read in the above paper, the current value of 4096 was chosen very conservatively when BLAS Level 1 was created, that is before IEEE floating point. A larger value of GAM would reduce the necessity for rescaling (flag = -1) which is good because having units in H saves operations in DROTM which is the whole point of DROTMG in my understanding. However, it is not clear what an appropriate value would be. What are your views on this?

Lastly, it would worth revising the documentation for DROTMG and thus improve slightly its status of an underdog. It took us some time to fully understand how to interpret the values the routine takes and returns. The most mysterious was the square root of D1 and D2, some projects that provide interface to BLAS even drop the square root from their documentation.

jacobbogers:
Hi, can anyone give the status on this ussue, is it resolved?

jacobbogers:
What is the status on this issue?

langou:
This might be related to #344

PR #1000, opened 2024-03-27
drotmg.f: Replace GAM**2 by existing GAMSQ

SebWouters:


langou (dismissed):

christoph-conrads:
> Also, for the DD1 = DD1/GAM**2, instead of dividing by GAMSQ, maybe we should multiply by RGAMSQ.

In the double precision code, `RGAMSQ` would have to be updated with an accurate value. At the moment a single-precision value with 7 decimal digits is used.

christoph-conrads:
> Also, for the DD1 = DD1/GAM**2, instead of dividing by GAMSQ, maybe we should multiply by RGAMSQ.

Does Fortran have [`ldexp(base, exp)`](https://linux.die.net/man/3/ldexp) which computes `base · 2^exp`? One could replace expressions `X * GAM**2` with `ldexp(x, 24)` and `X / GAM**2` with `ldexp(x, -24)`. This would keep the code readable and possibly enable legitimate compiler optimizations.

langou:
> "Also, for the DD1 = DD1/GAM**2, instead of dividing by GAMSQ, maybe we should multiply by RGAMSQ." => In the double precision code, RGAMSQ would have to be updated with an accurate value. At the moment a single-precision value with 7 decimal digits is used.

Yes, of course, you are correct. This is a good point. Thanks. The 5.9604645D-8 is not a good approximation! In base 10, RGAMSQ is 5.9604644775390625D-08, I think that this base 10 representation will be mapped exactly to RGAMSQ (which is 2^(-24)) in the computer in double precision arithmetic. I did not check.

christoph-conrads:
xROTMG was introduced in 1979 in [ACM Algorithm 539](https://dl.acm.org/doi/10.1145/355841.355848). The linked page contains a freely available tarball with the Fortran77 source code (open `Original/539.txt`). In my opinion the code should either be kept in its current form to make its origins crystal clear or the code should be simplified as described in the next paragraph.

Fortran supports exponentiation in its syntax. My suggestion is to use this notation and to let the compiler take care of optimizing expressions like `x / GAM`, `x * GAM**2` or `x <= GAM**(-2)` (i.e., `GAMSQ` and `RGAMSQ` are removed). I assume the variables `GAMSQ` and `RGAMSQ` were introduced to avoid deficiencies in compiler technologies 45 years ago. Moreover, a `PARAMETER` statement should be used because a missing variable initialization will cause compiler errors:
```fortran
REAL, PARAMETER :: GAM = 2.0E12
DOUBLE PRECISION, PARAMETER :: GAM = 2.0D12
```

christoph-conrads:
The longer that I look at this function, the longer I think it should be left as is: the change would not significantly improve performance, there are no callers of this function in LAPACK, and in general nobody seems to be using it.

The constant 4096 exists to avoid under- and overflows and the same value is used for single and double precision. Picking a larger value would avoid unnecessary re-scaling operations. Another optimization opportunity is the possibly dead [code branch starting in line 153](https://github.com/Reference-LAPACK/lapack/blob/7295ac130f1d880e10c4412b6b524498f0538635/BLAS/SRC/drotmg.f#L153-L164) (there is a comment above the branch saying that the authors do not expect the branch to be taken) which could be removed.

Finally, there exist clones of this implementation, e.g., a C clone created by f2c in [BLIS](https://github.com/flame/blis/blob/a316d2c6c33fc1f8f7c58c4210ab203f48349041/frame/compat/f2c/bla_rotmg.c) or in [golang](https://github.com/ziutek/blas/blob/master/drotmg.go)). The commit comment for the golang code is 13 years old and reads `Drotmg added (not tested!)`.

angsch:
See also https://github.com/Reference-LAPACK/lapack/issues/244, where a paper with a clearer implementation is referenced and a bug related to scaling is mentioned

langou:
I looked into this this morning. Here is a brain dump. I now see that @angsch added a comment. Thanks. I am copy-pasting my brain dump. I'll try to work more on this later in a few days. If anyone wants to make progress, please do.

- The fact that we initialize RGAMSQ which is 2**(-24) to the approximate value of 5.9604645D-8 in double precision bothers me.
- This seems to be related to the fix in gonum point by @angsch. (Thanks!)
- It could be better to change with
```fortran
DOUBLE PRECISION, PARAMETER :: GAM = 2.0D0**12
DOUBLE PRECISION, PARAMETER :: GAMSQ = 2.0D0**24
DOUBLE PRECISION, PARAMETER :: RGAMSQ = 2.0D0**(-24)
```
- I understand that RGAMSQ and GAMSQ are not used in computation, they are used as threshold but still. I understand that the goal was to have the same code base for SINGLE and DOUBLE, but why not write the exact value ( 5.9604644775390625E-008 ) in both codes and let the compiler figure it out. Oh well.
- Yes, we might want to replace all the DATA statements by PARAMETER statements
- Note that there are also some DATA statements in ROTM
- Note that GAM, GAMSQ, RGAMSQ are declared on line 106 of the current code. Then the DATA statement appears.
- I think it would be better to replace DD1 = DD1*GAM**2 by DD1 = (DD1*GAM)*GAM, but if GAMSQ 
- RGAMSQ is inexact (as the current code is done), but GAMSQ is a floating-point number (so is exact). At least this is true in SINGLE (32-bit) and DOUBLE (64-bit). This would even be true in 16-bit floating-point arithmetic, GAMSQ should be a floating-point number. Using GAMSQ with an 8-bit floating-point arithmetic code is not going to cut it though!
- I think that the statement "the values of GAMSQ and RGAMSQ set in the data statement may be inexact" might refer to non-IEEE arithmetics that were around in the seventies. 

I'll have to read more the gonum thread.

langou:
Related. I was reading the code and the WHILE loop looked suspicious. I think you can get an infinite loop if you call the function with DD1 is +Infinity for example. See code below. I did not spend much time on it. Just looked suspicious and the code below seems to have an infinite loop.

```fortran
      program test
      use iso_fortran_env, only: sp => real32, dp => real64
      IMPLICIT NONE
      DOUBLE PRECISION DPARAM(5)
      DOUBLE PRECISION DD1, DD2, DX1, DY1
      real(dp) :: pinf_dp = real(z'7ff0000000000000',dp)

      DD1 = pinf_dp
      DD2 = 1.0D+00
      DX1 = 1.0D+00
      DY1 = 1.0D+00
      CALL DROTMG(DD1,DD2,DX1,DY1,DPARAM)
      print *,DD1
```
      end program

Issue #973, opened 2024-01-10
fails of tests for QR and LQ decompositions


dklyuchinskiy:
Hello!

I have been running tests for QR and LQ  (`xlintstd_dtest`) and sometimes I have got failures like
```
M=   10, N=   10, K=    5, NB=   1, NX=    1, type  2, test( 1)= 0.16408E+16
M=   10, N=   10, K=    5, NB=   1, NX=    1, type  2, test( 2)= 0.62678E+15
```
Usually, I need to change input file to reproduce error.

My investigation shows, that root-cause is PR [900](https://github.com/Reference-LAPACK/lapack/pull/900) with replacement `GELQS` by `GELS` (in example of LQ). Behavior of the test is changed a bit  (see `TESTING/LIN/dchklq.f`):

**Before:**
`QLT01` is called to factorize matrix by `GELQF` for **every** combination of parameters `m` and `n` and this factorized matrix is used at any check of test parameter `k`.

**Now:**
`QLT01` is again called to factorize matrix, but later array `AF` is cleared to `A` to call `GELS` since factorization is now performed **inside** call of `GELS`. But, unfortunately, if `m = n` in call of `GELS` then factorization is done by `GEQRF` instead `GELQF` (check `SRC/dgels.f`). And this is a main difference. For upper triangular matrix QR factorization doing nothing, which lead to failure of `test(1)` and `test(2)` on the **next** iteration of parameter `k`, since arrays `A` and `AF` are equal.

@angsch please take a look.

Input is attached [input qr and lq.txt](https://github.com/Reference-LAPACK/lapack/files/13884747/input.qr.and.lq.txt)

dklyuchinskiy:
Stable reproduced, if there is only one parameter NB in input file. That's why there are no fails for general input file.

It seems, root-cause is also clear. Test have 2 cycles 
```
DO 40 IK = 1, NK
DO 30 INB = 1, NNB
```
If `NNB >= 2`, then factorized matrix becomes correct again, since factorization is performed by `QLT01` (or `GELQF`) due to `IK = 1`. But if `NNB = 1`, then factorization is done by `GELS`, which calls `GEQRF` for `m = n` and test fails for next iterations of `IK`.


Actually, I still have no idea for proper solution to fix this behavior. Construction of the test is complicated. I believe, that we should somehow avoid call of QR in test for LQ.

langou:
[...]

PR #1005, opened 2024-04-08
fix fails of tests for QR/LQ factorizations

dklyuchinskiy:
**Fix fails of tests for QR/LQ factorizations**

Hello! This PR is willing to close issue #973.

Briefly speaking, there are some problems, recently appeared after Release 3.12. Functions `GELQS/GEQRS` were made DEPRECATED and have been replaced by `GELS`  in tests of QR/LQ factorizations. But,

1. `GELS` calls specific factorization depending on input size: `QR` for `m >= n` and `LQ` for `m < n`. Test does not take into account this knowledge and wrong factorization can be called for specific parameters `m` and `n`.
2. Even if parameters, passed to `GELS` are correct, test can be also failed for matrices near overflow/underflow. I believe, that it happens, since `GELS` calls scaling routines inside, but previous version `GELQS/GEQRS` does not. Behavior is also changed in that way.

It is important to point out, that test is failed on the **next** iteration of parameter `K` while data is improperly changed on the **first** iteration. That's why fails are reproduced only on fixed input file.

Thinking of solution, I suggest to call the correct factorization on each iteration of inner loop for parameter `K`. It will solve any problems, related to occasionally changed data inside the test. Maybe this solution is not fully optimal, since it does not restrict any inputs into `GELS`, but it can be discussed.

### DIFF
--- a/TESTING/LIN/cchklq.f
+++ b/TESTING/LIN/cchklq.f
@@ -237,7 +237,7 @@
 *     .. External Subroutines ..
       EXTERNAL           ALAERH, ALAHD, ALASUM, CERRLQ, CGELS, CGET02,
      $                   CLACPY, CLARHS, CLATB4, CLATMS, CLQT01, CLQT02,
-     $                   CLQT03, XLAENV
+     $                   CLQT03, CGELQF, XLAENV
 *     ..
 *     .. Intrinsic Functions ..
       INTRINSIC          MAX, MIN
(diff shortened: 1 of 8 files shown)


dklyuchinskiy:
Hi, @langou !

Could I ask you or somebody to reproduce the issue and check the proposed fix?

In our project we use reduced inputs for testing, different to standard one, that's why it is stable reproduced on our side, and is not failed in LAPACK CI. Standard input file hides the fails and problem inside test.

I have attached logs of fails before and after the fix:

```
The following parameter values will be used:
    M   :       0     1     2     3     5    10    50
    N   :       0     1     2     3     5    10    50
    NRHS:       1     2    15
    NB  :       1
    NX  :       1
    RANK:      30    50    90

DLQ routines passed the tests of the error exits
DLQ:    120 out of   5936 tests failed to pass the threshold

DQR routines passed the tests of the error exits
DQR:    221 out of   8568 tests failed to pass the threshold
```
vs
```
DLQ routines passed the tests of the error exits
All tests for DLQ routines passed the threshold (   5936 tests run)

DQR routines passed the tests of the error exits
All tests for DQR routines passed the threshold (   8568 tests run)
```

Dmitry

dklyuchinskiy:
UPD. Similar fix was applied in https://github.com/Reference-LAPACK/lapack/pull/1298, but in different place of the source file. Maybe it is enough... did not check.

BTW, only LQ tests functions were covered. QR should still have a problem...

Issue #1011, opened 2024-04-18
xORMQR and xLARFB discrepancy with documentation


EduardFedorenkov:
Let's suppose that we have a task to apply vectors from A (returned by xGEQRF) to the set of matrices by xORMQR in parallel. I found that current LAPACK has discrepancy with documentation that introduce doubts when you try to solve this task. 

1. xORMQR : According to documentation the unblocked code branch - xORM2R modify diagonal of A matrix and restore it on the exit. But xORMQR has no such remark about A matrix. So, in this case we can't use xORMQR in parallel because of rewriting the diagonal of A.
2. xLARFB : According to documentation vectors in V are modified and restore on the exit. But according to the code GEMM and TRMM with diag='Unit' don't touch the diagonal elements of V.
![image](https://github.com/Reference-LAPACK/lapack/assets/79554504/bbb60c2c-3939-451f-8a7d-74fadbc58b57)

Maybe it's worth doing some corrections in LAPACK documentations for xLARFB and xORMQR?

langou:
For LARFB, you are correct. We could change: "The elements equal to 1 are not stored; the corresponding array elements are modified but restored on exit. The rest of the array is not used." to "The upper triangular part of V (including its diagonal) is not referenced." Thanks. Feel welcome to provide a PR for LARFB.

For ORMQR, you are correct too. If we use the "unblocked" version of the code, we use ORM2R and then "the diagonal elements of A are modified but restored on exit". And then after line 96 in ORMQR write "A is modified by the routine but restored on exit."

Also, in ORM2R and ORMQR, A could be labelled as "input/output" and not "input". Well, I am not sure about this. A is the same in input and in output, but it is changed and that would help for thread safety.

However we should really "fix" ORM2R (by fixing LARF) so that LARF does not require the "1" on the diagonal of V so that ORM2R does not "modify and restore on exit" entries of V. Maybe we create a new routine for this, such as LARF1. Or just change LARF proper.

Another fix for ORMQR would be to remove the call to ORM2R and always call the "blocked code". So remove 
```fortran
      IF( NB.LT.NBMIN .OR. NB.GE.K ) THEN
*
*        Use unblocked code
*
         CALL DORM2R( SIDE, TRANS, M, N, K, A, LDA, TAU, C, LDC, WORK,
     $                IINFO )
      ELSE
*
*        Use blocked code
*
***** [ KEEP THIS PART ] *****
      END IF
```
[...]

PR #1019, opened 2024-05-29
develop DLARF1F and implement in ORM2R, #1011

EduardFedorenkov:
Hi, @langou!

I develop the first version of DLARF1F.  Some **netlib** tests for QR are failed. Looks like errors in some corner cases.
Please take a look. I will send some description about idea of the algorithm later.


langou:
Hi @EduardFedorenkov, That's great. Thanks for the work

I think that @jprhyne already started a similar work, so I think it would be good to get in synch. @jprhyne: can you PR your work so that we compare the two PRs?

jprhyne:
Yes, I just made one. 

I have similar issues, and currently am looking at making a small test suite to more easily identify the errors.

EduardFedorenkov:
Looks like I fix the bug. Thanks to @jprhyne PR #1020. @jprhyne handled the case **LASTV = 1** correctly, unlike me!

jprhyne:
Awesome! I'll look at my implementation later today and hopefully fix the issue as well. I'll look at the difference between what we did and hopefully it'll be more obvious then!

@EduardFedorenkov since you wanted to work together, do you want to go ahead and do d/z and I do c/s separately or did you want to make a joint repo and we combine the PRs?

EduardFedorenkov:
Yes, I will take a look at the difference between my and your implementation. Then I think we need to align our versions. I think it is necessary to have almost the same code in {Z,D} and {C,S}. After this it was easy to finish the issue. If you already have some suggestions about the algorithm, please write it in this PR.

jprhyne:
Hello Eduard,

I was thinking that since we are only adding the assumption that v(1) = 1 [or v(lastv)=1], that the general structure could stay similar. I do like the blocking idea on C, but I'm not sure I follow why we need to break it up in 4 sections.

Instead, basically do something like C = [ C1 C2 ] since that reflects that we are breaking up v as 

v = [1 x ... x ]^T  (potentially trailing zeros)

Maybe I'm misunderstanding something that is explained by the 4 block style.

jprhyne:
Hey @EduardFedorenkov, I've merged differences aside from the ones I asked about above, I was wondering if this was an acceptable base for us to start with and then move on to the other precisions or if there were other comments or another preference.

Also, I do not mean to post many comments, just wanted to separate the thoughts into two separate blocks!

EduardFedorenkov:
Hi, Johnathan!

You did the great job! I think now your implementation is totally correct and better than mine!
I checked your realization on my bunch of tests, and it also passed them.

I have some comments about "clean code", but it is OK for now. Let's proceed with your realization.
I align versions with you soon. You did D-precision, so you can proceed with Z.

If you share my opinion, then I would remove my implementation of D and do {C, S} (like you did DLARF1F and DLARF1L).

SRC/clarf1l.f
```diff
@@ -0,0 +1,265 @@
...
+*              matrix assuming v(lastv) = 1, where lastv is the last non-zero
+*
+*  =========== DOCUMENTATION ===========
+*
+* Online html documentation available at
+*            http://www.netlib.org/lapack/explore-html/
+*
+*> \htmlonly
+*> Download CLARF1L + dependencies
+*> <a href="http://www.netlib.org/cgi-bin/netlibfiles.tgz?format=tgz&filename=/lapack/lapack_routine/clarf.f">
```
langou, line 11: Hi @EduardFedorenkov and @jprhyne, Can you change this link to clarf1l.f instead of clarf.f? The link won't work, but, next time, we release the link TGZ, ZIP and TXT will be created and the link will work. So we need to have the correct URLs there. And then ditto for each new files. So 1L x 1F x S x C for @EduardFedorenkov and 1L x 1F x D x Z for @jprhyne. Cheers, @langou

EduardFedorenkov: done for {S, C}

langou (approved):

PR #1033, opened 2024-07-12
Prevent compilers from using FMA instructions

### LINKED ISSUE #1031: Less accuracy due to FMAs
Hi,

I think the topic of FMA instructions came up a couple of times before, but my question is what kind of accuracy loss do we consider as acceptable given that this is a reference implementation. The point where I stumbled across it was an eigenvalue decomposition of a 2x2 matrix
```
   A = [  4 1 ]
       [ -4 0 ]
```

The eigenvalues should be `[2.0, 2.0]`. When the compiler uses FMAs we get `[1.999999988777289 2.000000011222711]` instead. Using gfortran on Linux one has to enable native intrinsic (`-march=native`) to accomplish that, but on arm64 we get the slightly wrong results even with default compiler flags. To get the correct values one has to use `-ffp-contract=off` which disables FMAs.

Should we maybe give users a build option which disables FMAs?

For everyone interested here is a quick reproducer of the example above 
```C
#include "lapacke.h"
#include <stdio.h>

int main() {

  int matrix_layout = LAPACK_COL_MAJOR;
  char balanc = 'P';
  char jobvl = 'N';
  char jobvr = 'N';
  char sense = 'N';
  lapack_int n = 2;
  lapack_int lda = 2;
  lapack_int ldvl = 2;
  lapack_int ldvr = 2;

  double A[n*lda];
  A[0] = 4;
  A[1] = -4;
  A[2] = 1;
  A[3] = 0;

  double WR[n];
  double WI[n];
  double VL[n*ldvl];
  double VR[n*ldvr];

  lapack_int ilo[1];
  lapack_int ihi[1];
  double scale[n];
  double abnrm[1];
  double rconde[1];
  double rcondv[1];

  LAPACKE_dgeevx(matrix_layout, balanc, jobvl, jobvr, sense,
                 n, A, lda, WR, WI,
                 VL, ldvl, VR, ldvr,
                 ilo, ihi, scale, abnrm, rconde, rcondv);

  printf("%1.15f %1.15f", WR[0], WR[1]);

  return 0;
}
```

The problematic line that causes the accuracy loss is line 253 in dlanv2.f:

```Fortran
B = BB*CS + DD*SN
```

This expression should evaluate to zero but instead evaluates to `-2.5189846806723163E-017` due to different rounding of the FMA instructions. This value is later compare to zero and used inside a `SQRT` which increases the inaccuracy to ~ `10^-9`:

```Fortran
  IF( C.NE.ZERO ) THEN
     IF( B.NE.ZERO ) THEN
        IF( SIGN( ONE, B ).EQ.SIGN( ONE, C ) ) THEN
*
*                    Real eigenvalues: reduce to upper triangular form
*
           SAB = SQRT( ABS( B ) )
           SAC = SQRT( ABS( C ) )
           P = SIGN( SAB*SAC, C )
           TAU = ONE / SQRT( ABS( B+C ) )
           A = TEMP + P
           D = TEMP - P
           .
           .
           .
```

Maybe an alternative approach would be to compare against machine accuracy instead of zero but that's not my field of expertise.

ACSimon33:
Fixes #1031

**Description**

As described in #1031 when the compiler uses FMA instructions, errors in `?GEEVX` can increase up to `SQRT(EPS)`. The different rounding of the FMA instructions causes a change in control flow, which produces the errors. This MR uses explicit parentheses to prevent the compiler from using FMA instructions for those results, which are compared against zero in subsequent if-statements. 

**Checklist**

### DIFF
--- a/SRC/dlanv2.f
+++ b/SRC/dlanv2.f
@@ -144,7 +144,7 @@
 *     ..
 *     .. Local Scalars ..
       DOUBLE PRECISION   AA, BB, BCMAX, BCMIS, CC, CS1, DD, EPS, P, SAB,
-     $                   SAC, SCALE, SIGMA, SN1, TAU, TEMP, Z, SAFMIN, 
+     $                   SAC, SCALE, SIGMA, SN1, TAU, TEMP, Z, SAFMIN,
      $                   SAFMN2, SAFMX2
       INTEGER            COUNT
 *     ..
--- a/SRC/dlanv2.f
+++ b/SRC/dlanv2.f
@@ -249,9 +249,13 @@
 *           Compute [ A  B ] = [ CS  SN ] [ AA  BB ]
 *                   [ C  D ]   [-SN  CS ] [ CC  DD ]
 *
+*           Note: Some of the multiplications are wrapped in parentheses to
+*                 prevent compilers from using FMA instructions. See
+*                 https://github.com/Reference-LAPACK/lapack/issues/1031.
+*
             A = AA*CS + CC*SN
-            B = BB*CS + DD*SN
-            C = -AA*SN + CC*CS
+            B = ( BB*CS ) + ( DD*SN )
+            C = -( AA*SN ) + ( CC*CS )
             D = -BB*SN + DD*CS
 *
             TEMP = HALF*( A+D )
--- a/SRC/slanv2.f
+++ b/SRC/slanv2.f
@@ -144,7 +144,7 @@
 *     ..
 *     .. Local Scalars ..
       REAL               AA, BB, BCMAX, BCMIS, CC, CS1, DD, EPS, P, SAB,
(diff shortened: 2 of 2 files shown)


SRC/dlanv2.f
```diff
@@ -248,10 +248,14 @@ SUBROUTINE DLANV2( A, B, C, D, RT1R, RT1I, RT2R, RT2I, CS, SN )
 *
 *           Compute [ A  B ] = [ CS  SN ] [ AA  BB ]
 *                   [ C  D ]   [-SN  CS ] [ CC  DD ]
+*
+*           Note: Some of the multiplications are wrapped in parentheses to
```
langou, line 252: The comments are great! Good idea to refer to the GitHub issue!

langou (approved):

PR #1036, opened 2024-07-19
Two fixes of STEVX

haox1:
**Description**
This PR is for two fixes of d/sSTEVX:

1. Set IFAIL when N = 1 and WANTZ = .TRUE. 
2. Scale ABSTOL when matrix being scaled to allowable range.

**Checklist**


haox1:
Clos due to a mistake.

PR #1039, opened 2024-07-19
Fixes of STEBZ

haox1:
**Description**

This PR fixes d/sSTEBZ:

1. Fix accumulation of number of eigenvalues by changing .LT. to .LE. in line 600 (599 for sstebz.f) as line 538 (537 for sstebz.f) states that “NWL accumulates the number of eigenvalues .le. WL” .
2. Fix possible out of bounds write of IBLOCK. There are IBLOCK(IW) = 0 in lines 704 and 719 (703 and 718 for sstebz.f), which IW are initialized to 0. IF statements are added to check whether IW equals to zero.
3. Fix failures of test case 53 of SSBEVX in sdrvst.f on Power8 and Power9 by increasing the FUDGE factor in sstebz.f to 2.3.

**Checklist**

### DIFF
--- a/SRC/sstebz.f
+++ b/SRC/sstebz.f
@@ -293,7 +293,7 @@
       PARAMETER          ( ZERO = 0.0E0, ONE = 1.0E0, TWO = 2.0E0,
      $                   HALF = 1.0E0 / TWO )
       REAL               FUDGE, RELFAC
-      PARAMETER          ( FUDGE = 2.1E0, RELFAC = 2.0E0 )
+      PARAMETER          ( FUDGE = 2.3E0, RELFAC = 2.0E0 )
 *     ..
 *     .. Local Scalars ..
       LOGICAL            NCNVRG, TOOFEW
@@ -596,7 +596,7 @@
             END IF
 *
             IF( IRANGE.GT.1 ) THEN
-               IF( GU.LT.WL ) THEN
+               IF( GU.LE.WL ) THEN
                   NWL = NWL + IN
                   NWU = NWU + IN
                   GO TO 70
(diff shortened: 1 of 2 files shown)


ilayn:
cc: @langou @oamarques

ilayn:
Some background on the first point 

 1. There is already some contractual hints in the comments 537-539
     ```
    *
    *     Find Eigenvalues -- Loop Over Blocks and recompute NWL and NWU.
    *     NWL accumulates the number of eigenvalues .le. WL,
    *     NWU accumulates the number of eigenvalues .le. WU
    *
    ```
 2. The test value `GU` is already fudged `GU = GU + FUDGE*BNORM*ULP*IN + FUDGE*PIVMIN` on line 587 hence, it is my understanding that, if fudged `GU` equals `WL`, every true eigenvalue of the block is strictly less than `WL` and the boundary is included in the spectrum.

langou (approved):

SRC/sstebz.f
```diff
@@ -293,7 +293,7 @@ SUBROUTINE SSTEBZ( RANGE, ORDER, N, VL, VU, IL, IU, ABSTOL, D,
       PARAMETER          ( ZERO = 0.0E0, ONE = 1.0E0, TWO = 2.0E0,
      $                   HALF = 1.0E0 / TWO )
       REAL               FUDGE, RELFAC
-      PARAMETER          ( FUDGE = 2.1E0, RELFAC = 2.0E0 )
+      PARAMETER          ( FUDGE = 2.3E0, RELFAC = 2.0E0 )
```
langou, line 296: I am slightly uncomfortable that this proposed change will imply that the `FUDGE` would not be the same for `SINGLE` and `DOUBLE`.

langou:
I am planning in merging this PR. Comments welcome before this happens.

ilayn:
Agree with the fudge factor comment. It is a test issue and should not be effecting the algorithms. The tests can be modified later independently should the fails continue.

PR #1002, opened 2024-03-28
Allow User Configuration of RPATH Settings

ahnaf-tahmid-chowdhury:
## Description

This pull request introduces user-configurable options for controlling RPATH settings in the CMake build system. Previously, these settings were forced in the CMakeLists.txt file, leading to inflexibility and potential conflicts, as reported in issue #1001. By making these options configurable, users have more control over the behavior of RPATH settings during the build process.

## Changes Made

- Introduce options for enabling macOS RPATH, skipping build-time RPATH, and building with install RPATH.
- Modify the logic to use user-configured options and provide feedback messages regarding the state of each option.
- Refactor the setting of install RPATH to be conditional upon whether the directory is considered a system directory.

## Checklist

- [ ] The documentation has been updated.
- [x] If the PR solves a specific issue, it is set to be closed on merge.

## Additional Notes

- Please review and provide feedback.
- This PR aims to address the reported issue and enhance the flexibility of RPATH settings in the CMake build system.

PR #1040, opened 2024-08-07
Remove CMake < 3 CMP0042 workarounds

haampie:
**Description**

TL;DR: This deletes obsolete CMake code and improves upon #1002
by sticking to (sensible) CMake defaults, in accordance to the principle
of least astonishment ;)

CMake only sets install names on darwin to `@rpath/<name>` (which is
desired otherwise rpaths don't work at all) when CMP0042 is ON. That's
the default when CMake 3.0 or higher is required. And lapack requires it
already for years: as of v3.9.1 (8f004b353a876a117b9e9428fff3becdd22c79ce).

So, delete the old workarounds that effectively set CMP0042 to ON.

Further, delete the following three options that are redefinitions of
builtins and have values that are default anyways:
- `CMAKE_MACOSX_RPATH`
- `CMAKE_SKIP_BUILD_RPATH`
- `CMAKE_BUILD_WITH_INSTALL_RPATH`

Lastly, lapack automatically sets `CMAKE_INSTALL_RPATH_USE_LINK_PATH` to
`ON` whenever installing to a non-system dir. The assumption is that
whenever you install something to a non-system dir, you need rpaths to
locate dependencies. But this is just an assumption which may or may
not hold. The downside of it is that the option can be annoying when
lapack is used as a sub-project as it affects a global CMake variable
(for example OpenBLAS uses lapack as a sub-project). Instead, let users
or packagers provide this on the command line if they really need it --
remove it from lapack as it's as helpful as it is harmful.

**Checklist**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,30 +123,6 @@ message(STATUS "Build Index-64 API as extended API with _64 suffix: ${BUILD_INDE
 
 include(GNUInstallDirs)
 
-# Updated OSX RPATH settings
-# In response to CMake 3.0 generating warnings regarding policy CMP0042,
-# the OSX RPATH settings have been updated per recommendations found
-# in the CMake Wiki:
-#  http://www.cmake.org/Wiki/CMake_RPATH_handling#Mac_OS_X_and_the_RPATH
-option(CMAKE_MACOSX_RPATH "Enable macOS RPATH" ON)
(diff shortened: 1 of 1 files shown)


haampie:
ping @langou / @ahnaf-tahmid-chowdhury (ref #1002)

ahnaf-tahmid-chowdhury (approved):
I agree. Leveraging CMake's built-in capabilities whenever possible is generally a good approach.

langou (approved):

PR #1044, opened 2024-08-20
Declare project dependency on C to avoid the CMAKE default of C,C++

martin-frbg:
**Description**
#834 had removed all language requirements from the `project` command in the toplevel CMakeLists.txt in order to move the dependencies on a Fortran compiler to those subprojects that actually need them (and make LAPACKE buildable without one). 
However, the default in CMAKE is to assume a project dependency on both C and C++ if no language was
expressly requested. In the absence of any actual C++ code, this creates a spurious dependency that can prevent building
LAPACK on systems where such a compiler is not available or at least not installed by default. (This was noted in a comment on #988)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.13)
 
-project(LAPACK)
+project(LAPACK C)
 
 set(LAPACK_MAJOR_VERSION 3)
 set(LAPACK_MINOR_VERSION 12)


langou (approved):

martin-frbg:
Indeed the "macos-latest" image has been updated from 14.6 to 14.6.1 recently, and following the embedded link in the CI log to the list of installed software in that runner image leads to a message stating that gcc-11 was to be removed from all images on August 12 https://github.com/actions/runner-images/issues/10213

PR #1045, opened 2024-08-20
CI: Update the MacOS jobs to use gcc-14 in github workflows

martin-frbg:
**Description**
gcc-11 was removed from the runner images provided by github on August 12, 2024. As gcc-11 was the most recent release available at the time the workflow files were created, I have updated them to gcc-14 directly, although gcc-12 and gcc-13 would also be available for the time being.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -75,12 +75,12 @@ jobs:
     - name: Install ninja-build tool
       uses: seanmiddleditch/gha-setup-ninja@16b940825621068d98711680b6c3ff92201f8fc0 # v3
 
-    - name: Use GCC-11 on MacOS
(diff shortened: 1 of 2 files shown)


martin-frbg:
Now the CI failure is caused by gcc-14 picking up missing API_SUFFIX wrapper macros in the LAPACKE  ?TFSM interfaces (which were touched by #1042 that got merged two weeks ago - looks like this is something to look out for in forthcoming contributions from people who based their commits on anything earlier than 3.12.0 and thus lack all the wrapper code for Kraynyuk's INDEX-64 API)

martin-frbg:
created #753 for the LAPACKE errors found here

weslleyspereira (approved):

PR #1057, opened 2024-10-01
Add forgotten API_SUFFIX wrappers to LAPACKE_?TFSM and fix the MacOS CI jobs

martin-frbg:
**Description**
The missing API_SUFFIX wrappers appear to be an oversight from the recent #1042
This version of the PR combines my previous #1046 with my unrelated PR #1045 to
fix the CI jobs that got broken by the removal of gcc-11 from the underlying image
(as requested by @weslleyspereira)

PR #1046, opened 2024-08-20
Add a few forgotten API_SUFFIX wrappers in the recently changed LAPACKE ?TFSM

martin-frbg:
**Description**
This appears to have been an oversight introduced with the fixes from #1042

PR #1059, opened 2024-10-02
Revert "Revert "Add a few forgotten API_SUFFIX wrappers in the recently changed LAPACKE ?TFSM""

weslleyspereira:
Reverts Reference-LAPACK/lapack#1058.

That got complicated as @martin-frbg predicted in https://github.com/Reference-LAPACK/lapack/pull/1057#issuecomment-2386273263. Sorry for that.

### DIFF
--- a/LAPACKE/src/lapacke_ctfsm.c
+++ b/LAPACKE/src/lapacke_ctfsm.c
@@ -45,7 +45,7 @@ lapack_int API_SUFFIX(LAPACKE_ctfsm)( int matrix_layout, char transr, char side,
 #ifndef LAPACK_DISABLE_NAN_CHECK
     if( LAPACKE_get_nancheck() ) {
         lapack_int mn = m;
-        if( LAPACKE_lsame( side, 'r' ) ) mn = n;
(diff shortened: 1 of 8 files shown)


martin-frbg:
Actually I'm a bit unsure what the problem is now, since there were no intervening commits ? My suggestion was simply that merging the second of my original PRs after inadvertently merging the first should be equivalent to, and less noisy in the git history than reverting one and then merging the PR that combined them. Now we're looking at revert&revert-the-revert which is bound to be even more noisy ?

langou (approved):

weslleyspereira:
> Actually I'm a bit unsure what the problem is now, since there were no intervening commits ? My suggestion was simply that merging the second of my original PRs after inadvertently merging the first should be equivalent to, and less noisy in the git history than reverting one and then merging the PR that combined them. Now we're looking at revert&revert-the-revert which is bound to be even more noisy ?

Github didn't let us merge your PR #1057 after we accepted #1046. I suspect it is because they shared same commit hashes, if that is really the case. Reverting the merge didn't help.

PR #1062, opened 2024-10-12
Remove comparison as these conditions are never true

schnorr:
**Description**
The problem is that dtpmqrt is impossible to use because `side` must be either `l` (`L`) or `r` (`R`). The condition not only uses `LAPACKE_lsame` to compare against the possible values, but also compare the boolean result against `side` (which is a `char`). This situation ultimately leads to a logical expression that will never be true, making it impossible to use `dtpmqrt` and others routines affected by similar code. The commit simply removes `side ==` as this fix the problem.

**Checklist**

  - No documentation change required as in previous version it was OK
  - No issues found describing this problem


martin-frbg:
looks like these comparisons came in with #540 @weslleyspereira

langou (approved):

PR #1064, opened 2024-10-14
fix leading dimension for matrix B in test routine get52

dklyuchinskiy:
**fix leading dimension for matrix B in test routine get52**

Led to fails of tests for eigensolver `ggev` for cases `lda != ldb`


langou (approved):

langou:
Good catch! Thanks!

PR #1069, opened 2024-10-25
Use GEMMTR for SY/HE linear updates

angsch:
Replace the linear updates that only update the lower or upper triangular part with calls to GEMMTR.


langou (approved):
This is beautiful. Thanks a lot for the contribution.

PR #1070, opened 2024-11-05
slassq.f90 need la_xisnan.mod

DragonPara:
**Description**
maybe a bug
image

**Checklist**


martin-frbg:
I believe this should not be necessary as the general rules for  f90 files right below your changes already define a dependency on "all" mod files. Did you encounter an actual problem compiling LAPACK with make ?

DragonPara:
When I used `make -j4 ` to generate liblapack.a, I encountered the problem.

Perhaps it was a sporadic error

martin-frbg:
I could not reproduce this problem on my system (Linux with GNU make version 4.4.1).

PR #1071, opened 2024-11-07
[Github workflow] Bump scorecard version

angsch:
For some time PR merges have listed a failure due to an unsuccessful update of the scorecard.

<details>
  <summary>Details</summary>

image

```
error signing scorecard results: getting key from Fulcio: getting CTFE public   keys: updating local metadata and targets: error updating to TUF remote mirror: invalid key
```
</details>

Let's hope that a version update fixes the problem...

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -32,12 +32,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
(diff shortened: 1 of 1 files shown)


langou:
This has been bothering me too (and I am not sure others) and I was not sure how to fix it. Let's hope that this fix works. Thanks a lot. Let's see. Julien

angsch:
Looks like the fix worked. However, the number of checks went down from the usual 13 to 2. I hope that this because the source code is unchanged and not a new issue... .

PR #1079, opened 2024-11-28
Fix line lengths for the extended API

hjjvandam:
**Description**

The Spack software stack management system builds LAPACK with the extended API. Using Spack to build the development version of LAPACK causes problems when the subroutine names are extended by appending "_64". In a number of cases the 3 extra characters push the line length over 72 characters, causing the compiler to truncate the code, and generate errors. 

Most of the instances I have hit are in deprecated subroutines. Nevertheless, Spack insists on building deprecated subroutines as many software packages don't actively track what code has been deprecated. Therefore leaving those subroutines out would cause many builds to fail. 

I have gone through every subroutine that failed to compile with the extended subroutine names and fixed the line lengths to ensure that even with the extended subroutine names the line lengths remain below 72 characters.

Just to be sure I have also run the test cases (source code truncation can lead to syntactically valid but semantically different Fortran). The output of the tests is
```
Test project /Users/hubertusvandam/Documents/spack.nosync/junk/lapack/build
        Start   1: BLAS-xblat1s
  1/107 Test   #1: BLAS-xblat1s .....................   Passed    0.03 sec
        Start   2: BLAS-xblat2s
  2/107 Test   #2: BLAS-xblat2s .....................   Passed    0.03 sec
        Start   3: BLAS-xblat3s
  3/107 Test   #3: BLAS-xblat3s .....................   Passed    0.03 sec
        Start   4: BLAS-xblat1d
  4/107 Test   #4: BLAS-xblat1d .....................   Passed    0.02 sec
        Start   5: BLAS-xblat2d
  5/107 Test   #5: BLAS-xblat2d .....................   Passed    0.02 sec
        Start   6: BLAS-xblat3d
  6/107 Test   #6: BLAS-xblat3d .....................   Passed    0.03 sec
        Start   7: BLAS-xblat1c
  7/107 Test   #7: BLAS-xblat1c .....................   Passed    0.02 sec
        Start   8: BLAS-xblat2c
  8/107 Test   #8: BLAS-xblat2c .....................   Passed    0.02 sec
        Start   9: BLAS-xblat3c
  9/107 Test   #9: BLAS-xblat3c .....................   Passed    0.03 sec
        Start  10: BLAS-xblat1z
 10/107 Test  #10: BLAS-xblat1z .....................   Passed    0.02 sec
        Start  11: BLAS-xblat2z
 11/107 Test  #11: BLAS-xblat2z .....................   Passed    0.02 sec
        Start  12: BLAS-xblat3z
[...]

PR #1080, opened 2024-11-30
Adding a recursive xLARFT

jprhyne:
Hey Y'all,

I've been working on a recursive implementation of the xLARFT routines. I have some figures below that show some increased performance for the DLARFT routine on its own as well as some performance graphs for when we use the DLARFT subroutine within DGEQRF.

One can argue that instead of using DGEQR2+DLARFT in DGEQRF one should use DGEQRT3 which is true, but DLARFT is useful in many other settings where a routine akin to DGEQRT3 does not exist.

Long story short: when we increase the block size of our routine calls (K in xLARFT), we see some drastic improvement in the performance of DLARFT. The performance improvement for DGEQRF are less impressive since DLARFT (with a reasonably small block size) represents a small portion of DGEQRF.

For the experiments, I used the HPC at my university, and to give context to the actual performance values, I have
included a figure that highlights the performance for square matrix-matrix multiplication.

[flopMMM.pdf](https://github.com/user-attachments/files/17964917/flopMMM.pdf)

[newBlocksizePerf.pdf](https://github.com/user-attachments/files/17965035/newBlocksizePerf.pdf)
Note: I previously attached an incorrect file for the above. Please disregard. Sorry for any confusion!

[varyBlocksizeDlarftPerf.pdf](https://github.com/user-attachments/files/17964924/varyBlocksizeDlarftPerf.pdf)

And if there is an interest in the methodology for how these graphs were generated or anything along
those lines, my development of these routines can be found in [my xor-xungr repository](https://github.com/jprhyne/xor-xungqr/)

The main algorithm for the recursive xLARFT recurses down to the tree node (the case of N=1 or K=1), and
not much investigation was done about if an earlier bailout (to the old method or another one) could
increase performance, so this is something that could be investigated but through some brief toying around on
my machine, I did not see any real difference when using optimized BLAS (AOCL)

Since this method will be changing how xLARFT is being done, I also moved the existing version as a variant titled
'LL-LVL2'.

I appreciate any feedback on this approach or anything else done!
**Checklist**


SRC/dormqr.f
```diff
@@ -309,8 +309,7 @@ SUBROUTINE DORMQR( SIDE, TRANS, M, N, K, A, LDA, TAU, C, LDC,
 *           Form the triangular factor of the block reflector
 *           H = H(i) H(i+1) . . . H(i+ib-1)
 *
-            CALL DLARFT( 'Forward', 'Columnwise', NQ-I+1, IB, A( I,
-     $                   I ),
+            CALL DLARFT( 'Forward', 'Columnwise', NQ-I+1, IB, A( I, I ),
```
langou, line 312: I think that works, the last comma of the line (the one after A(I,I)) seems to be the 72nd character of the line, so I think that changes work. I am more worried than anything about these types of changes. See #1079 where @hjjvandam proposes a fix to get some lines below 72 characters when we add _64 here and there.

SRC/la_xisnan.mod
langou, line 1: I do not think we want the .mod file

SRC/la_constants.mod
langou, line 1: I do not think we want the .mod file

SRC/dormqr.f
langou: I think it is best to leave the PR focused on LARFT. so please do not commit changes on GELQF, GERQF, etc. Only [S,D,,C,Z]LARFT. If you want to do a separate commit to improve some formatting in some routines here and there, sure. But please separate.

jprhyne: Good point, I didn't think about the extensions. So in the failings that I have elsewhere, I'll keep the line length at or below 68 and hopefully that'll resolve those issues.

Thanks for the point to the PR!

SRC/la_xisnan.mod
jprhyne: Is this something that could potentially be added to a gitignore file as I notice the file is created after I run 'make' so it seems to be a product of some compilation process. Not sure if its because of some machine tests that are failing (things like testing NaNs and Max behavior)

SRC/VARIANTS/larft/LL-LVL2/clarft.f
```diff
@@ -0,0 +1,328 @@
...
+*
+*  Authors:
+*  ========
+*
+*> \author Univ. of Tennessee
+*> \author Univ. of California Berkeley
+*> \author Univ. of Colorado Denver
+*> \author NAG Ltd.
+*
+*> \ingroup larft
```
angsch, line 133: If the old version shall live in VARIANTS,  should the group be updated for the documentation? The other files in this folder carry "variant" in the name. Also, could you please add the variants to the build system? `SRC/VARIANTS/Makefile` and `SRC/VARIANTS/README`.

jprhyne: Thank you for the build system catch! It should be in there for the most recent commit I've made. I'm also hoping that I ironed out all the line length related issues (mentioned above)

SRC/la_xisnan.mod
langou: "Is this something that could potentially be added to a gitignore file." Good idea. Feel welcome to submit a separate pull request with this feature.

langou (approved):

PR #1093, opened 2025-01-10
dlahrd.f: consistent line reflow for DTRMV calls

wdconinc:
**Description**
Under gcc-12.3.0, we are noticing spurious type mismatch errors due to different line reflow in dlahrd.f, see e.g. https://gitlab.spack.io/spack/spack/-/jobs/14595649#L1131. Notice that the types of different arguments are compared in that error message.

This PR makes the line reflow for two DTRMV calls identical, which resolves the spurious report.

**Checklist**

- [ ] The documentation has been updated.
  - No documentation updates needed
- [ ] If the PR solves a specific issue, it is set to be closed on merge.
  - No specific issue reported, but see https://github.com/spack/spack/pull/48318#issuecomment-2583610274.

PR #1094, opened 2025-01-10
?ge(lq|qr)s.f: WORK( LWORK ) -> WORK( * )

### LINKED ISSUE #1092: 3.12.1 fails to build
With GCC 14.2
```
/build/lapack/src/build/SRC/lapack_64_obj/cgelqs.f:132:30:

  132 |      $                   WORK( LWORK )
      |                              1
Error: Variable ‘lwork’ cannot appear in the expression at (1)
make[2]: *** [SRC/CMakeFiles/lapack_64_obj.dir/build.make:507: SRC/CMakeFiles/lapack_64_obj.dir/lapack_64_obj/cgelqs.f.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```
Build command:
```
cmake -B build -S lapack \
    -DCMAKE_SKIP_RPATH=ON \
    -DBUILD_SHARED_LIBS=ON \
    -DBUILD_TESTING=OFF \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DCMAKE_Fortran_COMPILER=gfortran \
    -DLAPACKE_WITH_TMG=ON \
    -DCBLAS=ON \
    -DBUILD_DEPRECATED=ON \
    -DBUILD_MAN_DOCUMENTATION=ON
  cmake --build build
```

wdconinc:
**Description**
In #1092, compilation of 3.12.1 is reported to fail due to `WORK( LWORK )` in `SRC/DEPRECATED/?ge(lq|qr)s.f`. I can reproduce this locally (gcc-14.2.0). Suggested in the report is a change to `WORK( * )`.

This PR changs `WORK( LWORK )` to `WORK( * )` in all `SRC/DEPRECATED/?ge(lq|qr)s.f`, which is confirmed to cause compilation to succeed (when combined with #1093).

**Checklist**

- [x] If the PR solves a specific issue, it is set to be closed on merge. Fixes #1092.

### DIFF
--- a/SRC/DEPRECATED/cgelqs.f
+++ b/SRC/DEPRECATED/cgelqs.f
@@ -16,7 +16,7 @@
 *       ..
 *       .. Array Arguments ..
 *       COMPLEX            A( LDA, * ), B( LDB, * ), TAU( * ),
-*      $                   WORK( LWORK )
+*      $                   WORK( * )
 *       ..
 *
 *
@@ -128,7 +128,7 @@
 *     ..
 *     .. Array Arguments ..
       COMPLEX            A( LDA, * ), B( LDB, * ), TAU( * ),
-     $                   WORK( LWORK )
+     $                   WORK( * )
 *     ..
 *
 *  =====================================================================
--- a/SRC/DEPRECATED/cgeqrs.f
+++ b/SRC/DEPRECATED/cgeqrs.f
@@ -16,7 +16,7 @@
 *       ..
 *       .. Array Arguments ..
 *       COMPLEX            A( LDA, * ), B( LDB, * ), TAU( * ),
-*      $                   WORK( LWORK )
+*      $                   WORK( * )
 *       ..
 *
 *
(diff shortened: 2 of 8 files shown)


antonio-rojas:
With this and #1093 it builds, but with broken symbols [1]

[1] https://gitlab.archlinux.org/archlinux/packaging/packages/lapack/-/issues/3

martin-frbg:
interesting that both this and #1093 are only showing up in the _64_ half of the build, makes me wonder if there might be underlying symbol collisions between the 32- and 64-bit integer versions of the functions

langou:
Thanks for the PR @wdconinc. 

On the one hand, 'WORK( LWORK )' indicates to the compiler the size of the array WORK. Doing so, we can hope that, if we do some out-of-bound accesses, we get errors notifications. When we write 'WORK( * )', we forego this possible feature. On the other hand, most of LAPACK uses 'WORK( * )' . The 'WORK( LWORK )' is scarcely used.

langou (approved):

wdconinc:
I'm wondering if this PR is held up by something, or whether it would be possible to merge it.

PR #1098, opened 2025-01-19
Add NaN propagation and array initialization notes to ?GEMM docs

### LINKED ISSUE #1077: ?GEMM documentation fuzzy on NaN handling when alpha or beta is zero
### Preamble
The descriptions for ?GEMM start as:

> DGEMM  performs one of the matrix-matrix operations
>    C := alpha*op( A ) \* op( B ) + beta\*C

Assuming this is computed using IEEE rules, if `alpha` is zero and `A` or `B` contain a `NaN` or `+/-Inf`, then the result should contain `NaN`s, as `0.0 * Inf == NaN` and `0.0 * NaN== NaN`. The same goes if `beta` is zero and `C` contains a `NaN` or `+/-Inf`.

### Issue 1
As far as I can tell from a cursory look, this is not how ?GEMM actually behaves in the Netlib implementation.

1. Due to the early return at line 271 of DGEMM, if `alpha` is zero and `beta` is one, `NaN`s or `Inf`s in `A` or `B` do not affect the result.
2. If both `alpha` and `beta` are zero, then a zero matrix is returned (line 279), irrespective of any `NaN`s or `Inf`s in `A`, `B` or `C`.
3. If `alpha` is zero and `beta` is neither zero or one, then `beta*C` is returned, irrespective of any `NaN`s or `Inf`s in `A` or `B`.
4. If only `beta` is zero, `alpha*op( A )*op( B )` is returned, irrespective of any `NaN`s or `Inf`s in `C`.

Therefore there are inconsistencies between the behavior expected from reading the docs and actual program behavior. These are arguably corner cases, but `NaN` propagation is an important mechanism, and callers should be made aware if a given subroutine ignores `NaN` inputs under certain circumstances.
One example of this mattering, is when one of the matrices may contain uninitialized data, leading to Issue#2

### Issue 2
The documentation does not make it clear enough when it is OK to call ?GEMM with matrices containing uninitialized data.
For example, the simplest use case for ?GEMM is computing `C = A*B`. The current documentation states that:

> When  BETA  is supplied as zero then C need not be set on input.

and

> Before entry, the leading  m by n  part of the array  C must contain the matrix  C,  except when  beta  is zero, in which case C need not be set on entry.

However depending on a person's interpretation of what it means for `C` to be "set", this may fail to convey that if `beta` is zero, then `C` is permitted to be contain arbitrary bytes (including ones that resolve to `NaN`s or `Inf`s) before entry. As a results users may unnecessarily zero-fill `C` before calling ?GEMM.

The same is also true for the other corner cases where `alpha` is zero, where both `A` and `B` are allowed to be filled with arbitrary bytes before entry without any effect on the output. Since the current docs do not mention any special behavior for `alpha == 0.0`, one would incorrectly assume that `A` and `B` has to be initialized to guarantee correct results, even if `alpha` is zero.

### Suggested fixes

1. Change the docs to point out that ?GEMM does not compute `C := alpha*op( A )*op( B ) + beta*C` as written and as a consequence does not adhere to all of the `NaN` and `Inf` propagation rules that the formula implies.
2. Document the corner cases.
3. Add notes about when the matrices are permitted to contain arbitrary bytes before entry.

**Checklist**
- [ ] I've included a minimal example to reproduce the issue

TiborGY:
**Description**
This PR fixes and closes #1077 and:
1. Adds this description of NaN/Inf propagation quirks to the ?GEMM docs:
```
Note: if alpha and/or beta is zero, some parts of the matrix-matrix operations are not performed.
This results in the following NaN/Inf propagation quirks:

 1. If alpha is zero, NaNs or Infs in A or B do not affect the result.
 2. If both alpha and beta are zero, then a zero matrix is returned in C, irrespective of any NaNs or Infs 
 in A, B or C.
 3. If only beta is zero, alpha*op( A )*op( B ) is returned, irrespective of any NaNs or Infs in C.
```
2. Adds notes like this to S/D GEMM about why they support complex conjugation:
```
Note: TRANSA = 'C' is supported for the sake of API consistency between all ?GEMM variants.
```
3. Adds this to the descriptions of `ALPHA` arguments:
```
If ALPHA is zero the values in A and B do not affect the result.
This also means that NaN/Inf propagation from A and B is inhibited if ALPHA is zero.
```
4. Changes descriptions of `A` arguments from `array A must contain the matrix A.` to:
```
array A must contain the matrix A, except if ALPHA is zero.
If ALPHA is zero, none of the values in A affect the result, even if they are NaN/Inf.
This also implies that if ALPHA is zero, the matrix elements of A need not be initialized by the caller.
```
5. Changes descriptions of `B` arguments from `array B must contain the matrix B.` to:
```
array B must contain the matrix B, except if ALPHA is zero.
If ALPHA is zero, none of the values in B affect the result, even if they are NaN/Inf.
This also implies that if ALPHA is zero, the matrix elements of B need not be initialized by the caller.
```
6. Changes descriptions of `BETA` arguments from
`When BETA is supplied as zero then C need not be set on input.`
to:
```
If BETA is zero the values in C do not affect the result. This also means that NaN/Inf propagation
from C is inhibited if BETA is zero.
```
7. Changes descriptions of `C` arguments from
`array C must contain the matrix C, except when beta is zero, in which case C need not be set on entry.`
to:
```
array  C must contain the matrix  C, except if beta is zero. If beta is zero, none of the values in C 
affect the result, even if they are NaN/Inf. This also implies that if beta is zero, the matrix elements of 
C need not be initialized by the caller.
```
This documents/clarifies existing behaviour without adding anything to the docs that might hypothetically clash with the behaviour of some optimized BLAS implementation. (eg. I have decided against adding no-touch-guarantees to A, B and C, so a conforming implementation might still read data from them, as long as it does not affect the output)

**Checklist**

### DIFF
--- a/BLAS/SRC/cgemm.f
+++ b/BLAS/SRC/cgemm.f
@@ -35,6 +35,19 @@
 *>
 *> alpha and beta are scalars, and A, B and C are matrices, with op( A )
 *> an m by k matrix,  op( B )  a  k by n matrix and  C an m by n matrix.
+*>
+*> Note: if alpha and/or beta is zero, some parts of the matrix-matrix
+*>  operations are not performed. This results in the following NaN/Inf
+*>  propagation quirks:
+*>
+*>  1. If alpha is zero and beta is one, NaNs or Infs in A or B do not affect
+*>   the result.
+*>  2. If both alpha and beta are zero, then a zero matrix is returned in C,
+*>   irrespective of any NaNs or Infs in A, B or C.
+*>  3. If alpha is zero and beta is neither zero or one, then beta*C is
+*>   returned in C, irrespective of any NaNs or Infs in A or B.
+*>  4. If only beta is zero, alpha*op( A )*op( B ) is returned, irrespective
+*>   of any NaNs or Infs in C.
 *> \endverbatim
 *
 *  Arguments:
@@ -92,7 +105,9 @@
 *> \param[in] ALPHA
 *> \verbatim
 *>          ALPHA is COMPLEX
-*>           On entry, ALPHA specifies the scalar alpha.
+*>           On entry, ALPHA specifies the scalar alpha. If ALPHA is zero the
+*>           values in A and B do not affect the result. This also means that
+*>           NaN/Inf propagation from A and B is inhibited if ALPHA is zero.
 *> \endverbatim
 *>
 *> \param[in] A
@@ -102,7 +117,10 @@
 *>           Before entry with  TRANSA = 'N' or 'n',  the leading  m by k
 *>           part of the array  A  must contain the matrix  A,  otherwise
 *>           the leading  k by m  part of the array  A  must contain  the
-*>           matrix A.
+*>           matrix A, except if ALPHA is zero.
+*>           If ALPHA is zero, none of the values in A affect the result, even
+*>           if they are NaN/Inf. This also implies that if ALPHA is zero,
+*>           the matrix elements of A need not be initialized by the caller.
 *> \endverbatim
 *>
 *> \param[in] LDA
@@ -121,7 +139,10 @@
 *>           Before entry with  TRANSB = 'N' or 'n',  the leading  k by n
 *>           part of the array  B  must contain the matrix  B,  otherwise
 *>           the leading  n by k  part of the array  B  must contain  the
-*>           matrix B.
+*>           matrix B, except if ALPHA is zero.
+*>           If ALPHA is zero, none of the values in B affect the result, even
+*>           if they are NaN/Inf. This also implies that if ALPHA is zero,
+*>           the matrix elements of B need not be initialized by the caller.
 *> \endverbatim
 *>
 *> \param[in] LDB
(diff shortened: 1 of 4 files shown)


ilayn:
May I suggest a less if, else language such as 

> If alpha is set to zero, A and B are not referenced to avoid unnecessary math. A potential downside of this is that if the arrays contain NaN/Inf values they are not transferred to C.

This is more to the point and explains the mechanism instead of enumerating rules. I don't claim a command on English but this reads easier to me.

TiborGY:
On a third reading, some of the conditions in the quirks list were redundant, so I have simplified that.

ilayn:
> Your version changes the meaning slightly, because "A and B are not referenced" would imply a no-touch-guarantee (no reads, no writes to any array element), which is something I do not want to include in the BLAS specification.

That's a standard LAPACK/BLAS wording that can be found in many routine specifications. See for example `vs` parameter of `?gees` to give an example that came to my mind first. The code does indeed not reference the arrays in case alpha is zero as you can see in the code.

TiborGY:
I would be interested to hear more opinions on this.
@langou ?

TiborGY:
I am still interested in having this merged, if there are no objections.

TiborGY:
Ping

langou (approved):

PR #1107, opened 2025-02-06
Fix the testsuite of xGEMMTR

grisuthedragon:
Introducing the GEMMTR routine leads to some errors in the test suite, these gets fixed with this PR. 

Namely this PR fixes:

- make the `SRNAME` in `XERBLA` more robust against strings passed from C (e.g. as  OpenBLAS does)
- adjust the maximum length of a function name from 6 to 7 characters to match the length of `xGEMMTR`.
- add a missing parameter in the error output for `c/zgemmtr`, causing an error if `c/zgemmtr` returns some wrong data

- [ ] The documentation has been updated. **Not required**
- [x] If the PR solves a specific issue, it is set to be closed on merge. **Recognized before an issue pops up**

### DIFF
--- a/BLAS/TESTING/cblat3.f
+++ b/BLAS/TESTING/cblat3.f
@@ -20,7 +20,7 @@
 *>
 *> The program must be driven by a short data file. The first 14 records
 *> of the file are read using list-directed input, the last 10 records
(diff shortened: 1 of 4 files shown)


langou (approved):

langou:
Thanks Martin. This is a great contribution.

PR #1116, opened 2025-03-27
Adds new icamax with exception handling

weslleyspereira:
Adds new ICAMAX as proposed in https://arxiv.org/abs/2207.09281 (see Appendix A).


BLAS/SRC/icamax.f90
```diff
@@ -0,0 +1,188 @@
...
+!
+!  .. Scalar Arguments ..
+   integer :: n, incx
+!
+!  .. Array Arguments ..
+   complex(wp) :: x(*)
+!  ..
+!  .. Local Scalars ..
+   integer :: i, j, ix, jx
+   real(wp) :: val, smax, scaledsmax
```
angsch, line 86: should `scaledsmax` be `logical`?

weslleyspereira: Yes, makes more sense. Thanks!

langou (approved):

PR #1117, opened 2025-03-27
Adds disclaimer about non listed values in ROTM

weslleyspereira:
There have been some research to verify correctness of exception handling in LAPACK led by @jacksonvanover. This is a comment missing in `rotm` saying it does not expect a flag that is not in the set {-2,-1,0,1}. I tagged it as "undefined behavior" so as to not touch the implementation. An alternative would be to add checks inside the routine.


TiborGY (approved):
LGTM

langou (approved):

PR #1119, opened 2025-04-07
fix missed error message in test for eigenvalue driver routines ed

dklyuchinskiy:
PR fixes missed error message "func_name failed the tests of the error exits" if checker is failed.

Affected **GEEV** and other routines.


langou (approved):

langou:
Thanks @dklyuchinskiy !

PR #1120, opened 2025-04-09
Call *rot to perform eigenvector update of *steqr

sh-zheng:
Call *rot instead of *lasr to perform eigenvector update of *steqr to fully utilize blas subroutines.

Code equivalence is verified with norm $\Vert{A-ZDZ^T}\Vert_2$. Entries is generated randomly.

Single Precision Case
|Dimension|Norm before PR|Norm after PR|
|-------|--------|---------|
|1000|8.621327e-05|8.621327e-05|
|2000|1.890181e-04|1.890181e-04|

Double Precision Case
|Dimension|Norm before PR|Norm after PR|
|-------|--------|---------|
|1000|2.297470e-13|2.297470e-13|
|2000|4.402642e-13|4.402642e-13|

Performance is measured in millisecond and shows an improvement. The platform is a Intel Xeon 1660 v3 @ 3GHz. As an additional test, we also measured the performance of subroutine *kteqr in PR #1049 .

Single Precision Case
|Dimension|Elapse before PR|Elapse after PR|Elapse of skteqr|
|-------|--------|---------|---------|
|1000|4281|3703|2359|
|2000|37063|31625|18781|

Double Precision Case
|Dimension|Elapse before PR|Elapse after PR|Elapse of dkteqr|
|-------|--------|---------|---------|
|1000|6063|5234|2562|
|2000|47499|40406|22469|

### DIFF
--- a/LAPACKE/src/lapacke_dsteqr.c
+++ b/LAPACKE/src/lapacke_dsteqr.c
@@ -59,23 +59,8 @@ lapack_int API_SUFFIX(LAPACKE_dsteqr)( int matrix_layout, char compz, lapack_int
         }
     }
 #endif
-    /* Additional scalars initializations for work arrays */
-    if( API_SUFFIX(LAPACKE_lsame)( compz, 'n' ) ) {
-        lwork = 1;
-    } else {
-        lwork = MAX(1,2*n-2);
-    }
-    /* Allocate memory for working array(s) */
-    work = (double*)LAPACKE_malloc( sizeof(double) * lwork );
-    if( work == NULL ) {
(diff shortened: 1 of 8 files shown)


sh-zheng:
Thank you very much for the review @angsch . I agree with your point. I indeed didn't take it into account in terms of algorithm efficiency in this PR. My original intention was to unify the existing algorithm of symmetric matrices and skew-symmetric matrices in #1049 .  It seems that a better approach would be to implement #1049 in the same way as steqr. I'll update that PR later. This PR will be closed.

PR #1129, opened 2025-05-17
Avoid memory corruption with non-finite inputs in ?GEEV

### LINKED ISSUE #1128: Calling ?GEEV with non-finite input leads to memory corruption in ?LASCAL
**Description**
See https://github.com/OpenMathLib/OpenBLAS/issues/5250 for full context and reproducer
Calling ?GEEV with non-finite input will fail in ?HSEQR. ?GEEV will then invoke ?LASCAL to undo any previous scaling of the input data, using INFO as a reference - with the existing guard for INFO.GT.0 covering only half the invocations.
The trivial fix would be to move the existing check for INFO.GT.0 into the outer `if (?CALEA) THEN` conditional - or add a check for INFO.GE.0 there. The same problem appears to be present in ?GEEVX, though I have not checked in detail.
**Checklist**

- [X] I've included a minimal example to reproduce the issue
- [X] I'd be willing to make a PR to solve this issue

martin-frbg:
**Description**
fixes #1128

### DIFF
--- a/SRC/cgeev.f
+++ b/SRC/cgeev.f
@@ -496,7 +496,7 @@
 *     Undo scaling if necessary
 *
    50 CONTINUE
-      IF( SCALEA ) THEN
+      IF( SCALEA .AND. INFO.GE.0) THEN
          CALL CLASCL( 'G', 0, 0, CSCALE, ANRM, N-INFO, 1,
      $                W( INFO+1 ),
      $                MAX( N-INFO, 1 ), IERR )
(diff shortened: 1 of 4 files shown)


Beautifulme8 (approved):

langou (approved):
Excellent. Thanks @martin-frbg

martin-frbg:
> Excellent. Thanks @martin-frbg

Excrement actually, I'm awfully sorry. Please apply #1142 to revert this

Issue #1138, opened 2025-06-23
DGELSD: SVD fails to converge for some matrices


MaartenBaert:
**Description**

I have found that DGELSD sometimes fails for certain matrices, even though the matrix is well-conditioned and does not contain any unreasonably large or small values. I have attached a 1602x1062 bidiagonal matrix that triggers the problem as an example (this is the smallest test case I have been able to find so far).

**Minimal example**

This includes a minimal Fortran example that triggers the problem, and the required data files with matrix data.
[example.zip](https://github.com/user-attachments/files/20873076/example.zip)

When the problem occurs, this example will produce the following error:
```
Error: DGELSD failed with info =           1
Note: The following floating-point exceptions are signalling: IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
```

Tested with the latest reference blas+lapack as well as various older versions, on Linux (RHEL8) with gcc/gfortran 8.5.0. Originally found in OpenBLAS: https://github.com/OpenMathLib/OpenBLAS/issues/5333

**Checklist**

- [x] I've included a minimal example to reproduce the issue
- [x] I'd be willing to make a PR to solve this issue

PR #1140, opened 2025-06-25
Fix documentation error and ordering bug in DLASD7

MaartenBaert:
**Description**

This fixes https://github.com/Reference-LAPACK/lapack/issues/1138.
Please refer to https://github.com/OpenMathLib/OpenBLAS/issues/5333 for a detailed discussion of the issue and fix.

**Checklist**

### DIFF
--- a/SRC/dlasd7.f
+++ b/SRC/dlasd7.f
@@ -99,7 +99,7 @@
 *>         On entry D contains the singular values of the two submatrices
 *>         to be combined. On exit D contains the trailing (N-K) updated
 *>         singular values (those which were deflated) sorted into
-*>         increasing order.
+*>         decreasing order.
 *> \endverbatim
 *>
 *> \param[out] Z
@@ -453,7 +453,7 @@
 *
 *        Check if singular values are close enough to allow deflation.
 *
-         IF( ABS( D( J )-D( JPREV ) ).LE.TOL ) THEN
+         IF( ( D( J )-D( JPREV ) ).LE.TOL ) THEN
 *
 *           Deflation is possible.
 *
@@ -489,7 +489,14 @@
             CALL DROT( 1, VF( JPREV ), 1, VF( J ), 1, C, S )
             CALL DROT( 1, VL( JPREV ), 1, VL( J ), 1, C, S )
             K2 = K2 - 1
-            IDXP( K2 ) = JPREV
+*
+*           Insert the deflated index in the correct position in IDXP.
+*           If J - JPREV is greater than 1, the indices in between
+*           must be shifted to preserve the correct output order.
+*
+            DO 130 JP = JPREV, J - 1
+               IDXP( K2 + J - 1 - JP ) = JP
+  130       CONTINUE
             JPREV = J
          ELSE
             K = K + 1


martin-frbg (dismissed):

SRC/dlasd7.f
```diff
@@ -99,7 +99,7 @@
 *>         On entry D contains the singular values of the two submatrices
 *>         to be combined. On exit D contains the trailing (N-K) updated
 *>         singular values (those which were deflated) sorted into
-*>         increasing order.
+*>         decreasing order.
```
langou, line 102: (1) The change (increasing &rarr; decreasing) is great. Thanks @MaartenBaert 

(2) As mentioned by @martin-frbg, it would better to report the same change to slasd7.f, slasd2.f and dlasd2.f. The same mistake is in these three files.

(3) Something to consider, I think "nonincreasing" would be better than "decreasing" in this context since we have D( I ) ≥ D(I+1), we do not necessarily have D( I ) > D(I+1).

langou: Most drivers routines use the terminology "ascending order" for HEEV (resp. descending order for SVD). Most drivers routines do not use the terminology "increasing order" for HEEV (resp. decreasing order for SVD). (This might be more due to comment-copy-paste than intentionality comment writing.)

Some examples for the HEEV drivers:
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dsyev.f#L86
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dsyevd.f#L92
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dsyevr.f#L240
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dsyevx.f#L167

Some examples for the SVD drivers:
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dgesvd.f#L47
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dgesvdx.f#L51

GESVDQ avoids to use any term and
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dgesvdq.f#L156


https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dgesvd.f#L121

MaartenBaert:
That's for the feedback, I've pushed equivalent changes to SLASD7, SLASD2 and DLASD2.

Interestingly, SLAED8/DLAED8 appears to contain code specifically to reorder the indices, presumably in order to solve the same problem I encountered, though it does this in a far more convoluted way:
```
            IF( K2+I.LE.N ) THEN
               IF( D( JLAM ).LT.D( INDXP( K2+I ) ) ) THEN
                  INDXP( K2+I-1 ) = INDXP( K2+I )
                  INDXP( K2+I ) = JLAM
                  I = I + 1
                  GO TO 90
               ELSE
                  INDXP( K2+I-1 ) = JLAM
               END IF
            ELSE
               INDXP( K2+I-1 ) = JLAM
            END IF
```
This is basically doing an insertion sort step to find the correct insertion position into the output array to avoid mis-ordering `D`. I *think* this is functionally equivalent to what I would have done:
```
            DO 85 JP = JLAM, J - 1
               INDXP( K2 + J - 1 - JP ) = JP
   85       CONTINUE
```
The documentation of SLAED8/DLAED8 is still wrong though (increasing -> decreasing) so I will change that.

MaartenBaert:
Regarding SLAED2/DLAED2/SLAED8/DLAED8, do you prefer to keep the existing (complicated) reordering code, or replace it with the new but simpler code?

martin-frbg:
Perhaps simplify the ..ED codes in a separate PR, if desired ? (Their rewrite is only tangentially related to the bug this one fixes)

PR #1148, opened 2025-07-23
Capture module file dependencies in CMake

ndingle-arm:
**Description**

The modulefile la_xisnan.F90 depends on the modulefile la_constants.f90 but this dependency is not expressed in the CMake rules. In some parallel builds we have observed failures where compilation of objects that depend on la_constants.mod starts before compilation of la_constants.f90 has finished.

This patch modifies CMakeLists.txt to ensure that la_constants.mod is built before any other object or modulefile that depends on it.

**Checklist**

### DIFF
--- a/SRC/CMakeLists.txt
+++ b/SRC/CMakeLists.txt
@@ -36,7 +36,9 @@
 #
 #######################################################################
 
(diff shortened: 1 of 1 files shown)


langou:
Thanks! Looks great to me. Julien.

PR #1149, opened 2025-07-29
Fix format string in LAPACK tests to match two-digit minor versions

grisuthedragon:
**Description**

After calling `ilaver` in the LAPACK tests, the version is displayed. But the minor version is a two-digit number and thus `I1` in the format string will display a `*` instead of  `12`. This PR fixes the format strings in the tests. 

**Checklist**
 - nothing apply here


langou (approved):
Nice catch. Thank!

PR #462, opened 2020-11-01
create INDEX64 target

epsilon-0:
Very initial proof of concept PR, related to https://github.com/Reference-LAPACK/lapack/issues/461

---

This adds a BUILD_INDEX64 option which manages names for libraries.
Should have no effect with the default set to OFF.
When switched on, it will build the 64-suffixed libs.

~~Not a lot of testing done yet.~~
~~Just checked that BLAS64, CBLAS64, LAPACK64 and LAPACKE64 are building, with correct cmake and pkg-config files.~~
1/11/2020:
~~BLAS and CBLAS tests are working fine, LAPACK tests are still failing.~~
All tests are passing!

TODO: 
Handle header file naming. 
Fix printf/fprintf so that they don't give warnings of incompatible types.

PR #1154, opened 2025-09-14
fix library names in `lapack64.pc`

mgorny:
**Description**

Add substitutions to `lapack.pc.in`, in order to ensure that the generated `lapack64.pc` file correctly references the index64 library and BLAS dependency.  It seems that other pkg-config templates have been updated as part of #462 but this one was omitted. As a result, finding a lapack64 via pkg-config gave non-index64 libraries.

**Checklist**

CC @epsilon-0 @langou


langou:
Thanks!

PR #1162, opened 2025-10-22
LAPACKE const correctness fixes

ACSimon33:
**Description**

Some matrix arguments in the LAPACKE interfaces were not specified as const, even though they are clearly marked as pure inputs on the Fortran side and are correctly specified as const in the LAPACK C interfaces.

I changed all the pure inputs to be const. That revealed some spurious transpositions of "output" matrices (row-major cases) that are, in fact, only inputs. I removed those as well.

**Checklist**


langou (approved):
Thanks. That's a lots of work and this is really nice to have proper 'const' in LAPACKE. Thanks a lot.

PR #1164, opened 2025-11-07
Implement MSVC complex types

mgorny:
**Description**

Implement the support for the `_Fcomplex` and `_Dcomplex` types used by MSVC in place of the standard C99 complex types.  This is necessary to make LAPACK build out of the box with MSVC, since it does not implement the default C99 types.

See: https://learn.microsoft.com/en-us/cpp/c-runtime-library/complex-math-support

I went for the simplest implementation possible, limiting the changes to swapping the default types used when MSVC is used as the compiler, also when `lapack_config.h` is not used.  I haven't added a `LAPACK_COMPLEX_*` ifdef for it, but I can do that if you prefer.

**Checklist**

### DIFF
--- a/LAPACKE/include/lapack.h
+++ b/LAPACKE/include/lapack.h
@@ -47,7 +47,11 @@
 #else
 #include <complex>
 #endif
+#if _MSC_VER
+#define lapack_complex_float    _Fcomplex
+#else
 #define lapack_complex_float    float _Complex
+#endif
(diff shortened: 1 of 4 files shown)


mgorny:
I don't understand your comment. These are the types MSVC defines. I don't have control over their names.

mgorny:
Thank you!

PR #1173, opened 2025-11-30
Update Function Arguments in `?gees`, `?geesx`, `?gges`, `?ggesx`, and `?gges3`

### LINKED ISSUE #1171: Provide explicit interfaces for external functions passed as arguments
Some routines such as [?GEES](https://github.com/Reference-LAPACK/lapack/blob/master/SRC/dgees.f), get functions as arguments. However, these arguments are declared just like `external`, so compilers can not actually check that passed arguments has proper types. It would be nice, if not only documentation will specify proper definition of such functions.

Actually, it is needed for generating safe Fortran 90 interfaces from the source code.

foxtran:
Instead of using external keyword for Function Arguments, this PR introduces explicit interfaces for Function Arguments. The code should be more safe now, and source code reflect documented behaviour. Later, it will help a lot to generate proper interfaces for these functions.

BIND(C) is used since code can be called from other than Fortran languages.

**Checklist**

- [x] The documentation has been updated: no need to update documentation.
- [x] Closes #1171

### DIFF
--- a/SRC/cgees.f
+++ b/SRC/cgees.f
@@ -208,8 +208,12 @@
       COMPLEX            A( LDA, * ), VS( LDVS, * ), W( * ), WORK( * )
 *     ..
 *     .. Function Arguments ..
-      LOGICAL            SELECT
-      EXTERNAL           SELECT
+      INTERFACE
+        LOGICAL FUNCTION SELECT_PROC_TYPE(EV) BIND(C)
+          COMPLEX EV
+        END FUNCTION SELECT_PROC_TYPE
+      END INTERFACE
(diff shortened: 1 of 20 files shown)


foxtran:
No, I definitely has no idea why appveyor fails.

martin-frbg:
Could be it's getting miscompiled, because the CI job running on Appveyor appears to be still using flang 11 from four years ago when the new flang was barely able to compile again. 
(I had to clamp the version number around then to keep the CI running at all, because conda was acting up and tried to update both flang and itself with the dependencies all wrong)

martin-frbg:
closing&reopening to trigger a new CI run

langou (approved):

PR #1179, opened 2026-01-07
Fix ASAN errors and FORMAT specifiers

ACSimon33:
**Description**

In the error tests of tall-skinny QR routines, the array `TAU` was of length `NMAX=2`, but the `GEMLQ` and `GEMQR` routines read the second and third entry. This was already fixed by https://github.com/Reference-LAPACK/lapack/commit/a1491006966d4486c727e54d5b53fbc1cd7d24a1 for the real routines. The complex routines still had this issue, and it was flagged by an address sanitizer. 

I increased the size of `TAU` to `5` since that is the expected minimal size of the array according to the documentation:

```
*> \param[out] T
*> \verbatim
*>          T is COMPLEX*16 array, dimension (MAX(5,TSIZE))
*>          On exit, if INFO = 0, T(1) returns optimal (or either minimal 
*>          or optimal, if query is assumed) TSIZE. See TSIZE for details.
*>          Remaining T contains part of the data structure used to represent Q.
*>          If one wants to apply or construct Q, then one needs to keep T 
*>          (in addition to A) and pass it to further subroutines.
*> \endverbatim
```

It doesn't need to be larger, since we never reach the part of the code that reads or writes the rest of the array. Additionally, I removed the unused array `T` in these tests.

There were also some FORMAT specifier lines longer than 72 characters in the BLAS tests, which the NAG Fortran compiler doesn't like by default. Those are also fixed now.

### DIFF
--- a/TESTING/LIN/cerrtsqr.f
+++ b/TESTING/LIN/cerrtsqr.f
@@ -73,8 +73,8 @@
       INTEGER            I, INFO, J, MB, NB
 *     ..
 *     .. Local Arrays ..
-      COMPLEX            A( NMAX, NMAX ), T( NMAX, NMAX ), W( NMAX ),
-     $                   C( NMAX, NMAX ), TAU(NMAX)
+      COMPLEX            A( NMAX, NMAX ), W( NMAX ),
+     $                   C( NMAX, NMAX ), TAU( 5 )
 *     ..
 *     .. External Subroutines ..
       EXTERNAL           ALAESM, CHKXER, CGEQR,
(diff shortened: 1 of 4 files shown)


TESTING/LIN/cerrtsqr.f
```diff
@@ -251,8 +251,9 @@ SUBROUTINE CERRTSQR( PATH, NUNIT )
 *
 *     CGEMLQ
 *
-      TAU(1)=1
-      TAU(2)=1
+      DO I = 1, 5
```
langou: Maybe a better fix could be to have a
```fortran
DO I = 1, NMAX
```
and then we would leave the 
```fortran
TAU(NMAX)
```
on line 77.

ACSimon33: The previous fix for the real routines, used `TAU(2*NMAX)` which was also a bit random and relied on the fact that NMAX >= 2. Since the routines expect an array of at least size 5, I'd like to use that size, even if it's not strictly necessary in this instance. (But implementation details could change ...)

PR #1221, opened 2026-03-28
Follow up on pr1146 (v2)

langou:
Follow up on PR #1146. ( Thanks to @yizeyi18 for the PR.)

While reviewing, @angsch suggested 

> Shouldn't jobu = 'f' and jobu = 'r' be included here too?

I implemented the suggestion. This ended being easier for me to do a substantial rewrite.

I followed the explanation at
https://github.com/Reference-LAPACK/lapack/blob/034ada14fa42534c64bb5cb684dfff6a9db69248/SRC/cgesvdq.f#L159-L198


LAPACKE/src/lapacke_cgesvdq_work.c
```diff
@@ -50,17 +50,41 @@ lapack_int API_SUFFIX(LAPACKE_cgesvdq_work)( int matrix_layout, char joba, char
...
+           ncols_u = n;
+        } else {
+           nrows_u = 1;
+           ncols_u = 1;
+        }
+
+        if( API_SUFFIX(LAPACKE_lsame)( jobv, 'a' ) ||
+                API_SUFFIX(LAPACKE_lsame)( jobv, 'v' ) ||
+                API_SUFFIX(LAPACKE_lsame)( jobv, 'r' ) ||
+                API_SUFFIX(LAPACKE_lsame)( joba, 'e' ) ) {
```
mgates3, line 81: A comment that `joba == e` is using V as workspace would be helpful. (Per later comment when transposing results.)

LAPACKE/src/lapacke_cgesvdq_work.c
```diff
@@ -78,81 +102,80 @@ lapack_int API_SUFFIX(LAPACKE_cgesvdq_work)( int matrix_layout, char joba, char
...
+                                   ldv );
+            }
         }
+
         /* Release memory and exit */
-        if( API_SUFFIX(LAPACKE_lsame)( jobv, 'a' ) || API_SUFFIX(LAPACKE_lsame)( jobv, 'v' ) ||
-                API_SUFFIX(LAPACKE_lsame)( jobv, 'r' ) ) {
-            LAPACKE_free( v_t );
-        }
+        if ( ( m > 0 ) && ( n > 0 ) ) LAPACKE_free( v_t );
```
mgates3: Ok, but since `free( NULL )` does nothing, it would be simpler to have just one exit level and free all the pointers:
```
exit_level_0:
    LAPACKE_free( v_t );  v_t = NULL;
    LAPACKE_free( u_t );  u_t = NULL;
    LAPACKE_free( a_t );  a_t = NULL;
```
I tend to reset pointers to NULL after freeing them, to avoid any possible confusion.

If desired, perhaps that can be a future simplification of all of LAPACKE.

mgates3 (dismissed):
Looks fine. Maybe a couple things could be improved later on.

LAPACKE/src/lapacke_cgesvdq_work.c
langou: Thanks. Good idea. I added this to the PR, this fits nicely with it. Thanks!

LAPACKE/src/lapacke_cgesvdq_work.c
langou: I am slowing down on that, so, as you suggest: "could be improved later on." So, if we get there, for sure, let us do a separate pull request. 

I do like the idea to reset pointers to NULL after freeing them, to avoid any possible confusion.

Here is why I am slowing down: While `free(NULL)` does nothing, is it true that `LAPACKE_free( NULL )` does nothing? What do we ask LAPACKE_free() to do? This proposed change would implicitly state that we require LAPACKE_free( NULL ) to do nothing.

I like the logic "if you fail to allocate at level 2, then do not forget to free level 0 and level 1 that were previously allocated." I feel it could be useful in some context, but maybe not for a C wrapper.

So all in all, are there more opinions about what @mgates3 suggest? In favor? Against?

mgates3 (approved):

PR #1226, opened 2026-04-10
cleanup: first batch of named-constant normalization in LAPACK sources

nakatamaho:
## Summary

This PR collects a small first batch of no-functional-change cleanup
in LAPACK sources by normalizing scalar constant spelling.

In particular, it replaces typed literal `1` values with the existing
routine-local named constant `ONE` in the `{s,d,c,z}bbcsd` family.

## Rationale

These routines already define named scalar parameters such as `ZERO`,
`ONE`, and `TEN`. Using those names consistently improves readability,
keeps the precision variants aligned, and avoids mixing routine-local
named constants with ad hoc typed literals.

## Scope

This PR is intentionally limited to mechanical consistency changes:
- replace existing scalar literals with already-defined named constants
- keep the s/d/c/z variants in sync
- no algorithmic or numerical behavior changes intended

## Non-goals

This PR does not:
- change convergence logic
- change thresholds or iteration behavior
- alter test expectations
- introduce algorithmic fixes

## Notes

This is intended as a small cleanup-first PR rather than a broad
refactor. If this direction is acceptable, similar purely mechanical
normalizations can be submitted in follow-up PRs.

### DIFF
--- a/SRC/cbbcsd.f
+++ b/SRC/cbbcsd.f
@@ -351,10 +351,9 @@
 *     .. Parameters ..
       INTEGER            MAXITR
       PARAMETER          ( MAXITR = 6 )
-      REAL               HUNDRED, MEIGHTH, ONE, TEN, TWO, ZERO
+      REAL               HUNDRED, MEIGHTH, ZERO, ONE, TEN
       PARAMETER          ( HUNDRED = 100.0E0, MEIGHTH = -0.125E0,
-     $                     ONE = 1.0E0, TEN = 10.0E0, TWO = 2.0E0,
-     $                     ZERO = 0.0E0 )
+     $                     ZERO = 0.0E0, ONE = 1.0E0, TEN = 10.0E0 )
       COMPLEX            NEGONECOMPLEX
       PARAMETER          ( NEGONECOMPLEX = (-1.0E0,0.0E0) )
       REAL               PIOVER2
--- a/SRC/cbbcsd.f
+++ b/SRC/cbbcsd.f
@@ -351,9 +351,10 @@
 *     .. Parameters ..
       INTEGER            MAXITR
       PARAMETER          ( MAXITR = 6 )
-      REAL               HUNDRED, MEIGHTH, ONE, TEN, ZERO
+      REAL               HUNDRED, MEIGHTH, ONE, TEN, TWO, ZERO
       PARAMETER          ( HUNDRED = 100.0E0, MEIGHTH = -0.125E0,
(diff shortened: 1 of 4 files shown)


SRC/cbbcsd.f
```diff
@@ -351,9 +351,10 @@ SUBROUTINE CBBCSD( JOBU1, JOBU2, JOBV1T, JOBV2T, TRANS, M, P,
 *     .. Parameters ..
       INTEGER            MAXITR
       PARAMETER          ( MAXITR = 6 )
-      REAL               HUNDRED, MEIGHTH, ONE, TEN, ZERO
+      REAL               HUNDRED, MEIGHTH, ONE, TEN, TWO, ZERO
       PARAMETER          ( HUNDRED = 100.0E0, MEIGHTH = -0.125E0,
-     $                     ONE = 1.0E0, TEN = 10.0E0, ZERO = 0.0E0 )
+     $                     ONE = 1.0E0, TEN = 10.0E0, TWO = 2.0E0,
+     $                     ZERO = 0.0E0 )
```
langou: Did you mean to commit this change? (So declaring TWO and defining TWO as a parameter = 2.0e0?)

I do not think you meant this.

SRC/cbbcsd.f
```diff
@@ -576,7 +577,7 @@ SUBROUTINE CBBCSD( JOBU1, JOBU2, JOBV1T, JOBV2T, TRANS, M, P,
                END IF
             ELSE
                NU = SIGMA21
-               MU = SQRT( 1.0 - NU**2 )
+               MU = SQRT( ONE - NU**2 )
```
langou, line 579: This change is all good with me.

SRC/cbbcsd.f
nakatamaho: You're right — this change was unnecessary. It came from an MPLAPACK-side change where
 `PIOVER2` was replaced with `pi(zero) / two`:

 ```cpp
 -    const REAL piover2 = 1.5707963267948966192313216916397514421;
 +    const REAL two = 2.0;
 +    const REAL piover2 = pi(zero) / two;
 ```

 So introducing `TWO` here did not serve any real purpose. I'll fix it.

SRC/cbbcsd.f
nakatamaho: thank you very much!

SRC/cbbcsd.f
nakatamaho: Fixed, please check my newer commit, thank you very much.

PR #1234, opened 2026-04-11
cleanup: separate real and complex zero/one constants in {c,z}csdts

nakatamaho:
## Summary

This cleanup makes the scalar and complex constants explicit in the CSD
test routines.

`ccsdts.f` and `zcsdts.f` previously used separate real constants
(`REALZERO`/`REALONE`) alongside complex `ZERO`/`ONE`. This change
switches to real `ZERO`/`ONE` and complex `CZERO`/`CONE`, then updates
the affected call sites to use constants with the intended type.

## Changes

- define real `ZERO`/`ONE` and complex `CZERO`/`CONE`
- update `{C,Z}LASET`, `{C,Z}GEMM`, and `{C,Z}HERK` call sites to use
  the matching real or complex constants
- replace literal imaginary-zero arguments in `CMPLX`/`DCMPLX` with the
  named real `ZERO` constant
- remove the now-redundant `REALZERO`/`REALONE` naming in these routines

## Notes

No functional change intended.

This is a test-code cleanup that makes the constant usage more explicit
and avoids mixing real and complex meanings under the same `ZERO`/`ONE`
names.

### DIFF
--- a/TESTING/EIG/ccsdts.f
+++ b/TESTING/EIG/ccsdts.f
@@ -246,10 +246,10 @@
 *  =====================================================================
 *
 *     .. Parameters ..
-      REAL               REALONE, REALZERO
-      PARAMETER          ( REALONE = 1.0E0, REALZERO = 0.0E0 )
-      COMPLEX            ZERO, ONE
-      PARAMETER          ( ZERO = (0.0E0,0.0E0), ONE = (1.0E0,0.0E0) )
+      REAL               ZERO, ONE
+      PARAMETER          ( ZERO = 0.0E0, ONE = 1.0E0 )
+      COMPLEX            CZERO, CONE
+      PARAMETER          ( CZERO = (0.0E0,0.0E0), CONE = (1.0E0,0.0E0) )
       REAL               PIOVER2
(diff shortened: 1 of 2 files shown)


langou (approved):

mgates3:
This may just be personal preference, but I would actually argue that the original code is nicer. If we think about translating code between complex and real, or if we had a templated implementation (as in C++), we want `zero` and `one` to be the native type (real (s, d) or complex (c, z)), and to call special attention to where we specifically need a real variable, such as the `realzero` in `herk`. This minimizes the differences between real and complex code; compare dcsdts.f zcsdts.f. There are many complex routines in LAPACK that use `zero` and `one` as complex numbers; changing these to `czero` and `cone` would introduce needless differences compared to their real counterparts. Also -- and this is just a guess -- I think in Fortran, if a developer accidentally passes a complex zero to a routine expecting a real value, such as herk, there won't be any harm, the imaginary part is just ignored; but passing a real zero to a routine expecting a complex value can cause strange, hard to reproduce and debug behavior from the uninitialized imaginary part.

I'm not advocating to revert this PR. I'm just saying I wouldn't do this kind of variable renaming all over the library. I think it makes maintenance more difficult, not less.

PR #1235, opened 2026-04-11
Add a new algorithm for the generalized eigenvalue problem

vlovero:
Currently, the algorithms used for the generalized eigenvalue algorithm have a few bottlenecks
 1. The hessenberg-triangular reduction step does not scale very well and givens rotations must be accumulated which adds a lot of redundant work
 2. the QZ algorithm can experience drastic slowdowns in the presence of infinite eigenvalues

The first issue can be addressed using a different HT reduction approach from [Steel et al](https://journals.uwyo.edu/index.php/ela/article/view/6483). The second issue can solved by identifying and deflating all of the infinite eigenvalues before the HT step.

These two bottlenecks are addressed in the xGGEV4 routines in this fork leading to 3-5x speedups when computing eigenvalues compared to xGGEV3.

### DIFF
--- a/SRC/cggdef.f
+++ b/SRC/cggdef.f
@@ -1,0 +1,315 @@
+*> \brief \b CGGDEF
+*
+* =========== DOCUMENTATION ===========
+*
+*> \par Purpose:
+* =============
+*>
+*> \details \b Purpose:
+*> \verbatim
+*>
+*> CGGDEF performs the deflation of infinite eigenvalues in the generalized
+*> eigenvalue problem.
+*> It applies QR and RQ factorizations to 
+*> sub-blocks of A and B to properly isolate and deflate these eigenvalues.
+*> \endverbatim
+*
+* Arguments:
+* ==========
+*
+*> \param[in] COMPVL
+*> \verbatim
+*>          COMPVL is LOGICAL
+*>          = .TRUE.:  Compute the left Schur vectors (VL).
+*>          = .FALSE.: Do not compute the left Schur vectors.
+*> \endverbatim
+*>
+*> \param[in] COMPVR
+*> \verbatim
+*>          COMPVR is LOGICAL
+*>          = .TRUE.:  Compute the right Schur vectors (VR).
+*>          = .FALSE.: Do not compute the right Schur vectors.
+*> \endverbatim
+*>
+*> \param[in] N
+*> \verbatim
+*>          N is INTEGER
+*>          The order of the matrices A, B, VL, and VR.
+*>          N >= 0.
+*> \endverbatim
+*>
+*> \param[in] K
+*> \verbatim
+*>          K is INTEGER
+*>          The number of infinite eigenvalues to deflate (the block size).
+*>          0 <= K <= N.
+*> \endverbatim
+*>
+*> \param[in,out] A
+*> \verbatim
+*>          A is COMPLEX array, dimension (LDA, N)
+*>          On entry, the matrix A to be deflated.
+*>          On exit, A has been updated by the unitary transformations.
+*> \endverbatim
+*>
+*> \param[in] LDA
+*> \verbatim
+*>          LDA is INTEGER
+*>          The leading dimension of the array A.  LDA >= max(1,N).
+*> \endverbatim
+*>
+*> \param[in,out] B
+*> \verbatim
+*>          B is COMPLEX array, dimension (LDB, N)
+*>          On entry, the matrix B to be deflated.
+*>          On exit, B has been updated by the unitary transformations.
+*> \endverbatim
+*>
+*> \param[in] LDB
+*> \verbatim
+*>          LDB is INTEGER
+*>          The leading dimension of the array B.  LDB >= max(1,N).
+*> \endverbatim
+*>
+*> \param[in,out] VL
+*> \verbatim
+*>          VL is COMPLEX array, dimension (LDVL, N)
+*>          If COMPVL = .TRUE., the left Schur vectors are accumulated in VL.
+*> \endverbatim
+*>
+*> \param[in] LDVL
+*> \verbatim
+*>          LDVL is INTEGER
+*>          The leading dimension of the array VL.
+*>          If COMPVL = .TRUE., LDVL >= max(1,N).
+*> \endverbatim
+*>
+*> \param[in,out] VR
+*> \verbatim
+*>          VR is COMPLEX array, dimension (LDVR, N)
+*>          If COMPVR = .TRUE., the right Schur vectors are accumulated in VR.
+*> \endverbatim
+*>
+*> \param[in] LDVR
+*> \verbatim
+*>          LDVR is INTEGER
+*>          The leading dimension of the array VR.
+*>          If COMPVR = .TRUE., LDVR >= max(1,N).
+*> \endverbatim
+*>
+*> \param[out] WORK
+*> \verbatim
+*>          WORK is COMPLEX array, dimension (MAX(1,LWORK))
+*>          On exit, if INFO = 0, WORK(1) returns the optimal LWORK.
+*> \endverbatim
+*>
+*> \param[in] LWORK
+*> \verbatim
+*>          LWORK is INTEGER
+*>          The dimension of the array WORK.
+*>          LWORK >= max(1, 2*N).
+*>          If LWORK = -1, then a workspace query is assumed;
+*>          the routine
+*>          only calculates the optimal size of the WORK array.
+*> \endverbatim
+*>
+*> \param[out] INFO
+*> \verbatim
+*>          INFO is INTEGER
+*>          = 0:  successful exit.
+*>          < 0:  if INFO = -i, the i-th argument had an illegal value.
+*> \endverbatim
+      SUBROUTINE CGGDEF( COMPVL, COMPVR, N, K, A, LDA, B, LDB, VL,
+     $                   LDVL, VR, LDVR, WORK, LWORK, INFO )
+*
+* -- LAPACK-style Code --
+*
+* .. Scalar Arguments ..
+      LOGICAL            COMPVL, COMPVR
+      INTEGER            INFO, K, LDA, LDB, LDVL, LDVR, LWORK, N
+* ..
+* .. Array Arguments ..
+      COMPLEX         A( LDA, * ), B( LDB, * ), VL( LDVL, * ),
+     $                   VR( LDVR, * ), WORK( * )
+* ..
+*
+* =====================================================================
+*
+* .. Local Scalars ..
+      LOGICAL            LQUERY
+      INTEGER            I, IINFO, J, LWKOPT
+* ..
+* .. External Subroutines ..
+      EXTERNAL           CGEQRF, CGERQF, CUNMQR, CUNMRQ, XERBLA
+* ..
+* .. Intrinsic Functions ..
+      INTRINSIC          REAL, CMPLX, INT, MAX
+* ..
+* .. Executable Statements ..
+*
+* Test the input parameters.
+*
+      INFO = 0
+      LQUERY = ( LWORK.EQ.-1 )
+*
+      IF( N.LT.0 ) THEN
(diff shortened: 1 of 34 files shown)


thijssteel:
This is great work!

Here are some initial comments, I'll try to give your code a more detailed read through when I find some time.

1. I think QZ really has three bottlenecks: the two you pointed out, but the eigenvector calculations can also be problematic. It amazes me that you still get such speedups in xGGEV3 without modifying that routine.
2. IterHT, while really fast, is not guaranteed to converge even if infinite eigenvalues are deflated (or at least, I'm not skilled enough to prove that it does.) I think that at the very least, you would need to detect this kind of convergence failure and report the error to the user. From a quick look at the PR, it doesn't seem like that is the case here? Maybe we can switch to the slower variant if that occurs?
3. I never really investigated the impact of IterHT on the accuracy of the eigenvalues. I was still operating under "if the backward error is small, that is all the user can ask from us" at the time. It wouldn't surprise me if some users prefer the slower version if it gives more accurate results. In that sense, making a new routine in addition to xGGEV3 makes sense.

vlovero:
I agree with the eigenvector computation and that's something I'm also working on improving! I've also been trying to prove convergence but haven't managed to come up with anything.

I still need to do a lot more testing for the accuracy, but in my initial tests, I've found that the xggev4 routine gives more accurate results most of the time, but can still struggle on a few edge cases such as when the norms of A and B have very magnitudes.

I'm happy to run the method through any tests!

thijssteel:
Now that you say it like that, the improved eigenvalue accuracy actually makes a lot of sense. Based on a gut feeling, I think that might be because you have cases with a lot of infinite eigenvalues. And deflating those beforehand can very significantly improve the accuracy of all eigenvalues.

It's something that has been known for a while, but I guess that before this PR, there weren't a lot of publically available tools to deflate those eigenvalues that aren't obvious from the nonzero structure of the matrices.

vlovero:
I've created a blocked variant of dtgevc. On my laptop, it's about 2-6x faster than the current unblocked code. I still need to incorporate things like scaling for under/overflow and `howmny`/`select` options then convert it to fortran, but initial tests and benchmarks are promising. Here are some timings using a block size of 32:

size | dgtevc3 | dgtevc | rel. runtime
-- | -- | -- | --
500 | 14.7 ms | 21.0 ms | 1.43
707 | 33.7 ms | 59.4 ms | 1.76
1000 | 72.3 ms | 167 ms | 2.31
1414 | 161 ms | 475 ms | 2.95
2000 | 376 ms | 1361 ms | 3.62
2828 | 908 ms | 3928 ms | 4.33
4000 | 2207 ms | 11317 ms | 5.13
5657 | 5480 ms | 32832 ms | 5.99
8000 | 14093 ms | 92807 ms | 6.59

I can incorporate it into ggev4 when I finish (assuming it passes all of the tests)

thijssteel:
I think that dtgevc fits better in a separate PR. Especially since ggev3 also benefits from it. I'll have a look at the rest of this PR today.

SRC/cggdef.f
```diff
@@ -0,0 +1,315 @@
...
+*> \verbatim
+*>          K is INTEGER
+*>          The number of infinite eigenvalues to deflate (the block size).
+*>          0 <= K <= N.
+*> \endverbatim
+*>
+*> \param[in,out] A
+*> \verbatim
+*>          A is COMPLEX array, dimension (LDA, N)
+*>          On entry, the matrix A to be deflated.
```
thijssteel, line 51: What are the requirements of A? Is it assumed with or without checking that these infinite eigenvalues are present in the beginning of the matrix? At the end? At least state that B indeed needs to be singular.

SRC/cggprp.f
```diff
@@ -0,0 +1,308 @@
...
+*> \par Purpose:
+* =============
+*>
+*> \details \b Purpose:
+*> \verbatim
+*>
+*> CGGPRP is a preprocessing routine for the generalized 
+*> eigenvalue problem.
+*> It prepares the matrices A and B by applying 
+*> an initial Rank-Revealing QR (RRQR) factorization on B, transposing
```
thijssteel, line 14: I would rephrase this. What it does is it applies a rank revealing RQ factorization to B. But since that is not implemented in LAPACK, we do some transposes and transformations to get that RQ factorization via the QR factorization. That makes it clearer.

SRC/cggprp.f
```diff
@@ -0,0 +1,308 @@
...
+* Compute norm of B and tolerance
+*
+      NORMB = CLANGE( 'F', N, N, B, LDB, RWORK )
+      TOL = NORMB * SLAMCH( 'E' )
+*
+* B = B^T
+*
+      CALL CLATRN( N, B, LDB )
+*
+* B = B[:, ::-1]
```
thijssteel, line 225: I'm not too familiar with this notation. Perhaps there is something more standard or perhaps you could clarify at bit.

SRC/cggprp.f
thijssteel: Perhaps we can also try investigating whether the RRRQ is really necessary, a simple RQ would really simplify the code and remove the requirement for the explicit transpose.

SRC/clatrn.f
```diff
@@ -0,0 +1,125 @@
...
+      INTEGER            OMP_GET_MAX_THREADS
+      EXTERNAL           OMP_GET_MAX_THREADS
+#endif
+* ..
+* .. Intrinsic Functions ..
+      INTRINSIC          MIN, CONJG
+* ..
+* .. Executable Statements ..
+*
+#if defined(_OPENMP)
```
thijssteel, line 73: I'm not familiar enough with openmp in lapack to check this. @langou do you know someone who can make sure this won't cause issues?

SRC/cgghd4.f
```diff
@@ -0,0 +1,369 @@
...
+*>          LWORK is INTEGER
+*>          The dimension of the array WORK.
+*>          If LWORK = -1, then a workspace query is assumed.
+*> \endverbatim
+*>
+*> \param[out] INFO
+*> \verbatim
+*>          INFO is INTEGER
+*>          = 0:  successful exit.
+*>          < 0:  if INFO = -i, the i-th argument had an illegal value.
```
thijssteel, line 130: This is an iterative algorithm that can fail, I would add an option to return a positive info argument is this is the case. This could potentially indicate how many columns have successfully been reduced.

SRC/cgghd4.f
```diff
@@ -0,0 +1,369 @@
...
+      END DO
+      IF (AMAX.EQ.RZERO) THEN
+         AMAX = RONE
+      END IF
+      TOL = EPS * AMAX
+
+*
+* Outer reduction loop: iterate until K reaches IHI or NITER hits 50.
+*
+      DO WHILE( K.LT.IHI .AND. NITER.LT.50 )
```
thijssteel, line 269: I think 50 is a bit large. If this hasn't converged after ~10 iterations, I think you are much better off just using XGGHD3.

SRC/cgghd4.f
```diff
@@ -0,0 +1,369 @@
...
+*> \verbatim
+*>          WORK is COMPLEX array, dimension (MAX(1,LWORK))
+*>          On exit, if INFO = 0, WORK(1) returns the optimal LWORK.
+*> \endverbatim
+*>
+*> \param[in] LWORK
+*> \verbatim
+*>          LWORK is INTEGER
+*>          The dimension of the array WORK.
+*>          If LWORK = -1, then a workspace query is assumed.
```
thijssteel, line 123: This routine requires a very large workspace (N**2). I think this is mostly fine, because the user already needs to store 4 N X N matrices. But perhaps it would still be useful to make this clear to the user?

SRC/cgghd4.f
```diff
@@ -0,0 +1,369 @@
...
+*
+         CALL CLACPY( 'A', NCOLS, NCOLS, A( K, K ), LDA, WORK, NCOLS )
+*
+* CTRSM to solve X * B = A
+*
+         CALL CTRSM( 'R', 'U', 'N', 'N', NCOLS, NCOLS, 
+     $               CMPLX( RONE / AMAX, RZERO ), B( K, K ), LDB, 
+     $               WORK, NCOLS )
+*
+* Hessenberg reduction of X
```
thijssteel, line 316: Maybe write this out, so that Qc has been defined for the next comment.

SRC/cgghd4.f
```diff
@@ -0,0 +1,369 @@
...
+            DO I = 1, COL_LEN
+               A( K + 1 + I, K ) = CZERO
+            END DO
+            K = K + 1
+         END DO
+
+         IF (K.GE.IHI) THEN
+            EXIT
+         END IF
+
```
thijssteel, line 302: I'm not very sure about this, but I think it is important that empty lines are actually comments. Otherwise, the coverage tracker gets confused. But that might be about another projects I'm misremembering. Someone else will likely confirm for me.

SRC/cgghd4.f
```diff
@@ -0,0 +1,369 @@
...
+*> \par Purpose:
+* =============
+*>
+*> \details \b Purpose:
+*> \verbatim
+*>
+*> CGGHD4 reduces a pair of complex matrices (A,B) to generalized upper
+*> Hessenberg form using unitary transformations, where A is a
+*> Hessenberg matrix and B is upper triangular. This routine
+*> employs a blocked algorithm (Steel et al) to improve efficiency
```
thijssteel, line 14: I'd prefer a reference to the actual paper. I've written multiple algorithms for fast HT reductions and a reference would make it clear which one.

SRC/cggev4.f
```diff
@@ -0,0 +1,379 @@
...
+      WORKNEEDED = MAX( WORKNEEDED, INT( REAL( DUMMY( 1 ) ) ) )
+      CALL CLAQZ0( JOBQZ, JOBVL, JOBVR, N, 1, N, A, LDA, B, LDB,
+     $             ALPHA, BETA, VL, LDVL, VR, LDVR, DUMMY,
+     $             -1, RWORK, 0, INFO )
+      WORKNEEDED = MAX( WORKNEEDED, INT( REAL( DUMMY( 1 ) ) ) )
+*
+* CGGBAL does not use a workspace query; it always needs 6*N
+* elements. Ensure the workspace is large enough to accommodate it.
+*
+      WORKNEEDED = MAX( WORKNEEDED, 6*N )
```
thijssteel, line 272: I don't think you use XGGBAL anymore. Rather than removing this extra workspace and the comment. I would recommend still applying XGGBAL before the preprocessing. Identifying zero columns before trying detect rank deficiencies numerically seems like a good idea.

SRC/cgghd4.f
```diff
@@ -0,0 +1,369 @@
...
+            CALL CUNMHR( 'R', 'N', N, NCOLS, 1, NCOLS, WORK, NCOLS,
+     $                   WORK( IWORK ), Q( 1, K ), LDQ,
+     $                   WORK( IWORK+NCOLS ), LWORK-IWORK-NCOLS+1,
+     $                   IERR )
+         END IF
+
+*
+* RQ factorization of B(K:IHI, K:IHI)
+*
+         CALL CGERQF( NCOLS, NCOLS, B( K, K ), LDB, WORK( IWORK ),
```
thijssteel, line 339: Especially if we replace the RRQR factorization with just an RQ, this RQ factorization here could reveal extra infinite eigenvalues. Perhaps these could also be detected and deflated?

thijssteel (changes requested):
Overall, a very nice PR. The implementation is quite a bit nicer than my original implementation of the algorithm.

SRC/cggprp.f
vlovero: I've found the RRRQ is necessary for efficiency of the IterHT algorithm. In some of my initial tests IterHT was requiring 3-4 iterations instead of just one when using standard RQ. The ordering of the diagonal entries of R in the RRRQ have been useful too when trying to prove tighter convergence bounds for IterHT.

PR #1236, opened 2026-04-13
cleanup: normalize ZERO/ONE constant order in TESTING/EIG error-exit tests

nakatamaho:
This PR normalizes the declaration and PARAMETER definition order
of the named constants ZERO and ONE across TESTING/EIG error-exit
test routines:

- {s,d,c,z}errec
- {s,d,c,z}erred

Specifically, the order is unified to:
    ZERO, ONE

This matches the prevailing style used in other parts of LAPACK
and improves consistency for automated translation and maintenance
(e.g., fable-based workflows in MPLAPACK).

No functional changes are intended. This is a pure cleanup.

### DIFF
--- a/TESTING/EIG/cerrec.f
+++ b/TESTING/EIG/cerrec.f
@@ -69,8 +69,8 @@
 *     .. Parameters ..
       INTEGER            NMAX, LW
       PARAMETER          ( NMAX = 4, LW = NMAX*( NMAX+2 ) )
-      REAL               ONE, ZERO
-      PARAMETER          ( ONE = 1.0E0, ZERO = 0.0E0 )
+      REAL               ZERO, ONE
+      PARAMETER          ( ZERO = 0.0E0, ONE = 1.0E0 )
 *     ..
 *     .. Local Scalars ..
       INTEGER            I, IFST, ILST, INFO, J, M, NT
--- a/TESTING/EIG/cerred.f
+++ b/TESTING/EIG/cerred.f
@@ -83,8 +83,8 @@
 *     .. Parameters ..
       INTEGER            NMAX, LW
       PARAMETER          ( NMAX = 4, LW = 5*NMAX )
-      REAL               ONE, ZERO
-      PARAMETER          ( ONE = 1.0E0, ZERO = 0.0E0 )
+      REAL               ZERO, ONE
+      PARAMETER          ( ZERO = 0.0E0, ONE = 1.0E0 )
 *     ..
 *     .. Local Scalars ..
       CHARACTER*2        C2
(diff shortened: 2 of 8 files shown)


langou:
On the one hand, I think LAPACK has been trying to keep alphabetical order when declaring variable. 

https://github.com/Reference-LAPACK/lapack/blob/f6fb1114abedc83dcde51740d6318c390d8cdd89/SRC/dgeev.f#L209-L238

In that sense, first ONE then ZERO makes sense to me. 

On the other hand, this specific 
```fortran
*     .. Parameters ..
      DOUBLE PRECISION   ZERO, ONE
      PARAMETER          ( ZERO = 0.0D0, ONE = 1.0D0 )
```
is really not consistent in LAPACK.
Indeed
```unix
grep PARAMETER SRC/d*f | grep ZERO | grep ONE
```
We can see that 176 subroutines uses first ZERO then ONE, and 96 subroutines use first ONE then ZERO.
(Only counting in SRC/ double precision subroutines.)

All in all, I am fine with anything. I am fine going one way, or the other way, or leaving it all as the mess it is. Not everything has to be consistent. But happy to approve the pull request.

nakatamaho:
Thanks for the flexibility and for approving.

That said, one thing I’ve learned through this process is that we do need some level of code discipline. Otherwise we keep revisiting the same style/consistency issues in every PR.

My current thinking is:
- Contributors don’t have to strictly follow all conventions when submitting
- But changes that move the code toward a consistent style should be accepted

For example:

- Avoid hard-coded constants like 0.5, -0.125, 1.0D0, 0.0D0  
  → use named PARAMETERs such as ZERO, ONE, TWO (and CZERO, CONE for complex)

  This is not just stylistic. It helps:
  - keep type consistency (REAL vs COMPLEX)
  - avoid precision-dependent inconsistencies (especially across double, quad etc.)
     cf. https://github.com/Reference-LAPACK/lapack/blob/master/INSTALL/make.inc.gfortran-quad
  - centralize numerical definitions for future changes

- Use LSAME for character comparisons instead of direct comparisons

  This is also not just style. It ensures:
  - case-insensitive behavior consistent with LAPACK
  - portability across compilers and systems where character comparison may differ

To make this explicit and reduce ambiguity, I propose adding a short guideline document (e.g., CODE_DISCIPLINE.md).
This should help reduce friction in reviews and gradually improve overall code quality.

I can draft an initial version if that sounds reasonable.

mgates3:
I do like named parameters, although I'm not a fan of CZERO and CONE. See my comment on #1234. (FWIW, SLATE doesn't use czero or cone; it does occasionally use r_zero and r_one. See [SLATE's style guide](https://github.com/icl-utk-edu/slate/wiki/code-style-guide#constants) regarding constants.)

The order of ZERO, ONE seems a trivial matter not worth normalizing. It's unlikely to stay normalized, so any automated tool needs to deal with all the various permutations it can come in.

A style guide document could be nice -- I would name it something like "style-guide.md". It can get included in the doxygen.

PR #1242, opened 2026-04-13
SRC: use LSAME for NORM checks in condition routines

nakatamaho:
This PR updates the condition estimation routines to use LSAME for checking
  the one-norm NORM option.

  The affected routines already use LSAME to accept the equivalent 'O' option:

      ONENRM = NORM.EQ.'1' .OR. LSAME( NORM, 'O' )

  This PR makes the handling consistent by replacing the direct character
  comparison with LSAME:

      ONENRM = LSAME( NORM, '1' ) .OR. LSAME( NORM, 'O' )

  This is a cleanup-only change. It does not change the accepted NORM options
  or the numerical behavior of the routines.

### DIFF
--- a/SRC/cgbcon.f
+++ b/SRC/cgbcon.f
@@ -201,7 +201,7 @@
 *     Test the input parameters.
 *
       INFO = 0
-      ONENRM = NORM.EQ.'1' .OR. LSAME( NORM, 'O' )
+      ONENRM = LSAME( NORM, '1' ) .OR. LSAME( NORM, 'O' )
       IF( .NOT.ONENRM .AND. .NOT.LSAME( NORM, 'I' ) ) THEN
          INFO = -1
       ELSE IF( N.LT.0 ) THEN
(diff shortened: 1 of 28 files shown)


toxtran:
Any reason to do so? LSAME is used for case-insensitize checks which is not necessary for digits.

langou:
> Any reason to do so? LSAME is used for case-insensitive checks which is not necessary for digits.

One reason is that it could be considered good software engineering to isolate the comparison in a specific routine (i.e. 'LSAME') "just in case". For example, this could be convenient for automatic Fortran-to-C tools, etc. This would be akin to a library having its own "print", its own "malloc/free", its own etc., that way you can replace these with machine-specific / compiler-specific versions as needed. So adding an 'LSAME' here could be a good idea. I do not find this argument very strong. I am trying to answer the question!

langou (dismissed):

langou (approved):

nakatamaho:
I just didn’t know that Fortran does so. I thought that we have to use lsame for every string comparison. iPhoneから送信2026/04/19 16:06、toxtran ***@***.***>のメール:﻿toxtran left a comment (Reference-LAPACK/lapack#1242)
Any reason to do so? LSAME is used for case-insensitize checks which is not necessary for digits.

PR #1245, opened 2026-04-15
avoid large intermediates in complex LARTG

nakatamaho:
This PR rewrites the complex sine computation in `CLARTG` and `ZLARTG`
  to avoid forming unnecessarily large intermediate values.

  The existing implementation uses expressions of the following form:

      s = conjg(g) * ( f / sqrt(f2*h2) )

  and similarly in the scaled branch:

      s = conjg(gs) * ( fs / sqrt(f2*h2) )

  The existing scaling and branch conditions already keep `f2*h2` in a
  safe range. Therefore, this is not primarily an overflow or underflow
  fix for the product `f2*h2` itself.

  However, even when `sqrt(f2*h2)` is safely representable, the division

      f / sqrt(f2*h2)

  can still create a large intermediate value before it is multiplied by
  `conjg(g)`. This is mathematically equivalent to multiplying normalized
  factors instead:

      ( f / sqrt(f2) ) * ( conjg(g) / sqrt(h2) )

  This PR applies that rewrite in both the unscaled and scaled branches of
  `CLARTG` and `ZLARTG`.

  The goal is to keep intermediate values closer to unit scale while
  preserving the same mathematical result.

  Updated files:

  - `SRC/clartg.f90`
  - `SRC/zlartg.f90`

### DIFF
--- a/SRC/cbbcsd.f
+++ b/SRC/cbbcsd.f
@@ -351,9 +351,9 @@
 *     .. Parameters ..
       INTEGER            MAXITR
       PARAMETER          ( MAXITR = 6 )
-      REAL               HUNDRED, MEIGHTH, ONE, TEN, ZERO
+      REAL               HUNDRED, MEIGHTH, ZERO, ONE, TEN
       PARAMETER          ( HUNDRED = 100.0E0, MEIGHTH = -0.125E0,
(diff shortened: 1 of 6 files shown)


langou (approved):
This is a good idea. Thanks.

langou:
Note that, in addition to the LARTG fix, there are some cosmetic changes to BBCSD too in this PR. 
These changes are good too.

PR #1246, opened 2026-04-16
Appveyor CI: Don't clamp flang to the ancient version 11 anymore

martin-frbg:
**Description**
Hopefully this old kludge is no longer necessary...
**Checklist**


martin-frbg:
will follow up with a less messy git history

PR #1250, opened 2026-04-20
Allow using gfortran even with the Cray PE_ENV set to something else

### LINKED ISSUE #1118: Logic error in CheckLAPACKCompilerFlags.cmake
**Description**

https://github.com/Reference-LAPACK/lapack/blob/master/CMAKE/CheckLAPACKCompilerFlags.cmake#L38-L47

This produces the wrong flags in an environment where PE_ENV is `CRAY` but the Fortran_COMPILER_ID is `GNU`.

**Checklist**

- [ ] I've included a minimal example to reproduce the issue

martin-frbg:
**Description**
Fixes #1118 by checking if the gfortran compiler was requested directly


langou (approved):
Thanks @martin-frbg

PR #1257, opened 2026-04-22
Fixed wrong external symbol declarations

ACSimon33:
**Description**

This fixes two wrong external symbol declarations:
- `CGER` -> `CGERC`
- `ZLARF1` -> `ZLARF1L`

This was flagged by the Intel Classic compiler (no other compiler/linker complained ...).


langou (approved):
good catches! thanks Intel Classic compiler! thanks @ACSimon33!

PR #1278, opened 2026-05-12
Fix Index-64 conversion errors

ACSimon33:
**Description**

Compiling with `-Werror=conversion` and `-DBUILD_INDEX64` results in a lot of conversion warnings/errors. This MR fixes those by adding explicit conversions to the appropriate types or fixing wrong declarations. The conversion errors are mostly in the testing routines, but there are a few in the XBLAS-related routines as well. 

The three types of conversion warnings are:

- `Error: Possible change of value in conversion from COMPLEX(8) to REAL(8) at (1) [-Werror=conversion]`
- `Error: Possible change of value in conversion from INTEGER(8) to REAL(4) at (1) [-Werror=conversion]`
- `Error: Possible change of value in conversion from REAL(4) to INTEGER(8) at (1) [-Werror=conversion]`


langou (approved):
Thanks @ACSimon33. In favor of explicit casts and in favor of less compiler warnings on my end. Thanks.

PR #1279, opened 2026-05-13
Fix external declarations

ACSimon33:
**Description**

This PR is part of #1276 which introduces `_64` suffixed tests for the extended Index-64 API. Those tests exposed a lot of missing external declarations and also unused or wrong external declarations of subroutines and functions. This PR fixes all those external declarations.


langou:
That's great. Thanks!

langou (approved):

Issue #1183, opened 2026-01-13
lacpy routines are not optimal for copying the lower triangular part of a fat matrix


putianyi889:
**Description**
For the lower triangular part, the routines go through all columns and then below-diagonal entries in each column. However, for a matrix with a small height and a large width, unnecessary columns are looped, hurting performance.

Example for reference https://github.com/JuliaLang/LinearAlgebra.jl/pull/1541

**Checklist**

- [x] I've included a minimal example to reproduce the issue
- [x] I'd be willing to make a PR to solve this issue

martin-frbg:
Hmm, between the "let the compiler do this" and "let the user do this", I do not think I saw a clear argument for **not** implementing this trivial optimization in LAPACK ?

putianyi889:
My bad. I thought [#14](https://github.com/ilayn/semicolon-lapack/pull/14) fixed this.

Issue #1188, opened 2026-01-23
cblas_ctrmm invalid diag problem


edwsmyth:
Hi LAPACK team,

In https://github.com/Reference-LAPACK/lapack/blob/master/CBLAS/src/cblas_ctrmm.c, for the layout CblasColMajor case (line ~76) the code is missing the

         CBLAS_CallFromC = 0;
         RowMajorStrg = 0;
         return;

statements present in other error checks in this file and other trmm variants.

Thanks,
Ed

Issue #1090, opened 2025-01-06
Overly restrictive argument checking in LAPACKE_?gesvd


martin-frbg:
**Description**
As pointed out by Gonum's Dan Kortschak in https://github.com/OpenMathLib/OpenBLAS/issues/5051 , the LAPACKE bindings for the GESVD functions appear to be overly restrictive in the LDVT argument check performed on row-major calls:
 When the JOBVT argument is "N" and consequently the output matrix "VT" will not be referenced at all according to the provided documentation, the current code still insists on a nonzero LDVT, taking the INFO=12 error exit instead of providing a work size estimate in LAPACKE_?gesvd_work.c

It occurs to me that the code in e.g.
https://github.com/Reference-LAPACK/lapack/blob/6ec7f2bc4ecf4c4a93496aa2fa519575bc0e39ca/LAPACKE/src/lapacke_dgesvd_work.c#L75
should be conditional on `if( LAPACKE_lsame( jobvt, 'a' ) || LAPACKE_lsame( jobvt, 's' ) ) {` 

**Checklist**

- [x] I've included a minimal example to reproduce the issue (see in referenced OpenBLAS ticket)
- [x] I'd be willing to make a PR to solve this issue

martin-frbg:
(A minimal fix would be to change the preceding ternary operator that sets the minimum ncols_vt so that it is zero (like in the corresponding lapacke_?gesvdx_work functions) instead of one - however I see no requirement that this entirely unused value cannot be negative)

PR #1284, opened 2026-06-10
Misc fixes

jschueller:
random bugfixes:
- #244  DROTMG/SROTMG rescaling overwriting DH12
- #587  Mislabeled comments in dget38.f
- #771  LAPACKE_[cz]tgexc passing ifst/ilst by value
- #1090 GESVD LDU/LDVT checks not conditional on JOBU/JOBVT
- #1128 ?GEEV/?GEEVX undo-scaling with negative INFO
- #1188 cblas_ctrmm missing cleanup on Diag error path

### DIFF
--- a/LAPACKE/include/lapacke.h
+++ b/LAPACKE/include/lapacke.h
@@ -4153,14 +4153,14 @@ lapack_int LAPACKE_ctgexc( int matrix_layout, lapack_logical wantq,
                            lapack_complex_float* b, lapack_int ldb,
                            lapack_complex_float* q, lapack_int ldq,
                            lapack_complex_float* z, lapack_int ldz,
-                           lapack_int ifst, lapack_int ilst );
+                           lapack_int ifst, lapack_int* ilst );
 lapack_int LAPACKE_ztgexc( int matrix_layout, lapack_logical wantq,
                            lapack_logical wantz, lapack_int n,
                            lapack_complex_double* a, lapack_int lda,
                            lapack_complex_double* b, lapack_int ldb,
                            lapack_complex_double* q, lapack_int ldq,
                            lapack_complex_double* z, lapack_int ldz,
-                           lapack_int ifst, lapack_int ilst );
+                           lapack_int ifst, lapack_int* ilst );
 
 lapack_int LAPACKE_stgsen( int matrix_layout, lapack_int ijob,
                            lapack_logical wantq, lapack_logical wantz,
@@ -9933,14 +9933,14 @@ lapack_int LAPACKE_ctgexc_work( int matrix_layout, lapack_logical wantq,
                                 lapack_complex_float* b, lapack_int ldb,
                                 lapack_complex_float* q, lapack_int ldq,
                                 lapack_complex_float* z, lapack_int ldz,
-                                lapack_int ifst, lapack_int ilst );
+                                lapack_int ifst, lapack_int* ilst );
 lapack_int LAPACKE_ztgexc_work( int matrix_layout, lapack_logical wantq,
                                 lapack_logical wantz, lapack_int n,
                                 lapack_complex_double* a, lapack_int lda,
                                 lapack_complex_double* b, lapack_int ldb,
                                 lapack_complex_double* q, lapack_int ldq,
                                 lapack_complex_double* z, lapack_int ldz,
-                                lapack_int ifst, lapack_int ilst );
+                                lapack_int ifst, lapack_int* ilst );
 
 lapack_int LAPACKE_stgsen_work( int matrix_layout, lapack_int ijob,
                                 lapack_logical wantq, lapack_logical wantz,
--- a/LAPACKE/include/lapacke_64.h
+++ b/LAPACKE/include/lapacke_64.h
@@ -4119,14 +4119,14 @@ int64_t LAPACKE_ctgexc_64( int matrix_layout, lapack_logical wantq,
                            lapack_complex_float* b, int64_t ldb,
                            lapack_complex_float* q, int64_t ldq,
                            lapack_complex_float* z, int64_t ldz,
-                           int64_t ifst, int64_t ilst );
+                           int64_t ifst, int64_t* ilst );
 int64_t LAPACKE_ztgexc_64( int matrix_layout, lapack_logical wantq,
                            lapack_logical wantz, int64_t n,
                            lapack_complex_double* a, int64_t lda,
                            lapack_complex_double* b, int64_t ldb,
                            lapack_complex_double* q, int64_t ldq,
                            lapack_complex_double* z, int64_t ldz,
-                           int64_t ifst, int64_t ilst );
+                           int64_t ifst, int64_t* ilst );
 
 int64_t LAPACKE_stgsen_64( int matrix_layout, int64_t ijob,
                            lapack_logical wantq, lapack_logical wantz,
@@ -9900,14 +9900,14 @@ int64_t LAPACKE_ctgexc_work_64( int matrix_layout, lapack_logical wantq,
                                 lapack_complex_float* b, int64_t ldb,
                                 lapack_complex_float* q, int64_t ldq,
                                 lapack_complex_float* z, int64_t ldz,
-                                int64_t ifst, int64_t ilst );
+                                int64_t ifst, int64_t* ilst );
 int64_t LAPACKE_ztgexc_work_64( int matrix_layout, lapack_logical wantq,
                                 lapack_logical wantz, int64_t n,
                                 lapack_complex_double* a, int64_t lda,
                                 lapack_complex_double* b, int64_t ldb,
                                 lapack_complex_double* q, int64_t ldq,
                                 lapack_complex_double* z, int64_t ldz,
-                                int64_t ifst, int64_t ilst );
+                                int64_t ifst, int64_t* ilst );
 
 int64_t LAPACKE_stgsen_work_64( int matrix_layout, int64_t ijob,
                                 lapack_logical wantq, lapack_logical wantz,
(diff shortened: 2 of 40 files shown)


martin-frbg:
Isn't #1047 already handled by PR #1264 , whose author pinged it just days ago ? Similarly, I have #1251 already open to fix #1183
And I think a mixed bag of unrelated fixes is a bad idea for other reasons as well - makes it harder to revert a single change, should that ever become necessary, and obscures the link between individual issue and fix in the git history

jschueller:
my bad, lets drop those
fixes are in separate commits (referencing the respective issues) so it can still be reverted individually

langou:
I am reviewing this pull request.

Side comment: @martin-frbg wrote "I think a mixed bag of unrelated fixes is a bad idea for other reasons as well - makes it harder to revert a single change, should that ever become necessary, and obscures the link between individual issue and fix in the git history." I agree with @martin-frbg. For future, it would be better to have PR that are more focused. I understand that there are a collection of different commits in this PR, I would prefer a collection of separated PRs. This is a comment for the future, not for this specific PR. Thanks for the PR!

langou (dismissed):

langou:
[ Note that I approved the whole PR by mistake, I intended to only approve a commit. ]

langou:
With respect to 
https://github.com/Reference-LAPACK/lapack/pull/1284/changes/4d7c5fa75b42b0df2a8225c81e3f7a2895d38c79

> Per LAPACK docs, in [cz]tgexc, ilst is [in,out]. The LAPACKE C interface was passing both ifst and ilst by value as input-only scalars, so callers could never observe the updated ilst value. Change both ifst and ilst to lapack_int* pointers in the complex tgexc variants to match the real-precision stgexc/dtgexc interface and the underlying Fortran semantics.

WRT `ifst`, the behavior is not the same between c/z (complex arithmetic case) and s/d (real arithmetic case).

In real arithmetic, there are 2-by-2 blocks, and so `ifst` can be changed. See:
https://github.com/Reference-LAPACK/lapack/blob/54b6620d277c8de205941a6d03966e8db6b75683/SRC/dtgexc.f#L299-L301

There are only 1-by-1 blocks in complex arithmetic. So `ifst` is input only. As correctly indicated in the comments:
https://github.com/Reference-LAPACK/lapack/blob/54b6620d277c8de205941a6d03966e8db6b75683/SRC/ztgexc.f#L131

Right now, with this PR, we have the LAPACKE interface with in/out for `ifst` in complex, and the comment says input only so something needs to be changed. I suggest to leave `ifst` as input only for LAPACKE interface. (So this is a request to change the commit.)

WRT `ilst`, thanks for pointing the mistake for `ilst` for c/z (complex arithmetic case). I agree. 

`ilst` is in/out. 

In the complex case, we only have 1-by-1 blocks but the reordering can fail and then `ilst` is used to return when the error happen. (I would prefer this to be indicated in `INFO` but so be it.)

In the real case, we can have an error in the reordering and we also have the issue of 1-by-1 and 2-by-2 blocks which means that, even w/o errors, `ilst` is only known +/-1 and so the final value is returned by the function.

I repeat the request: please leave `ifst` as input for [c/z] in LAPACKE.

langou:
With respect to
https://github.com/Reference-LAPACK/lapack/commit/4d7c5fa75b42b0df2a8225c81e3f7a2895d38c79
I am not sure we are ready for `testing/` in LAPACKE. I agree that we should have something. I do not know that what this PR proposes is what we want. I do not know what we want. LAPACKE is tested by the test suite in [LAPACK++](https://github.com/icl-utk-edu/lapackpp/), and maybe this is how we want to go about it. I suggest to remove the `testing/` bit for now.

langou:
With respect to:
https://github.com/Reference-LAPACK/lapack/pull/1284/changes/3078feadd2b0bb26d438c9490ae6b12f1303b175

Can you please propagate the same changes from [c] to [s,c,z]?

langou:
Well, reviewing the PR, and going commit by commit was actually not too bad. It kind of worked on my end and I was worried about the mix-and-match of commits. At the end, this worked OK on my end.

There are two requests for change. 
1. please leave ifst as input for [c/z] in LAPACKE for [cz]tgexc
2. remove `testing/` from LAPACKE.

Otherwise, I am done reviewing the PR. I am ready to merge. 

This is a terrific amount of work, it fixes a lots of issues, thanks a lot.

jschueller:
ok, done

LAPACKE/include/lapacke.h
```diff
@@ -9932,15 +9932,15 @@ lapack_int LAPACKE_ctgexc_work( int matrix_layout, lapack_logical wantq,
...
-                                lapack_complex_double* b, lapack_int ldb,
-                                lapack_complex_double* q, lapack_int ldq,
-                                lapack_complex_double* z, lapack_int ldz,
-                                lapack_int ifst, lapack_int ilst );
+                                 lapack_logical wantz, lapack_int n,
+                                 lapack_complex_double* a, lapack_int lda,
+                                 lapack_complex_double* b, lapack_int ldb,
+                                 lapack_complex_double* q, lapack_int ldq,
+                                 lapack_complex_double* z, lapack_int ldz,
+                                 lapack_int* ifst, lapack_int* ilst );
```
langou: I think we want `lapack_int ifst` here.

LAPACKE/include/lapacke_64.h
```diff
@@ -9900,14 +9900,14 @@ int64_t LAPACKE_ctgexc_work_64( int matrix_layout, lapack_logical wantq,
...
-                                int64_t ifst, int64_t ilst );
+                                int64_t* ifst, int64_t* ilst );
 int64_t LAPACKE_ztgexc_work_64( int matrix_layout, lapack_logical wantq,
                                 lapack_logical wantz, int64_t n,
                                 lapack_complex_double* a, int64_t lda,
                                 lapack_complex_double* b, int64_t ldb,
                                 lapack_complex_double* q, int64_t ldq,
                                 lapack_complex_double* z, int64_t ldz,
-                                int64_t ifst, int64_t ilst );
+                                int64_t* ifst, int64_t* ilst );
```
langou: I think we want `int64_t* ifst` here

LAPACKE/src/lapacke_ztgexc_work.c
```diff
@@ -38,13 +38,13 @@ lapack_int API_SUFFIX(LAPACKE_ztgexc_work)( int matrix_layout, lapack_logical wa
                                 lapack_complex_double* b, lapack_int ldb,
                                 lapack_complex_double* q, lapack_int ldq,
                                 lapack_complex_double* z, lapack_int ldz,
-                                lapack_int ifst, lapack_int ilst )
+                                lapack_int* ifst, lapack_int* ilst )
```
langou: I think we want `lapack_int ifst` here.

jschueller:
yes, fixed the work functions

martin-frbg:
Umm, can you please update the PR text before this gets merged ? I get the impression you added another bunch of valuable fixes in the latest round of changes ?

jschueller:
no, I did not add commits, I just rebased to master, I had 12 then removed the 2 commits duplicating other PRs

langou (approved):

martin-frbg:
> no, I did not add commits, I just rebased to master, I had 12 then removed the 2 commits duplicating other PRs

Oh, ok - thought the gemmtr test fixes and lapacke_zgedmdq rename were new (guess it illustrates how _I_ have comprehension issues with mixed ~~pickles~~ fixes)

martin-frbg:
Whoops - just noticed that the "Guard GEEV/GEEVX undo-scaling against negative INFO" part of this PR (91c0f61) is **exactly** the same as my previous #1129 that we reverted in #1142 due to the test suite regressions it caused

ilayn:
Indeed multiple commits are difficult to spot check things for such reasons.

langou:
> Whoops - just noticed that the "Guard GEEV/GEEVX undo-scaling against negative INFO" part of this PR ([91c0f61](https://github.com/Reference-LAPACK/lapack/commit/91c0f61b8768eb43ab5d95741fd740b93ef6c938)) is **exactly** the same as my previous #1129 that we reverted in #1142 due to the test suite regressions it caused

Thanks @martin-frbg. Do you remember which failures were caused? Do you think they are happening now? I do not see them. I am tempted to not revert ([91c0f61](https://github.com/Reference-LAPACK/lapack/commit/91c0f61b8768eb43ab5d95741fd740b93ef6c938)). However if this was an issue in the past, we can revert. @martin-frbg, @jschueller, someone has an opinion here? Revert? Not revert? I am looking into it now and I am fine with "not revert". But if there are arguments to revert, it is easy to do.

martin-frbg:
I will need to retest - but it was around 150 errors in the xeigtst? for each precision, and seen not only by me - https://github.com/OpenMathLib/OpenBLAS/issues/5392 . Of course it is possible that some of the intervening code changes have closed the former failure paths.

martin-frbg:
OK, I don't see any errors resulting from this change with Reference-LAPACK - OpenBLAS however still reacts quite strongly - the difference almost certainly lies in the implementation of XERBLA (the OpenBLAS one returns to the caller after reporting an error).

langou:
> OpenBLAS however still reacts quite strongly - the difference almost certainly lies in the implementation of XERBLA (the OpenBLAS one returns to the caller after reporting an error).

It is good to have a few valid XERBLA implementations. The default LAPACK XERBLA `STOP` the execution of the program. The one in the TESTING does something funky to check error codes. The one in OpenBLAS returns to the caller after reporting an error.

The point is that as soon as there is a XERBLA call, if you returns to the caller after reporting an error, then LAPACK has a RETURN statement. All GEEV and GEEVX have the correct "RETURN" after the call to XERBLA. So I feel good about all this.

martin-frbg:
I agree that there has to be something funky going on, maybe it's another case where the linker happened to pick the wrong flavor of xerbla - returning from OpenBLAS' implementation should send it back into unadulterated LAPACK code... anyway this does not appear to have any bearing on the validity of this PR for Reference-LAPACK

PR #1285, opened 2026-06-11
bdsqr: Fix INFO return when NCVT=NRU=NCC=0

### LINKED ISSUE #242: Returned INFO when NCVT = NRU = NCC = 0
If NCVT = NRU = NCC = 0, the algorithm will try to converge using DLASQ1, if DLASQ1 fails, say, return 1 or 3, the algorithm will try to converge using the standard form.
https://github.com/Reference-LAPACK/lapack/blob/4f7f2aab27814a920613f27f4b0d5055a421dcaf/SRC/dbdsqr.f#L339-L346
If it fails again, `INFO` could be a value between 1 and N-1.
https://github.com/Reference-LAPACK/lapack/blob/4f7f2aab27814a920613f27f4b0d5055a421dcaf/SRC/dbdsqr.f#L853-L860
So, the description is wrong, when it says:
https://github.com/Reference-LAPACK/lapack/blob/4f7f2aab27814a920613f27f4b0d5055a421dcaf/SRC/dbdsqr.f#L178-L183
Furthermore, the comment "If INFO equals 2" and the if-statement mismatch.
I suggest remove the if-statement or update the description.

jschueller:
Problem: When NCVT = NRU = NCC = 0 (no singular vectors requested), BDSQR calls the dqds algorithm. If dqds failed with INFO = 1 or 3, the routine returned immediately with that error, even though the standard QR algorithm could still compute the singular values. Only INFO = 2 triggered the fallback.

Change `IF( INFO .NE. 2 ) RETURN` to `IF( INFO .EQ. 0 ) RETURN` after the dqds call, so the standard QR fallback runs on any dqds failure (INFO=1,2,3) and not only INFO=2.

Closes #242

### DIFF
--- a/SRC/cbdsqr.f
+++ b/SRC/cbdsqr.f
@@ -330,9 +330,10 @@
       IF( .NOT.ROTATE ) THEN
          CALL SLASQ1( N, D, E, RWORK, INFO )
 *
-*     If INFO equals 2, dqds didn't finish, try to finish
+*     If the dqds algorithm failed, try to finish with
+*     the standard QR algorithm
 *
-         IF( INFO .NE. 2 ) RETURN
+         IF( INFO .EQ. 0 ) RETURN
          INFO = 0
       END IF
 *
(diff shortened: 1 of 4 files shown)


langou (approved):

langou:
Looks good to me. Thanks!

martin-frbg (approved):

PR #1289, opened 2026-06-11
gelsy: fix ILAENV workspace query routines for TZRZF/UNMRZ

### LINKED ISSUE #676: ZGELSY uses ILAENV with ZGERQF and ZUNMRQ
ZGELSY uses ILAENV with ZGERQF and ZUNMRQ for workspace computation
https://github.com/Reference-LAPACK/lapack/blob/f40d2208b99956aa7913aacb18ac4a39bacae3ad/SRC/zgelsy.f#L266
I think this should be ZTZRZF and ZUNMRZ instead.

jschueller:
The workspace NB queries in all four ?gelsy variants used ?GERQF and ?UNMRQ, but the routines actually called at runtime are ?TZRZF and ?UNMRZ (real) / ?UNMRZ (complex).

While ?GERQF and ?TZRZF are related (both produce an RQ-like factorization), their block sizes may differ in ILAENV's tuned values.  If the user relied on LWKOPT from the query to allocate exactly the right workspace, the actual ?TZRZF/?UNMRZ calls could underflow or overflow the work array.

Fix the ILAENV calls to match the routines actually invoked:
  ?GERQF -> ?TZRZF
  ?UNMRQ -> ?UNMRZ

Closes #676

### DIFF
--- a/SRC/cgelsy.f
+++ b/SRC/cgelsy.f
@@ -266,9 +266,9 @@
 *
       INFO = 0
       NB1 = ILAENV( 1, 'CGEQRF', ' ', M, N, -1, -1 )
-      NB2 = ILAENV( 1, 'CGERQF', ' ', M, N, -1, -1 )
+      NB2 = ILAENV( 1, 'CTZRZF', ' ', M, N, -1, -1 )
       NB3 = ILAENV( 1, 'CUNMQR', ' ', M, N, NRHS, -1 )
-      NB4 = ILAENV( 1, 'CUNMRQ', ' ', M, N, NRHS, -1 )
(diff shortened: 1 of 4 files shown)


martin-frbg (approved):

langou (approved):

martin-frbg:
Getting belated qualms here - it appears the "wrong" routines may have been used here precisely because ILAENV has no matching configuration for any `RZF` or `MRQ` and returning "1" for unblocked was not desirable. Can anybody comment please ?

langou:
> Getting belated qualms here - it appears the "wrong" routines may have been used here precisely because ILAENV has no matching configuration for any RZF or MRQ and returning "1" for unblocked was not desirable. Can anybody comment please ?

@martin-frbg: you are correct. See #1325

PR #1294, opened 2026-06-12
Fix integer overflow in LAPACKE nancheck helpers for large matrices

changangela:
Hit this while running Cholesky on a 47000x47000 matrix through OpenBLAS: `LAPACKE_dpotrf` segfaults in `LAPACKE_dtr_nancheck` before the factorization even starts, because `j*lda` overflows 32-bit `lapack_int` once `n = lda >= 46341`.

The `ge`/`gb`/`tp`/`tf` helpers already cast to `size_t`, but a few others were missed:

- `?tr_nancheck`: `a[i+j*lda]` (this is the potrf path, also used by the po/sy/he checks)
- `?_nancheck`: the `n*inc` loop bound overflows when called with a large stride (e.g. from the hs/tf checks)
- `?sp`/`?pp`/`?pf`/`?hp`/`?tp`/`?tf`: `len = n*(n+1)/2` overflows, the check then silently passes NaNs through
- `?tz_nancheck`: the part offsets overflow, widened to `int64_t` to keep the -1 sentinel

This bug also affects OpenBLAS (it vendors these files) - identical fix in OpenMathLib/OpenBLAS#5835.

Tested with UBSan at n = 47000: the old code reports signed integer overflow and either segfaults (tr, tz) or misses a planted NaN (vector, packed); the fixed code finds the NaN in all cases. Old and new give identical results on a sweep of small sizes covering all layouts/uplo/diag/shapes with a NaN at every position.

### DIFF
--- a/LAPACKE/utils/lapacke_ctr_nancheck.c
+++ b/LAPACKE/utils/lapacke_ctr_nancheck.c
@@ -68,14 +68,14 @@ lapack_logical API_SUFFIX(LAPACKE_ctr_nancheck)( int matrix_layout, char uplo, c
     if( ( colmaj || lower ) && !( colmaj && lower ) ) {
         for( j = st; j < n; j++ ) {
             for( i = 0; i < MIN( j+1-st, lda ); i++ ) {
-                if( LAPACK_CISNAN( a[i+j*lda] ) )
+                if( LAPACK_CISNAN( a[i+(size_t)j*lda] ) )
                     return (lapack_logical) 1;
             }
         }
     } else {
         for( j = 0; j < n-st; j++ ) {
             for( i = j+st; i < MIN( n, lda ); i++ ) {
-                if( LAPACK_CISNAN( a[i+j*lda] ) )
+                if( LAPACK_CISNAN( a[i+(size_t)j*lda] ) )
                     return (lapack_logical) 1;
             }
         }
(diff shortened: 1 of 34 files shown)


langou (approved):
This looks good to me.

LAPACKE/utils/lapacke_ctr_nancheck.c
```diff
@@ -68,14 +68,14 @@ lapack_logical API_SUFFIX(LAPACKE_ctr_nancheck)( int matrix_layout, char uplo, c
     if( ( colmaj || lower ) && !( colmaj && lower ) ) {
         for( j = st; j < n; j++ ) {
             for( i = 0; i < MIN( j+1-st, lda ); i++ ) {
-                if( LAPACK_CISNAN( a[i+j*lda] ) )
+                if( LAPACK_CISNAN( a[i+(size_t)j*lda] ) )
```
langou, line 71: The use of `size_t` here is certainly going to have a significant impact. Thanks!

LAPACKE/utils/lapacke_ctz_nancheck.c
```diff
@@ -103,27 +103,27 @@ lapack_logical API_SUFFIX(LAPACKE_ctz_nancheck)( int matrix_layout, char direct,
     }
 
     /* Initial offsets and sizes of triangular and rectangular parts */
-    lapack_int tri_offset = 0;
+    int64_t tri_offset = 0;
```
langou, line 106: And then using `int64_t` instead of `size_t` because we need to be signed (as opposed to unsigned) for this variable sounds good to me. Great work.

martin-frbg (approved):

PR #1295, opened 2026-06-13
*evr, *stevr: use fast CSTEMR path for partial eigenvalue ranges on I…

### LINKED ISSUE #1277: DSYEVR should call DSTEMR for RANGE="V" or "I"
DSYEVR only calls DSTEMR when ALL eigenvalues are requested. Otherwise, when a subset is requested, so with option RANGE="V" or "I", DSYEVR calls DSTEBZ/DSTEIN.

Suggestion: DSYEVR could call DSTEMR also when a subset is requested. (So with option RANGE="V" or "I".)

See
https://github.com/JuliaLang/LinearAlgebra.jl/pull/1493

See
https://github.com/Reference-LAPACK/lapack/blob/67c7c8ac05bddd39088fff0894bf2bfc20fcaae5/SRC/dsyevr.f#L580

jschueller:
…EEE machines

The *evr and *stevr routines now call CSTEMR/DSTEMR/SSTEMR with the user-specified RANGE argument on IEEE-754 compliant machines, instead of only using the fast path when the full spectrum is requested (ALLEIG or IL=1, IU=N). Partial eigenvalue ranges with eigenvectors now also benefit from the faster algorithm.

Fix a bug where WANTZ=.FALSE. with a partial range on IEEE machines would incorrectly set M=N and skip the DSTEBZ fallback, because INFO remained 0 from initialization when SSTERF was skipped (it can only compute all eigenvalues). The IF(INFO.EQ.0) success check is now moved inside each computational branch so it only triggers when SSTERF or CSTEMR was actually called and succeeded.

Closes #1277

### DIFF
--- a/SRC/cheevr.f
+++ b/SRC/cheevr.f
@@ -93,10 +93,9 @@
 *>   UC Berkeley, May 1997.
 *>
 *>
-*> Note 1 : CHEEVR calls CSTEMR when the full spectrum is requested
-*> on machines which conform to the ieee-754 floating point standard.
-*> CHEEVR calls SSTEBZ and CSTEIN on non-ieee machines and
-*> when partial spectrum requests are made.
+*> Note 1 : CHEEVR calls CSTEMR when possible (i.e., on machines
+*> which conform to the ieee-754 floating point standard). CHEEVR
+*> calls SSTEBZ and CSTEIN on non-ieee machines.
 *>
 *> Normal execution of CSTEMR may create NaNs and infinities and
 *> hence may abort due to a floating point exception in environments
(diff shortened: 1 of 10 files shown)


langou (approved):

langou:
Thanks @jschueller. Long overdue. 

This is not a criticism of this PR since this PR is going in the right direction, but, personally, I do not like the fact that HEEVR calls STERF when 
```fortran
IF( ALLEIG .OR. ( INDEIG .AND. IL.EQ.1 .AND. IU.EQ.N ) )
```

I think HEEVR should always call STEMR, no matter what.

If users want to use STERF in the case          
```fortran
IF( .NOT.WANTZ ) THEN
   IF( ALLEIG .OR. ( INDEIG .AND. IL.EQ.1 .AND. IU.EQ.N ) ) THEN
```
then they can call HEEV directly.

The way we do it now makes it hard for user to use STERF with
```fortran
IF( ALLEIG .OR. ( INDEIG .AND. IL.EQ.1 .AND. IU.EQ.N ) )
```

This can useful for example if a user wants bit-by-bit eigenvalue reproducibility whether ``WANTZ`` or not.

PR #1302, opened 2026-06-13
dstedc, sstedc: remove dead ICOMPZ.EQ.0 branch

### LINKED ISSUE #1143: ( nor urgent ) lines of code never called in STEDC
In DSTEDC, it seems that this lines of code are never called and so can safely been removed:
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dstedc.f#L439-L444

Lines 439 to 444 are never called since, at lines 306, IF( ICOMPZ.EQ.0 ) THEN we call STERF and GOTO 50
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dstedc.f#L306-L311
and, after anchor 50, STEDC ends
https://github.com/Reference-LAPACK/lapack/blob/efbd2fdf4439cdf5e11f9ff3ee9c65dd81164580/SRC/dstedc.f#L468-L476

It is not possible that ICOMPZ is 0 in lines 439 to 444, so the lines are never executed.

Thanks to @BrianDang03 for pointing this.

jschueller:
The IF( ICOMPZ.EQ.0 ) block calling DLASRT/SLASRT quick sort after the endwhile loop is unreachable -- ICOMPZ.EQ.0 already exits via CALL DSTERF/CALL SSTERF + GO TO 50 earlier. Remove the dead branch and its external declaration, keeping only the selection sort path.

Closes #1143

### DIFF
--- a/SRC/sstedc.f
+++ b/SRC/sstedc.f
@@ -213,7 +213,6 @@
 *     ..
 *     .. External Subroutines ..
       EXTERNAL           SGEMM, SLACPY, SLAED0, SLASCL, SLASET,
-     $                   SLASRT,
      $                   SSTEQR, SSTERF, SSWAP, XERBLA
 *     ..
 *     .. Intrinsic Functions ..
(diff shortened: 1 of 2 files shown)


martin-frbg (dismissed):

SRC/sstedc.f
```diff
@@ -438,34 +437,25 @@ SUBROUTINE SSTEDC( COMPZ, N, D, E, Z, LDZ, WORK, LWORK, IWORK,
...
-                 END IF
-   30         CONTINUE
-              IF( K.NE.I ) THEN
-                 D( K ) = D( I )
-                 D( I ) = P
-                 CALL SSWAP( N, Z( 1, I ), 1, Z( 1, K ), 1 )
-              END IF
-   40      CONTINUE
-         END IF
-      END IF
```
martin-frbg, line 468: think that second ENDIF doesn't belong to the elided block and needs to be kept ?

jschueller: yes, it does not build...

martin-frbg (approved):

langou:
Argh. I do not like it. 

Yes, this fixes a dead branch for the current code. 

But the current code reads
https://github.com/Reference-LAPACK/lapack/blob/693f114224c435e0f560c3988086e2ab232d37ab/SRC/dstedc.f#L297-L312

And so the idea is that someone could comment out
```fortran
      IF( ICOMPZ.EQ.0 ) THEN
         CALL DSTERF( N, D, E, INFO )
         GO TO 50
      END IF
```
so as to use the divide and conquer symmetric tridiagonal eigensolver to compute the eigenvalue only of the matrix instead of QR/QL symmetric tridiagonal eigensolver (STERF).

Why would someone want to do this? (1) maybe D&C is faster than STERF for eigenvalue only, (2) maybe the user wants to get the bit-by-bit same eigenvalue when calling STEDC with `COMPZ = 'N'` or `COMPZ = 'I'` or `COMPZ = 'V'`. 

Related, note the recent work of Ruiyi Zhan and Shaoshuai Zhan on accelerating D&C when only eigenvalues are desired: https://arxiv.org/abs/2605.26599

I'd almost advocate to comment out
```fortran
      IF( ICOMPZ.EQ.0 ) THEN
         CALL DSTERF( N, D, E, INFO )
         GO TO 50
      END IF
```
and have D&C stay true to D&C. Even if it is slower for eigenvalue computation, that's what it is. I want to say to STEDC "Be yourself! Be proud of you are! Do not try to be a QR algorithm". 

Note: It'd be interesting to see if this passes the test suite since D&C for eigenvalue only is not really tested currently.

All in all, I do not support this PR.

jschueller:
what do you suggest then ?

PR #1303, opened 2026-06-13
lsame: simplify using IACHAR

### LINKED ISSUE #701: Simplify lsame
Is their any interest in simplifying the [`lsame`](https://github.com/Reference-LAPACK/lapack/blob/master/INSTALL/lsame.f) function using F95 intrinsics?

```fortran
! lsame_example.f
! compile with: gfortran -Wall -std=f95 -o lsame_example lsame_example.f
      pure logical function lsame( ca, cb )
         implicit none
         character(len=1), intent(in) :: ca, cb

         integer, parameter :: diff_case = iachar('a')-iachar('A')
         integer :: ia, ib

         lsame = ca == cb
         if (lsame) return 

         ! case-insensitive test
         ia = iachar(ca)
         ib = iachar(cb)

         ! Capital alphabetic characters are between 65 and 90
         ! in the ASCII character encoding
         if ( 64 < ia .and. ia < 91) ia = ia + diff_case
         if ( 64 < ib .and. ib < 91) ib = ib + diff_case

         lsame = ia == ib

      end

      program main

         logical :: lsame
         external :: lsame

         print *, lsame('a','b')  ! F
         print *, lsame('z','z')  ! T

         print *, lsame('R','r')  ! T
         print *, lsame('t','U')  ! F

      end
```

The gfortran produced assembly is the same for the version [shown above](https://godbolt.org/z/9MroGszK8), and the [current version](https://godbolt.org/z/ze1oEM7G1). (The file above is valid both as free-from and fixed-form.)

Potential issues:
* old compilers and platforms not supporting F95 features including `iachar`, intent and pure attributes, modern relational operators, etc.

jschueller:
Replace ICHAR with the F95 intrinsic IACHAR, which always returns ASCII codes regardless of the platform's native character encoding. This eliminates the entire EBCDIC/Prime machine detection branch (logic testing ZCODE against ASCII/EBCDIC/Prime codes), producing shorter, simpler, and equivalent code.

Closes #701

### DIFF
--- a/BLAS/SRC/lsame.f
+++ b/BLAS/SRC/lsame.f
@@ -63,57 +63,22 @@
 * =====================================================================
 *
 *     .. Intrinsic Functions ..
-      INTRINSIC ICHAR
+      INTRINSIC IACHAR
 *     ..
 *     .. Local Scalars ..
-      INTEGER INTA,INTB,ZCODE
+      INTEGER INTA,INTB
 *     ..
-*
-*     Test if the characters are equal
 *
       LSAME = CA .EQ. CB
       IF (LSAME) RETURN
 *
-*     Now test for equivalence if both characters are alphabetic.
+*     Use IACHAR to get ASCII codes for case-insensitive comparison,
+*     regardless of the platform's native character encoding.
 *
-      ZCODE = ICHAR('Z')
-*
-*     Use 'Z' rather than 'A' so that ASCII can be detected on Prime
-*     machines, on which ICHAR returns a value with bit 8 set.
-*     ICHAR('A') on Prime machines returns 193 which is the same as
-*     ICHAR('A') on an EBCDIC machine.
-*
-      INTA = ICHAR(CA)
-      INTB = ICHAR(CB)
-*
-      IF (ZCODE.EQ.90 .OR. ZCODE.EQ.122) THEN
-*
-*        ASCII is assumed - ZCODE is the ASCII code of either lower or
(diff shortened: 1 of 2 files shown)


ACSimon33:
We might want to change the file extension to `.f95` to make the required standard explicit. We did that for some new routines with `.f90`.

jschueller:
ah, yes, I did not get that, are we allowed to use f95 in lapack though ?

langou:
(1) I'd be OK to allow F95
(2) if we do F95, it would be good practice to indicate it in the file name as suggested by @ACSimon33 
(3) in the past, I would have argued to move the current lsame.f to the "deprecated" folder, but I guess that's why we have source control so I think changing lsame.f and not keeping an older version is just fine. 

More comments on the wisdom to allow F95 welcome.

imciner2:
> We might want to change the file extension to .f95 to make the required standard explicit. We did that for some new routines with .f90.

Please don't use `.f95` extensions to mean Fortran 95 code, keep the separation `.f` for fixed form and `.f90` for free-form. Some compilers (such as the Intel one), don't recognize any file extension other than `.f` or `.f90` without special command line options (https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2026-0/file-extensions.html), and the usual advice has been to not use any other extensions (https://stevelionel.com/drfortran/2013/01/11/doctor-fortran-in-source-form-just-wants-to-be-free/).

Issue #1231, opened 2026-04-11
TESTING/EIG/csd.in: please add M=P=Q=1 (and small degenerate cases) — exposes a [CZ]UNCSD2BY1 bug already fixed between 3.12.1 and current


nakatamaho:
## Summary

The shipped `TESTING/EIG/csd.in` does not exercise the trivial `M=P=Q=1` corner case. Adding it makes `xeigtstz` on **LAPACK 3.12.1** report 6 failing test ratios on `M=55, P=40, Q=20` (2-by-1 CSD, tests 10 and 11). The same modified input file passes cleanly on **current master**, and also passed on **3.9.1**. So a bug was introduced after 3.9.1, was present in 3.12.1, and has since been fixed somewhere on master — but because the released test data never covered this seed sequence, the regression slipped through release testing for multiple versions.

I would like to request that the corresponding small/degenerate cases be added to `csd.in` upstream so this class of regression is caught by `make check` in future releases.

## Reproducer

Diff against shipped `TESTING/EIG/csd.in`:

```diff
-10                                   Number of values of M, P, Q
-0  10 10 10 10 21 24 30 22 32 55     Values of M
-0  4   4 0  10 9  10 20 12 12 40     Values of P
-0  0  10 4  4  15 12 8  20 8  20     Values of Q
+12                                   Number of values of M, P, Q
+0 1  10 10 10 10 21 24 30 22 32 55     Values of M
+0 1  4   4 0  10 9  10 20 12 12 40     Values of P
+0 1  0  10 4  4  15 12 8  20 8  20     Values of Q
```

Run `./xeigtstz < csd.in`.

### LAPACK 3.12.1 (FAIL)

```
M= 55 P= 40, Q= 20, type 1, test 10, ratio= 0.152394E+15
M= 55 P= 40, Q= 20, type 1, test 11, ratio= 0.496574E+15
M= 55 P= 40, Q= 20, type 2, test 10, ratio= 0.324446E+12
M= 55 P= 40, Q= 20, type 2, test 11, ratio= 0.466353E+12
M= 55 P= 40, Q= 20, type 3, test 10, ratio= 0.443792E+15
M= 55 P= 40, Q= 20, type 3, test 11, ratio= 0.663765E+15
CSD:    6 out of  720 tests failed to pass the threshold
```

### LAPACK current master (PASS)

```
All tests for CSD routines passed the threshold (    720 tests run)
```

### LAPACK 3.9.1

[...]

PR #1304, opened 2026-06-13
{s/c/d/z}geqrf: document workspace query for empty matrices

jschueller:
Clarify that the workspace query returns a valid LWORK (>= 1) even when MIN(M,N) = 0. The code already handles this correctly; this is purely a documentation update.

see #257

### DIFF
--- a/SRC/cgeqrf.f
+++ b/SRC/cgeqrf.f
@@ -102,6 +102,8 @@
 *>          only calculates the optimal size of the WORK array, returns
 *>          this value as the first entry of the WORK array, and no error
 *>          message related to LWORK is issued by XERBLA.
+*>          For an empty matrix (MIN(M,N) = 0), the workspace query
+*>          returns LWORK >= 1.
 *> \endverbatim
 *>
 *> \param[out] INFO
--- a/SRC/dgeqrf.f
+++ b/SRC/dgeqrf.f
@@ -102,6 +102,8 @@
 *>          only calculates the optimal size of the WORK array, returns
 *>          this value as the first entry of the WORK array, and no error
 *>          message related to LWORK is issued by XERBLA.
+*>          For an empty matrix (MIN(M,N) = 0), the workspace query
+*>          returns LWORK >= 1.
 *> \endverbatim
 *>
 *> \param[out] INFO
--- a/SRC/sgeqrf.f
+++ b/SRC/sgeqrf.f
@@ -102,6 +102,8 @@
 *>          only calculates the optimal size of the WORK array, returns
 *>          this value as the first entry of the WORK array, and no error
 *>          message related to LWORK is issued by XERBLA.
+*>          For an empty matrix (MIN(M,N) = 0), the workspace query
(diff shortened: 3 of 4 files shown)


martin-frbg:
If I read the later comments on the linked issue correctly, the intention was to keep it open until everybody is convinced that there are no similar cases (of returning an invalid workspace size when there's no work to do) left in other functions ?

jschueller:
fair enough, lets leave it open

martin-frbg (approved):

langou (approved):

langou:
I went too fast.

This sentence is clear enough for me
https://github.com/Reference-LAPACK/lapack/blob/36c931cac2259d4839215404189f2004dd48de98/SRC/dgeqrf.f#L97

It says that **if** `MIN(M,N) = 0`, **then** `LWORK >= 1` **else** `LWORK >= N`

So, then, the statement
https://github.com/Reference-LAPACK/lapack/blob/36c931cac2259d4839215404189f2004dd48de98/SRC/dgeqrf.f#L105-L106
feels redundant.

We can revise the sentence though if some feel it is confusing. Right now, we are redundant, I prefer revert and do again.

If we do again, the statement "the workspace query returns `LWORK >= 1`." is not good. `LWORK` is "input only", what is returned is `WORK(1)`. So we need to be careful with this.

Also, since we are at it, we might as well impose `WORK(1) = 1` as a convention for `MIN(M,N) = 0`. (The `>=` is weird.) 

And, also, since we are at it, maybe we move the quick return `IF` statement 
https://github.com/Reference-LAPACK/lapack/blob/36c931cac2259d4839215404189f2004dd48de98/SRC/dgeqrf.f#L206-L212
before the workspace computation and just after
https://github.com/Reference-LAPACK/lapack/blob/36c931cac2259d4839215404189f2004dd48de98/SRC/dgeqrf.f#L181
That'd be cleaner. No need to launch all this workspace computation for an empty matrix. We might as well bail out early.

langou:
I looked at the code again. There were some issues in my last suggestions. 

Suggestion 1. About redundancy. I was not being correct. The comment before this PR was about valid values for LWORK. The comment added by the PR is about what the workspace query returns. So I'll amend suggestion 1 in moving @jschueller wording from the LWORK section to the WORK section.

Suggestion 2. My second suggestion seems good. We do not return LWORK. LWORK is input only, so that needs to be changed.

Suggestion 3. Before doing a "quick return" if one of the two dimensions is zero, we want to test the input argument. All in all, the code structure is good. 

All in all, I'll suggest a PR.

jschueller:
see #1321

langou:
Thanks @jschueller. We criss-crossed. See: 
https://github.com/Reference-LAPACK/lapack/pull/1304#issuecomment-4924910383
Can you amend #1231 based on this comment?

langou:
Or do you want me to suggest a PR? Either is fine with me.

jschueller:
yes, go ahead, I dropped #1321

PR #1307, opened 2026-06-13
DLASQ3, SLASQ3: stop modifying QMAX

### LINKED ISSUE #326: xLASQ3: argument qmax labeled 'in' but modified
The argument qmax is labeled as an 'in' parameter in the doc, but it is potentially changed (line 320), and used in xLASQ2 after the call to xLAQS3.
I believe it should be relabeled as 'in/out'
Also, let me know if I'm being overly pedantic.

jschueller:
QMAX was documented as [in] but was written inside the qd-array reversal block. The caller recomputes QMAX immediately after the call, so the update has no useful effect. Remove it to match the documented intent

Closes #326

### DIFF
--- a/SRC/dlasq3.f
+++ b/SRC/dlasq3.f
@@ -89,7 +89,7 @@
 *>         Lower order part of SIGMA
 *> \endverbatim
 *>
-*> \param[in] QMAX
+*> \param[in,out] QMAX
 *> \verbatim
 *>          QMAX is DOUBLE PRECISION
 *>         Maximum value of q.
--- a/SRC/slasq3.f
+++ b/SRC/slasq3.f
@@ -89,7 +89,7 @@
 *>         Lower order part of SIGMA
 *> \endverbatim
 *>
-*> \param[in] QMAX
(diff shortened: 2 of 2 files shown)


langou:
Thanks @jschueller. 

I agree with Issue #326. 

That being said, I do not think this PR is the good fix. 

A better fix could be to change QMAX from [in] in DLASQ3 to [in/out]. 

I disagree with the statement: "The caller recomputes QMAX immediately after the call, so the update has no useful effect." (0) It is correct that `QMAX` gets recomputed whenever a split is found but this is not all the time. (1) QMAX is used in DLASQ3 as in/out and we keep on building up the value as we go, (2) also, it is used in in the `IF` statement `Z( 4*N0 ).LE.TOL2*QMAX` in DLASQ2. 

All in all, I think this is PR needs to be changed to QMAX being flagged as [in/out] in DLASQ3. I agree that this would fix Issue #326.

jschueller:
ok, done

langou (approved):

PR #1310, opened 2026-06-15
sblas3: Exclude from double precision build

jschueller:
When only `BUILD_DOUBLE` is enabled (without `BUILD_SINGLE`), the single-precision Level 3 BLAS routines (sgemm, ssymm, ssyrk, strsm, etc.) are still compiled into the BLAS library. This appears to be a copy-paste error — `SBLAS3` should not be in the `BUILD_DOUBLE` section. While the object files may be ignored at link time, they bloat the build and may cause unexpected symbol exports.


ACSimon33:
I think we need the single precision BLAS routines because if we build LAPACK with only double precision, these single precision routines are also compiled (SRC/CMakeLists.txt:166):

```
set(DSLASRC
    sgetrf.f sgetrf2.f sgetrs.f sisnan.f slaisnan.f slaswp.f spotrf.f spotrf2.f
    spotrs.f)
```

Those are needed for the mixed precision routines, like `dsposv`.

We could build those only if both single and double precision are enabled. Or we introduce a  new CMake variable `BUILD_MIXED_PRECISION` ...

Issue #729, opened 2022-10-13
Correctness and efficiency of LAPACKE_xLACPY when matrix_layout is LAPACK_ROW_MAJOR


digikar99:
**Description**

Going by the discussion [here](https://www.reddit.com/r/lisp/comments/y24hel/lapacke_slacpy_works_correctly_in_c_but_lapack/) (which I'm summarizing below), it seems that lapacke_slacpy and friends are (i) copying more than twice as much when asked to copy only the lower or upper triangular parts (ii) copying over the uninitialized malloc-ed region back to `b`.

https://github.com/Reference-LAPACK/lapack/blob/28f7e8309608b92aaec2e2556d4b25d758ccada9/LAPACKE/src/lapacke_slacpy_work.c#L70-L76

This results in the following code working as expected while using `LAPACK_COL_MAJOR` but not working as expected while using `LAPACK_ROW_MAJOR`.

Code, compile using `gcc test-lacpy.c -llapacke -o test-lacpy`:

```c
#include <lapacke.h>
#include <stdio.h>

#define size 9

int main(){
  float a[size] = {1,2,3,4,5,6,7,8,9};
  float b[size];

  for(int i=0; i<size; i++) b[i] = 100;

  // Works as expected with LAPACK_COL_MAJOR but not with LAPACK_ROW_MAJOR
  LAPACKE_slacpy(LAPACK_COL_MAJOR, 'U', 3, 3, a, 3, b, 3);

  for(int i=0; i<3; i++){
    for(int j=0; j<3; j++){
      printf("%f ", b[i*3+j]);
    }
    printf("\n");
  }

  return 0;
}
```

When compiled with `LAPACK_COL_MAJOR`, this produces the output:

```
1.000000 100.000000 100.000000 
4.000000 5.000000 100.000000 
7.000000 8.000000 9.000000 
```

But when compiled with `LAPACK_ROW_MAJOR`, this produces the unexpected output (expectation is 100 instead of 0):

```
1.000000 2.000000 3.000000 
0.000000 5.000000 6.000000 
0.000000 0.000000 9.000000
```

Might a native C solution be preferable for this particular function/family?

**Checklist**

- [X] I've included a minimal example to reproduce the issue
- [X] I'd be willing to make a PR to solve this issue

angsch:
That's an interesting corner case. Strictly speaking, the code is correct
[...]

PR #1318, opened 2026-07-06
Fix LAPACKE_?lacpy_work row-major triangular copy corrupting the untouched triangle (#729)

meng004:
### Problem

`LAPACKE_?lacpy_work` (`s`/`d`/`c`/`z`) with `matrix_layout == LAPACK_ROW_MAJOR`
and a triangular `uplo` (`'U'`/`'L'`) corrupts the part of the destination `B`
that the operation is supposed to leave untouched, and reads uninitialized heap
memory in the process. As reported in #729, only the requested triangle of `B`
should change; instead the complementary region of `B` is overwritten with
garbage. The column-major path is correct.

### Root cause

`LAPACKE/src/lapacke_slacpy_work.c` (and the `d`/`c`/`z` siblings) implement the
row-major branch by transposing through temporaries:

```c
    ...
    a_t = LAPACKE_malloc(...);   /* full m*n temporary */
    b_t = LAPACKE_malloc(...);   /* full m*n temporary, NOT initialized */
    LAPACKE_sge_trans( matrix_layout, m, n, a, lda, a_t, lda_t );
    LAPACK_slacpy( &uplo, &m, &n, a_t, &lda_t, b_t, &ldb_t ); /* writes only the triangle of b_t */
    LAPACKE_sge_trans( LAPACK_COL_MAJOR, m, n, b_t, ldb_t, b, ldb ); /* transposes the WHOLE b_t back */
    ...
}
```

The kernel `LAPACK_?lacpy` with a triangular `uplo` writes **only** the requested
triangle of `b_t`, so the complementary region of `b_t` stays uninitialized. The
final `LAPACKE_?ge_trans` then transposes the **entire** `m`-by-`n` `b_t` back
into the caller's `B` — including that never-written region — overwriting the
bytes of `B` that `uplo` requires to be preserved with uninitialized heap
contents. (In the file as shipped this is
`LAPACKE/src/lapacke_slacpy_work.c:76`, the "Transpose output matrices" call.)

### Fix

A row-major `m`-by-`n` matrix with leading dimension `lda` is bit-identical to a
column-major `n`-by-`m` matrix — its mathematical transpose. Copying triangle
`uplo` of the row-major matrix is therefore exactly copying the **opposite**
triangle (`U`↔`L`) of that transpose, with the `m`/`n` extents swapped. So the
row-major branch can call the Fortran kernel directly on `a`/`b` with `uplo`
flipped and `m`/`n` swapped:

```c
    char uplo_t = uplo;
    /* leading-dimension checks unchanged */
    if( uplo == 'U' || uplo == 'u' ) uplo_t = 'L';
    else if( uplo == 'L' || uplo == 'l' ) uplo_t = 'U';
    LAPACK_slacpy( &uplo_t, &n, &m, a, &lda, b, &ldb );
    info = 0;
}
```

Like the already-correct column-major path, the kernel now writes exactly the
requested triangle of `B` and never touches the complementary region. This is
minimal and, in fact, net-negative in size: it removes the two `LAPACKE_malloc`s,
the two `LAPACKE_?ge_trans` calls, and the error-exit ladder rather than adding
special cases. The existing row-major leading-dimension checks (`lda < n` →
`-6`, `ldb < n` → `-8`) are preserved verbatim, the column-major and
invalid-layout branches are untouched, the public signatures/headers are
unchanged, and `uplo` values other than `'U'`/`'L'` (e.g. `'A'`, a full copy)
are layout-agnostic and pass straight through.

### Test evidence

Validated with a two-arm build in which both arms share the identical
`liblapack`/`libblas`; only the four LAPACKE `*lacpy_work` objects differ
(pristine vs. patched). The oracle copies triangle `uplo` of the same logical
matrix `A` into a sentinel-filled `B` once in `LAPACK_COL_MAJOR` (ground truth)
and once in `LAPACK_ROW_MAJOR`; the two logical results must be identical.

| check | before (pristine) | after (patched) |
|---|---|---|
| row-vs-col logical result, `uplo='U'` | mismatch (maxdiff 9.0) | identical (0.0) |
| row-vs-col logical result, `uplo='L'` | mismatch (maxdiff 29.0) | identical (0.0) |
| copied triangle vs. `A` | correct (0.0) | correct (0.0) |
| complementary region vs. sentinel | **corrupted** (9.0 / 29.0) | preserved (0.0) |
| full-copy control `uplo='A'` (row vs col) | identical | identical |
| rectangular M=3,N=5, `uplo='U'`/`'L'` | mismatch (7.0 / 22.0) | correct |

The copied-triangle content was already correct before the fix; the sole defect
is the complementary region, which the fix restores to being preserved. The
`uplo='A'` full-copy control and the column-major reference are identical on both
arms, confirming the change does not perturb the paths it should not.

### How to reproduce

Minimal C reproducer (double precision; the same holds for `s`/`c`/`z`):

```c
#include <lapacke.h>
#include <stdio.h>
int main(void) {
    const int n = 5;
    double A[25], Bcol[25], Brow[25];
    for (int i = 0; i < 25; i++) A[i] = i + 1;      /* same logical A */
    for (int i = 0; i < 25; i++) Bcol[i] = Brow[i] = -9.0; /* sentinel */

    /* col-major: copy upper triangle of A into B (ground truth) */
    LAPACKE_dlacpy_work(LAPACK_COL_MAJOR, 'U', n, n, A, n, Bcol, n);
    /* row-major: same logical operation on the same logical A */
    LAPACKE_dlacpy_work(LAPACK_ROW_MAJOR, 'U', n, n, A, n, Brow, n);

    /* Brow, read row-major, must equal Bcol read col-major, element for element.
       Before the fix the strictly-lower part of Brow is garbage (not -9). */
    for (int i = 0; i < n; i++)
        for (int j = 0; j < n; j++)
            if (Brow[i*n + j] != Bcol[j*n + i])
                printf("mismatch at (%d,%d): row=%g col=%g\n",
                       i, j, Brow[i*n + j], Bcol[j*n + i]);
    return 0;
}
```

Build with `-llapacke -llapack -lblas`. Before the fix it prints mismatches in
the complementary (strictly-lower) triangle; after the fix it prints nothing.

---

### DIFF
--- a/LAPACKE/src/lapacke_clacpy_work.c
+++ b/LAPACKE/src/lapacke_clacpy_work.c
@@ -42,10 +42,7 @@ lapack_int API_SUFFIX(LAPACKE_clacpy_work)( int matrix_layout, char uplo, lapack
         /* Call LAPACK function and adjust info */
         LAPACK_clacpy( &uplo, &m, &n, a, &lda, b, &ldb );
     } else if( matrix_layout == LAPACK_ROW_MAJOR ) {
-        lapack_int lda_t = MAX(1,m);
-        lapack_int ldb_t = MAX(1,m);
-        lapack_complex_float* a_t = NULL;
-        lapack_complex_float* b_t = NULL;
+        char uplo_t = uplo;
         /* Check leading dimension(s) */
         if( lda < n ) {
             info = -6;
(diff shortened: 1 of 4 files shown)


LAPACKE/src/lapacke_clacpy_work.c
```diff
@@ -57,34 +54,27 @@ lapack_int API_SUFFIX(LAPACKE_clacpy_work)( int matrix_layout, char uplo, lapack
...
-        a_t = (lapack_complex_float*)
-            LAPACKE_malloc( sizeof(lapack_complex_float) * lda_t * MAX(1,n) );
-        if( a_t == NULL ) {
-            info = LAPACK_TRANSPOSE_MEMORY_ERROR;
-            goto exit_level_0;
+        /* Fix (issue #729): the previous code transposed the FULL m-by-n
+         * matrix through temporary buffers, which (a) read the uninitialized
+         * complementary region of the destination temporary and (b) wrote it
+         * back over the part of B that `uplo` requires to stay untouched,
+         * corrupting the caller's data and reading uninitialized memory. A
```
mgates3, line 61: I would put comments like this in the PR description, not in the code itself. It clutters up the code -- years from now, the old behavior is irrelevant.

Same comment in other precisions.

mgates3 (commented):
Other than perhaps shortening the comments, this looks good. Nice that it greatly simplified the code and eliminated malloc and copy! (Though I didn't personally validate the code, and LAPACKE unfortunately lacks testers. LAPACK++ tests only column-major.)

langou:
I agree with @mgates3 that this is a great patch. (1) This is an elegant way to handle the row major triangular matrices and, (2) in addition, there was a bug so that fixes it, great.

Also, my preference (similar to @mgates') would be to leave out the lengthy comment.

langou:
Let us do it in two steps. I'll merge and submit a PR to remove the comments.

langou (approved):

PR #1336, opened 2026-07-26
Modernize lapack_testing.py script

ACSimon33:
### Shortcomings of the original script

- **Silently ignored test results.** Only 74 hard-coded `.out` files were analyzed:
  - `*_64.out` extended API results (`BUILD_INDEX64_EXT_API=ON`, the CMake default)
    were never read — in an extended-API-only build the summary analyzed **zero**
    files and still reported success; (to be fair, I just added those tests recently ...)
  - `{s,d,c,z}dmd.out` were excluded by an off-by-one error in the test tables, and
    `-t dmd` crashed with an `IndexError`;
  - `{s,d,c,z}{bal,bak,gbal,gbak}.out` were never read;
  - the `Error in XTRSYL:` failure records of the eigencondition checkers matched no
    parsing heuristic, so the failures were invisible even in analyzed files.
- **Incorrect accounting.** Run counts were only taken from `( N tests run)` lines,
  which `alasum.f`/`alasvm.f` print only when a test set has zero failures — failing
  sets were missing from the denominator, so error percentages were wrong precisely
  when there were failures.
- **Unusable for automation.** The exit code never reflected test results, and in
  `-n` mode a directory with no results printed `0 / 0` — indistinguishable from a
  fully passing suite. Tests added to the harness but not to the script were
  silently ignored (exactly how the gaps above went unnoticed).

### What the new script does

- Typed Python 3 (>= 3.7, standard library only) using `argparse` and dataclasses.
  All existing options keep their meaning, and the three in-tree invocations
  (`make lapack_testing`, `LAPACK_Test_Summary`, `CTEST_CUSTOM_POST_TEST`) work
  unchanged.
- Collects everything the harnesses produce: DMD, balancing/backtransformation and
  eigencondition failure records, with parsers derived from the actual Fortran `FORMAT`
  statements. Extended API (`*_64.out`) results are auto-detected and reported in a
  separate summary section so default-API totals stay comparable.
- Correct run accounting: failing summary lines (`N out of M tests failed ...`)
  contribute M to the run count.
- Detects future coverage gaps: `.out` files matching no known test are reported on
  stderr, so a new test that isn't wired into the script can no longer go unnoticed.
- Opt-in gating: `--fail-on-error`, `--fail-if-empty` and
  `--fail-on-unrecognized`. The `LAPACK_Test_Summary` CTest and the `make lapack_testing` test now pass `--fail-if-empty --fail-on-unrecognized`, so a build whose test suite produced no output — or
  produced output the summary cannot analyze — can no longer report a passing
  summary.
- Robust run mode: probes CMake and Makefile binary locations, no shell, writes via
  a temporary file so a failed launch cannot clobber earlier results, and reports
  launch failures without aborting the remaining tests.

### Behavior changes

- Default summary totals grow slightly (DMD and balancing tests, ~40-50 runs per
  precision).
- A non-finite `largest test error` in the balancing output counts as a numerical
  failure (e.g. `sgebal` currently reports `Infinity` on some platforms —
  previously invisible).
- Invalid `-p`/`-t` values exit 2 with a usage message instead of silently meaning
  "all".
- `LAPACK_Test_Summary` and `make lapack_testing` fail when zero test results were
  analyzed or when unrecognized `.out` files are present in the testing directory.

### Validation

Old and new scripts were compared against complete CMake build outputs (188 `.out`
files including `_64` variants); every count difference is attributable to a listed
fix. All option combinations were compared against the old script, and run mode was
exercised end-to-end including driver-launch failure paths.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ cblas_testing: cblaslib blaslib
 lapack_testing: tmglib lapacklib blaslib
 	$(MAKE) -C TESTING/LIN cleanexe
 	$(MAKE) -C TESTING
-	./lapack_testing.py
+	./lapack_testing.py --fail-if-empty --fail-on-unrecognized
 
 .PHONY: variants_testing
 variants_testing: tmglib variants lapacklib blaslib
--- a/TESTING/CMakeLists.txt
+++ b/TESTING/CMakeLists.txt
@@ -21,7 +21,7 @@ if(Python3_EXECUTABLE AND LAPACK_TESTING_USE_PYTHON)
   add_test(
     NAME LAPACK_Test_Summary
     WORKING_DIRECTORY ${LAPACK_BINARY_DIR}
-    COMMAND ${Python3_EXECUTABLE} "lapack_testing.py"
+    COMMAND ${Python3_EXECUTABLE} "lapack_testing.py" "--fail-if-empty" "--fail-on-unrecognized"
     )
 endif()
 
--- a/lapack_testing.py
+++ b/lapack_testing.py
@@ -81,6 +81,20 @@ EIG_BALANCE_SETS: "Tuple[Tuple[str, str], ...]" = (
     ("gbak", "Generalized Balancing Backtransformation"),
 )
 
+# Linear equation test sets: (family, name stem, executable prefix,
+# description).  Names and executables are built from the precision
+# letter plus the stem, e.g. stest.in/stest.out/xlintsts; the mixed
+# precision sets use the letter of both precisions (dstest, xlintstds)
+# and exist only for the precisions in MIXED_PARTNER.
+LIN_SETS: "Tuple[Tuple[str, str, str, str], ...]" = (
+    ("lin", "test", "xlintst", "Linear Equation routines"),
+    ("mixed", "test", "xlintst", "Mixed Precision linear equation routines"),
+    ("rfp", "test_rfp", "xlintstrf", "RFP linear equation routines"),
+)
+
+# All test families, in reporting order per precision.
+ALL_FAMILIES: "Tuple[str, ...]" = ("eig",) + tuple(s[0] for s in LIN_SETS) + ("dmd",)
+
 # API suffixes that may exist: default API and index-64 extended API.
 KNOWN_SUFFIXES: "Tuple[str, ...]" = ("", "_64")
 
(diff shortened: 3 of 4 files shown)


langou (approved):

langou:
Thanks. This is great.

I have one question that is related, maybe someone following the repository can answer. 

We have a CI, and if we look at an example of one of the output:
https://github.com/Reference-LAPACK/lapack/actions/runs/30222442487/job/89847048381
We see (starting line 515)
```
                        -->  LAPACK TESTING SUMMARY  <--
                   Processing LAPACK Testing output found in:
                 /Users/runner/work/lapack/lapack/build/TESTING

 -------------------------------- Default API --------------------------------- 

  SUMMARY                nb test run      numerical error          other error
  ==================   =============   ==================   ==================
  REAL                       1630125           1   (0.0%)           0   (0.0%)
  DOUBLE PRECISION           1630921           4   (0.0%)           4   (0.0%)
  COMPLEX                    1089881           0   (0.0%)           0   (0.0%)
  COMPLEX16                  1090948           0   (0.0%)           0   (0.0%)

  --> ALL PRECISIONS         5441875           5   (0.0%)           4   (0.0%)

 ----------------------------- Extended API (_64) ----------------------------- 

  SUMMARY                nb test run      numerical error          other error
  ==================   =============   ==================   ==================
  REAL                       1630125           1   (0.0%)           0   (0.0%)
  DOUBLE PRECISION           1630921           4   (0.0%)           4   (0.0%)
  COMPLEX                    1089881           0   (0.0%)           0   (0.0%)
  COMPLEX16                  1090948           0   (0.0%)           0   (0.0%)

  --> ALL PRECISIONS         5441875           5   (0.0%)           4   (0.0%)
```

I do not understand how many errors the CI tolerate for a "green" checkmark, ✅. We run 5,441,875 tests and, for this machine, compiler flag, etc. we have 5 numerical errors, and 4 other errors. Is the number of errors even checked by the CI when it gives a green checkmark? Maybe the CI only checks that the code compiles and executes and return 0 or something like this. I do not understand this. If someone can explain, that'd be great.

ACSimon33:
> I do not understand how many errors the CI tolerate for a "green" checkmark, ✅. We run 5,441,875 tests and, for this machine, compiler flag, etc. we have 5 numerical errors, and 4 other errors. Is the number of errors even checked by the CI when it gives a green checkmark? Maybe the CI only checks that the code compiles and executes and return 0 or something like this. I do not understand this. If someone can explain, that'd be great.

As far as I can tell, the test programs themselves only fail if the inputs are invalid, if segfaults occur, or if anything crashes because we compile with `-fcheck=all` (out-of-bounds, bad allocation, ...). The test summary never failed before; even if all tests failed, the pipeline would be green. I left it that way for now. We can enable fail on numerical and other errors with `--fail-on-error` in the new script. My plan was to fix the current numerical issues and eventually enable it in the pipeline.

langou:
> My plan was to fix the current numerical issues and eventually enable it in the pipeline.

That would be fantastic. 

> A non-finite largest test error in the balancing output counts as a numerical failure (e.g. sgebal currently reports Infinity on some platforms — previously invisible).

Yes, so that test should be removed from the test suite. So we should have some tests for GEBAL, but we should remove this specific test. 

For some numerical errors, I think it would be fine to increase our threshold to say 100. I have a couple of numerical tests failing on my machine at 50+ or so. So 60 would do, but we might as well put some wiggle room. 64? 100? Maybe 64?

Then the testing should pass all tests on my machine. 

We would have to check all machines used by the CI to see what the numerical errors are. 

Is there a way to see the 'testing_results.txt' file from the CI? (Or all the .out files for the matter.)

PR #1341, opened 2026-07-30
CBLAS: count tests and report CBLAS results in the test summary

ACSimon33:
## Summary

The CBLAS test drivers now report per-routine test counts, and
`lapack_testing.py` analyzes the CBLAS output as its own suite alongside
LAPACK and BLAS. See #1339.

## Driver changes

Counts are reported in the same shape as the BLAS drivers, tagged with the
data layout so the column-major and row-major passes stay distinguishable;
existing output lines are unchanged:

```
cblas_sgemv      ROW-MAJOR    COMPUTATIONAL TESTS:  3456 RUN,     0 FAILED
cblas_sgemv      ERROR-EXIT TESTS:       13 RUN,        0 FAILED
```

- Level 2/3 drivers count computational tests per routine; the paths that set
  `FATAL` bump the failure count too, so an aborting routine cannot report
  zero failures.
- The complex Level 2 drivers (`c_cblat2.f`, `c_zblat2.f`) now tag their
  verdicts with the layout, like all the other Level 2/3 drivers already did —
  previously they printed two textually identical lines per routine.
- The error-exit checkers count each `chkxer()` call. A new `cblas_xbad` flag
  catches the case where `xerbla` is reached with the wrong routine name
  (which the extended API build hits), so the counts agree with the verdict.
- The Level 1 drivers count one test per element comparison, matching their
  BLAS counterparts.

## lapack_testing.py

CBLAS becomes a third library section in the summary, with a test case per
precision and BLAS level and a `--cblas-dir` option (default `CBLAS/testing`;
a missing directory is skipped silently since CBLAS is off by default).
Computational failures count as numerical errors, error-exit failures as
other errors. The parser is shared with BLAS; only the layout tags and test
tables differ. Extended API (`*_64.out`) outputs are reported in their own
sections.

## Other changes

- `--merge-apis`: collapses a library's two API sections into one table when
  they report the same errors per output file. Used by the CTest and Makefile
  test summaries.
- `--junit-xml PATH`: writes a JUnit XML report of the analyzed output files
  (one testcase per file), e.g. for GitLab CI `artifacts:reports:junit`. The
  CMake `LAPACK_Test_Summary` test and `make lapack_testing` write
  `lapack_testing_junit.xml` by default.


langou (approved):

langou:
That's very nice, so now it looks like on my machine
```
                        -->  LAPACK TESTING SUMMARY  <--
                   Processing LAPACK Testing output found in:
       /Users/langou/Documents/repositories/lapack__reference.git/TESTING

 ---------------------------- LAPACK: Default API ----------------------------- 

  SUMMARY                nb test run      numerical error          other error
  ==================   =============   ==================   ==================
  REAL                       1630125           1   (0.0%)           0   (0.0%)
  DOUBLE PRECISION           1630947           0   (0.0%)           0   (0.0%)
  COMPLEX                    1089881           2   (0.0%)           0   (0.0%)
  COMPLEX16                  1090948           0   (0.0%)           0   (0.0%)

  --> ALL PRECISIONS         5441901           3   (0.0%)           0   (0.0%)

 ----------------------------- BLAS: Default API ------------------------------ 

  SUMMARY                nb test run      numerical error          other error
  ==================   =============   ==================   ==================
  REAL                         81216           6   (0.0%)           0   (0.0%)
  DOUBLE PRECISION             81200           6   (0.0%)           0   (0.0%)
  COMPLEX                      76728           0   (0.0%)           0   (0.0%)
  COMPLEX16                    76728           0   (0.0%)           0   (0.0%)

  --> ALL PRECISIONS          315872          12   (0.0%)           0   (0.0%)

 ----------------------------- CBLAS: Default API ----------------------------- 

  SUMMARY                nb test run      numerical error          other error
  ==================   =============   ==================   ==================
  REAL                             0           0   (0.0%)           0   (0.0%)
  DOUBLE PRECISION                 0           0   (0.0%)           0   (0.0%)
  COMPLEX                          0           0   (0.0%)           0   (0.0%)
  COMPLEX16                        0           0   (0.0%)           0   (0.0%)

  --> ALL PRECISIONS               0           0   (0.0%)           0   (0.0%)
```

The fact that we do not have testing for LAPACKE appears glaring. We should work on this one of these days. LAPACKE is kinda tested with the test suite of [LAPACK++ ](https://github.com/icl-utk-edu/lapackpp), but maybe we will want our own test suite one day.

mgates3:
You could make an LAPACKE test suite _the_ test suite, and deprecate the Fortran test suite. Maybe for new routines, test them via LAPACKE rather than their Fortran interface.

Although I'm partial to the LAPACK++ test suite, being templated C++, but of course that isn't currently run when compiling LAPACK and LAPACKE.

langou:
> You could make an LAPACKE test suite the test suite, and deprecate the Fortran test suite. Maybe for new routines, test them via LAPACKE rather than their Fortran interface.

Yes, that's a good idea. 

> Although I'm partial to the LAPACK++ test suite, being templated C++, but of course that isn't currently run when compiling LAPACK and LAPACKE.

Yes, we need to think. Testing LAPACK with the LAPACK++ test suite is one way to go. That'll stress LAPACK++, LAPACKE and LAPACK (and BLAS). I like it. 

Thanks!

ACSimon33:
> The fact that we do not have testing for LAPACKE appears glaring. We should work on this one of these days. LAPACKE is kinda tested with the test suite of [LAPACK++ ](https://github.com/icl-utk-edu/lapackpp), but maybe we will want our own test suite one day.

Yes, that also bugs me. I thought about creating a new wrapper of test LAPACK routines, that call the LAPACKE routines inside, kinda what the CBLAS tests do. Then we could reroute the LAPACKE calls through the LAPACK test suite. CBLAS basically copies the BLAS test suite and renames the BLAS routines to the BLAS test routines (which wrap the CBLAS calls). I don't think that's maintainable for the LAPACK suite. I would rather adjust the extended API generator to generate a new test suite with renamed symbols (like DGELS_TEST, which calls LAPACKE_dgels).

> Testing LAPACK with the LAPACK++ test suite is one way to go. That'll stress LAPACK++, LAPACKE and LAPACK (and BLAS). I like it.

I also like this idea. How much overlap is there between the test suites? Would it make sense to use both?

ilayn:
I used the CMocka test framework in the [C translation](https://github.com/ilayn/semicolon-lapack/tree/main/tests) it took a bit of modification for tests to be performed, especially the randomization style is changed to external instead of the sequential seeds. They are not exactly transferrable but feel free to slice and dice as you see fit for the salvagable C parts.

mgates3:
The LAPACK++ test suite is roughly a subset of the LAPACK test suite. LAPACK++ is structured quite differently — each routine has its own tester (e.g., getrf, getrs), rather than a collection of related routines as in LAPACK, which tests getrf, getrs, and other LU-related routines in one tester (zchkge). LAPACK++ also reports benchmarking results. Some LAPACK++ testers just compare their output directly to LAPACKE, because that was simple to automate, but many have been re-written to use LAPACK's backward error checks. The comparison to LAPACKE implicitly tested many LAPACKE functions and found many bugs.

LAPACK++ tests some auxiliary routines, like larfg and lange, that LAPACK doesn't explicitly test. Likely these are tested by comparison to LAPACKE's output.

And of course, LAPACK++ itself is a (large) subset of LAPACK. Comparing their src directories, around 380 vs. 500 routines.

PR #1344, opened 2026-07-30
BLAS: Restore compact ROTMG flag output when no rescaling occurs

ACSimon33:
**Description**

Since 6dc1a78 (#1284, the fix for #244), `SROTMG`/`DROTMG` normalize
`DFLAG` 0/1 to -1 for *every* regular result, not just rescaled ones. That
contradicts the documented compact `DPARAM` forms (implied values "ARE NOT
STORED"), diverges from other implementations (OpenBLAS, MKL, Gonum), costs
`xROTM` its cheaper flag-0/1 paths, and fails `xblat1`'s ROTMG points 1–2,
whose expected results still encode the compact forms:

```
SROTMG COMPUTATIONAL TESTS:       81 RUN,        6 FAILED
DROTMG COMPUTATIONAL TESTS:       81 RUN,        6 FAILED
```

The commit message of 6dc1a78 describes a minimal fix — change the scaling
loops' `ELSE` to `ELSE IF (DFLAG.EQ.ONE)` so the implied-element
initialization can't clobber an already-scaled `DH12`/`DH21` — but the code
instead hoisted the normalization out of the loops and made it
unconditional. This PR implements what the message described: normalization
back inside both loops, with the guard.

### DIFF
--- a/BLAS/SRC/drotmg.f
+++ b/BLAS/SRC/drotmg.f
@@ -193,48 +193,54 @@
             END IF
          END IF
 
-*     Normalize flag to -ONE so all H elements are explicit
-*
-           IF( DFLAG.EQ.ZERO ) THEN
-              DH11 = ONE
-              DH22 = ONE
-              DFLAG = -ONE
-           ELSE IF( DFLAG.EQ.ONE ) THEN
-              DH21 = -ONE
-              DH12 = ONE
-              DFLAG = -ONE
-           END IF
-*
 *     PROCEDURE..SCALE-CHECK
-          IF (DD1.NE.ZERO) THEN
-             DO WHILE ((DD1.LE.RGAMSQ) .OR. (DD1.GE.GAMSQ))
-                 IF (DD1.LE.RGAMSQ) THEN
-                    DD1 = DD1*GAM**2
-                    DX1 = DX1/GAM
-                    DH11 = DH11/GAM
-                    DH12 = DH12/GAM
-                 ELSE
-                    DD1 = DD1/GAM**2
-                    DX1 = DX1*GAM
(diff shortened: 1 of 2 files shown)


langou:
Thanks. 

Related, I wonder why we do not do what I suggested at: https://github.com/Reference-LAPACK/lapack/pull/1000#issuecomment-2033208128. That is to replace these lines:
https://github.com/Reference-LAPACK/lapack/blob/cf47d31736e2f0e825edf8a66325192b952e8604/BLAS/SRC/drotmg.f#L115-L119
with
```fortran
DOUBLE PRECISION, PARAMETER :: GAM = 2.0D0**12
DOUBLE PRECISION, PARAMETER :: GAMSQ = 2.0D0**24
DOUBLE PRECISION, PARAMETER :: RGAMSQ = 2.0D0**(-24)
```

ACSimon33:
No objections from my side, but should that be part of this PR? I'd create a separate PR that replaces the DATA statements with PARAMETERS.

langou (approved):

langou:
@ACSimon33 wrote: "No objections from my side, but should that be part of this PR? I'd create a separate PR that replaces the DATA statements with PARAMETERS."

OKey doc. Let us do a separate PR. Good idea. I'll work on it when I can. Thanks! 

Note: there are two reasons for the future PR, (1) yes, replace DATA statements with PARAMETERS, and (2) enter accurate values for the variables GAMSQ and RGAMSQ. Right now, the variables GAMSQ and RGAMSQ are approximations of their exact values. Since we can store them exactly, we might as well store them exactly as opposed to use approximations.

PR #1357, opened 2026-08-07
CI: BLAS, CBLAS & LAPACKE coverage

ACSimon33:
**Description**

Enable coverage for BLAS, CBLAS & LAPACKE. Also update the coverage cmake modules for more modern CMake and fix some latent bugs.


ACSimon33:
I'll reopen this PR, because it is so cluttered with deployment objects...

PR #1360, opened 2026-08-07
CI: Report coverage and test results per library on Codecov

ACSimon33:
**Description**

The Codecov report was a single number for the whole repository, built from a
coverage job that measured almost nothing, and the test results were not
reported at all.

* **Coverage per library.** BLAS, CBLAS, LAPACK and LAPACKE are instrumented and uploaded separately, each from its own build directory, under a flag and a component of its own (`codecov.yml`). The test programs and tmglib are components too: a routine that never runs there is a test that was written and never reached. The README badge becomes one badge per library.
* **The coverage job now measures something.** The instrumentation is applied through one `lapack_add_coverage()` helper instead of ad-hoc per-directory code, gcov runs with `-b -c` so branch coverage exists, and the job prints a per-component table and fails when a component measured nothing — an empty flag would otherwise disappear from the report silently.
* **Test results.** The JUnit XML from `lapack_testing.py` is uploaded as a test-results report from every test job, so failures show up in the pull request comment (this is a preview feature on codecov.io and doesn't work reliably yet).  Each `<testsuite>` gained `assertions` and `time`, and the BLAS, CBLAS and DMD drivers gained the `Total time used` footer the LAPACK drivers already printed, so that every test case is timed.

`CODECOV_TOKEN` has to be set as a secret of the `codecov` environment for the uploads to be attributed.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -144,7 +144,7 @@ jobs:
       # test is exactly what the test analytics report is for.
       if: ${{ !cancelled() && matrix.os != 'windows-latest' }}
       continue-on-error: ${{ secrets.CODECOV_TOKEN == '' }}
-      uses: codecov/codecov-action@0fb7174895f61a3b6b78fc075e0cd60383518dac # v5.5.5
+      uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         # The uploader defaults to coverage; this is a test analytics report.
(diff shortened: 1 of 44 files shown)


langou (dismissed):

langou (approved):

ACSimon33:
The codecov job is expected to fail because the overall project coverage is lower now with the inclusion of LAPACKE, BLAS and CBLAS. Will rise again once we add testing for LAPACKE.

ACSimon33:
@langou Could you enable the Components and Flags analysis on Codecov.io? That way, it is easier to view coverage of the separate projects in the repo.

See here: https://app.codecov.io/gh/Reference-LAPACK/lapack/components/master
and here: https://app.codecov.io/gh/Reference-LAPACK/lapack/flags/master

Screenshot 2026-08-24 175328

PR #1365, opened 2026-08-23
TESTING/MATGEN: guard zero leading element in ZLAGHE

nakatamaho:
## Summary

`ZLAGHE` computes the phase of a Householder reflector with expressions of the form:

```fortran
WA = ( WN / ABS( A( K+I, I ) ) )*A( K+I, I )
```

When `WN` is nonzero but the leading element is exactly zero, this evaluates as `Inf * 0` and produces NaNs. This can occur at higher precision because rounding may make the leading element exactly zero while the remaining vector has a nonzero norm.

This patch handles both reflector-generation loops:

- zero norm: use `WA = ZERO`, `TAU = ZERO`;
- nonzero norm with zero leading element: use `WA = DCMPLX(WN,0)`;
- otherwise preserve the existing phase computation.

Choosing a positive-real phase is valid for a zero leading element and produces a finite reflector with `TAU = 1`.

## Motivation

The issue was observed through MPLAPACK's MPFR binary128 (`113`-bit) ZHB test with `N=20`, `K=16`, clustered eigenvalues. `ZLAGHE` returned a matrix containing NaN, which later caused `CSTEQR` to fail. The same case did not reproduce at 112 or 114 bits.

## Validation

- `gfortran -ffixed-form -ffixed-line-length-72 -fsyntax-only TESTING/MATGEN/zlaghe.f`
- The corresponding generated C++ path in MPLAPACK was tested with the 113-bit reproducer; the ZHB case completed without NaN and passed all six checks.

### DIFF
--- a/TESTING/MATGEN/zlaghe.f
+++ b/TESTING/MATGEN/zlaghe.f
@@ -175,10 +175,15 @@
 *
          CALL ZLARNV( 3, ISEED, N-I+1, WORK )
          WN = DZNRM2( N-I+1, WORK, 1 )
-         WA = ( WN / ABS( WORK( 1 ) ) )*WORK( 1 )
          IF( WN.EQ.ZERO ) THEN
+            WA = ZERO
             TAU = ZERO
          ELSE
+            IF( ABS( WORK( 1 ) ).EQ.0.0D+0 ) THEN
+               WA = DCMPLX( WN, 0.0D+0 )
+            ELSE
+               WA = ( WN / ABS( WORK( 1 ) ) )*WORK( 1 )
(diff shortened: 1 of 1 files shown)


langou (approved):

langou:
Should we push the same change for 'CLAGHE' too? if yes, @nakatamaho, can you do it?

nakatamaho:
I'll reaise a PR soon

nakatamaho:
https://github.com/Reference-LAPACK/lapack/pull/1366 here it is. Thanks!

langou:
Thanks!

PR #1367, opened 2026-08-24
README update & removal of old CI configs

ACSimon33:
**Description**
 
As mentioned in #1363, the AppVeyor builds are now completely covered by the new GitHub Actions workflow. Since the AppVeyor builds are usually the bottleneck (single runner), I'd vote to remove them, alongside the old Travis config.

@langou If you think we should keep them or just turn off AppVeyor but keep the config, I'd be ok with that as well. If we want to remove them as this PR does, you probably have to remove the AppVeyor GitHub integration.

This PR also updates the badges in the README, and adds missing token permissions in the release.yml which should increase the OpenSSF Scodecard rating.


langou (approved):

ACSimon33:
@langou Hi, it seems the AppVeyor webhook is still active and tries to build even after the config was removed. As a maintainer, you'll need to remove it, I think. It should be under **Settings** -> **Webhooks**.

langou:
Hi @ACSimon33, I just deleted the AppVeyor webhook. Cheers, Julien.

PR #1384, opened 2026-09-07
Fix rescaling after a partial failure in the xSTEBZ-based drivers; accept Inf input and xSTEBZ's negative block numbers

rmlarsen:

**Summary**

The 32 symmetric eigensolver drivers that reduce to `xSTEBZ` + `xSTEIN` (`xSTEVX`, `xSYEVX`, `xHEEVX`, `xSPEVX`, `xHPEVX`, `xSBEVX`, `xHBEVX`, `xSTEVR`, `xSYEVR`, `xHEEVR` and the `_2STAGE` variants) scale a matrix whose largest entry lies outside `[RMIN, RMAX]` and undo the scaling at the end with

```fortran
      IF( INFO.EQ.0 ) THEN
         IMAX = M
      ELSE
         IMAX = INFO - 1
      END IF
      CALL DSCAL( IMAX, ONE / SIGMA, W, 1 )
```

That block was copied from `xSTEV`/`xSYEV`, where a nonzero `INFO` comes from `xSTEQR`. In these drivers a nonzero `INFO` at that point means either that `INFO` eigenvectors failed to converge in `xSTEIN` (`INFO <= N`; the documentation says the eigenvalues are valid) or that `xSTEBZ` returned `INFO - N`. In the first case `W(INFO:M)` are returned still multiplied by `SIGMA`, that is wrong by the factor `RMIN/|T|` or `RMAX/|T|`; in the second `IMAX = N + i - 1 > N` and `xSCAL` writes past the end of `W`. Two related robustness defects sit in the same drivers: when the matrix contains an `Inf`, `SIGMA = RMAX/ANRM = 0`, so for `RANGE = 'V'` the scaled interval is `VLL = VUU = 0` and `xSTEBZ` stops the process in `XERBLA` (`parameter number 5 had an illegal value`); and `xSTEIN` rejects with `INFO = -6` (again `XERBLA`) or silently skips the negative block numbers with which `xSTEBZ` flags eigenvalues that did not converge, although every driver deliberately continues into `xSTEIN` when `xSTEBZ` returns `INFO = 1`. This PR rescales all `M` eigenvalues, leaves a matrix with an infinite norm unscaled, and makes `xSTEIN` accept `xSTEBZ`'s negative block numbers. 32 driver files plus `{s,d,c,z}stein.f`.

**Description**

*Rescaling.* At the rescale label `INFO` can be nonzero only because of `xSTEBZ` or `xSTEIN`: a failure of `xSTERF`, `xSTEQR` or `xSTEMR` earlier in the driver resets `INFO` and falls through to `xSTEBZ`. Both leave `M` valid eigenvalues in `W` (`xSTEBZ` with `INFO = 1` returns its best estimates for the unconverged ones, with `INFO = 2, 3` the `M` it found, with `INFO = 4` none), so the right count is `M` in every case and `IMAX` disappears. With finite input the wrong branch is reached whenever inverse iteration fails for at least one vector, which the drivers report as `INFO = i <= N`, "i eigenvectors failed to converge, their indices are stored in IFAIL", with no hint that eigenvalues `i .. M` are off by up to 1e146:

```
DSTEVX on T = 1e-150 * tridiag(1, 0.5, 1), ABSTOL = 0.3 |T| (so that inverse iteration fails):
BEFORE (master):
DSTEVX info =  6  ifail =  1 2 3 4 5 6
w * 1e150, computed:   9.904E-02  3.765E-01  7.775E-01  1.223E+00  1.623E+00  1.903E+04
            exact:   9.903E-02  3.765E-01  7.775E-01  1.223E+00  1.623E+00  1.901E+00
AFTER (this branch):
DSTEVX info =  6  ifail =  1 2 3 4 5 6
w * 1e150, computed:   9.904E-02  3.765E-01  7.775E-01  1.223E+00  1.623E+00  1.901E+00
            exact:   9.903E-02  3.765E-01  7.775E-01  1.223E+00  1.623E+00  1.901E+00
```

*Infinite norm.* `SIGMA` is now computed first and the scaling is applied only if it is positive, which fails exactly for `ANRM = Inf`. An infinite matrix then takes the same path as one containing a `NaN` (which has `ANRM = NaN` and is never scaled): no abort, `INFO = 0` or `N + 4` depending on `RANGE`, `NaN` in the output, nothing written beyond `W(M)`.

```
DSYEVX(RANGE = 'V') on a 6x6 matrix with A(2,2) = Inf:
BEFORE (master):     ** On entry to DSTEBZ parameter number  5 had an illegal value   (process stops)
AFTER (this branch): DSYEVX(RANGE=V) on a matrix with an Inf entry: info =   0  m =  5
```

*Negative block numbers.* `xSTEBZ` documents `IBLOCK(i) = -j` for an eigenvalue of block `j` that did not converge, and the drivers rely on `xSTEIN` to accept that array (`IF( IINFO.NE.1 ) GO TO 20` skips `xSTEIN` for every other `xSTEBZ` failure). `xSTEIN` now uses `ABS( IBLOCK( . ) )` at its four uses and documents it; the drivers then return `INFO = N + 1` together with eigenvectors, as intended, instead of stopping in `XERBLA` when a negative entry follows a positive one or returning an untouched `Z` when the last entry is negative.

**Regression test.** `xERRST` gets the second case: `xSTEVX` and `xSYEVX` with `RANGE = 'V'` on a tridiagonal matrix whose first diagonal entry is an infinity, which must return without reaching `XERBLA`. On the parent commit both report the illegal fifth argument of `xSTEBZ`, since the scale factor `RMAX / ANRM` is zero and both interval ends are scaled to zero; with the fix the matrix goes through unscaled and the drivers return.

The other two changes cannot be reached from the test suite. A partial `xSTEIN` failure needs a matrix whose norm is below the range the drivers document, and the negative block numbers need `xSTEBZ` to return `INFO = 1`, which finite in-range input cannot produce. The sweep below covers both.

**Validation**

- A sweep of 1584 cases (8 routines: `DSTEBZ`+`DSTEIN` direct, `DSTEVX`, `DSTEVR`, `DSYEVX`, `DSYEVR`, `ZHEEVX`, `DSPEVX`, `DSBEVX`; `NaN`, `+Inf`, `-Inf` in each of five positions of `D` or `E`; `n = 1, 2, 3, 5, 8`; `RANGE = 'A', 'V', 'I'`), one process per case with `XERBLA` overridden to exit, canaries around every output and work array and a 10 s timeout. Master: 280 aborts (every driver family, `Inf`, `RANGE = 'V'`), 200 cases writing three entries past the end of `W` (`RANGE = 'I'`, `Inf`, `INFO = N + 4`), 18 hangs. This branch: 0 aborts, 0 out-of-bounds writes, the same 18 hangs. The hangs are `DSTEVR`/`DSYEVR` with `RANGE = 'A'` and a `NaN` in the first two rows, inside `DLARRB` (its bracket-widening loops have no iteration bound); they are an MRRR issue independent of this PR; #1389 fixes them.
- The direct pair with a matrix outside `xSTEBZ`'s documented range (`T * 2**512`, where `e(j)**2` overflows) aborted in `DSTEIN` on master and returns `INFO = 1` from `DSTEBZ` on this branch.
- For every finite matrix inside `[RMIN, RMAX]` the drivers do not scale and this PR changes nothing; for a scaled matrix with `INFO = 0` the rescaling is the same `xSCAL( M, ... )` as before.


rmlarsen:

PR #1385, opened 2026-09-07
Drop the absolute floor from the splitting test in xSTEBZ

rmlarsen:

**Summary**

`xSTEBZ` treats an off-diagonal entry as negligible when `e(j)**2 < ulp**2 * |d(j) d(j+1)| + SAFEMN`. The `SAFEMN` term makes the test absolute: any `|e(j)|` below `sqrt(SAFEMN)` (1.5e-154 in double precision) splits the matrix whatever its neighbours are. The drivers that call `xSTEBZ` (`xSTEVX`, `xSYEVX`, `xHEEVX`, `xSPEVX`, `xHPEVX`, `xSBEVX`, `xHBEVX`, and `xSTEVR`/`xSYEVR`/`xHEEVR` for `RANGE = 'V'` or after an `xSTEMR` failure) scale a small matrix up only to `RMIN = sqrt(SAFMIN/ULP)` (1e-146), where that floor corresponds to `|e(j)|/|T| < sqrt(ULP)` = 1.5e-8. So for every matrix whose largest entry is below `sqrt(SAFMIN)/ULP` (6.7e-139) the drivers silently drop off-diagonal entries of relative size between `ULP` and 1.5e-8: the eigenvalues are wrong by up to `1.5e-8 * |T|` and the eigenvectors have residuals of the same size, all with `INFO = 0`, while `xSTEV`/`xSYEV` (QR) return them to working precision. This PR replaces the test by Kahan's relative criterion `e(j)**2 < ulp**2 * |d(j) d(j+1)|` (or `e(j) = 0`), which is what `xLARRA` already uses for the MRRR drivers, and fixes an "underflow"/"overflow" typo in the `xLAEBZ` documentation. Two code files, two documentation files.

**Description**

For `T = s * tridiag(1, r, 1)` of order 6 the eigenvalues are $`s\,(1 + 2r\cos(k\pi/7))`$ exactly. Maximum eigenvalue error divided by `n * ulp * |T|` on master; `DSTEVX(A)` uses the documented "most accurate" `ABSTOL = 2*DLAMCH('S')`:

| `s` | `r` | `DSTEV` | `DSTEVX(A)` | `DSTEVX(I)` | `DSTEVX(V)` | `DSTEVR(V)` | `DSYEVX(I)` | `DSYEVR(V)` | `ZHEEVX(I)` | `DSPEVX(I)` | `DSBEVX(I)` |
|---|---|---|---|---|---|---|---|---|---|---|---|
| 1e-146 | 1e-8 | 0.2 | 1.3e7 | 9.4e6 | 1.3e7 | 1.3e7 | 9.4e6 | 1.3e7 | 9.4e6 | 9.4e6 | 9.4e6 |
| 1e-150 | 1e-9 | 0.1 | 1.3e6 | 9.4e5 | 1.3e6 | 1.3e6 | 9.4e5 | 1.3e6 | 9.4e5 | 9.4e5 | 9.4e5 |
| 1e-290 | 1e-10 | 0.2 | 1.3e5 | 9.4e4 | 1.3e5 | 1.3e5 | 9.4e4 | 1.3e5 | 9.4e4 | 9.4e4 | 9.4e4 |
| 1e-140 | 1e-14 | 0.2 | 13 | 9.4 | 13 | 13 | 9.4 | 13 | 9.4 | 9.4 | 9.4 |

The eigenvector residuals $`\|Tz - wz\| / (\|T\|\,\|z\|\,n\,\mathrm{ulp})`$ are the same numbers to within 25%. With this branch every entry of the table is below 0.25. At `s = 1e-140` the drivers do not scale and the floor still discards `|e| = 1e-154`; at `r = 1e-6`, and at every `r` for `s >= 1e-138`, master is already accurate, which locates the affected region exactly at `|e(j)| < sqrt(SAFEMN)` next to diagonal entries below `sqrt(SAFEMN)/ULP`.

The test suite cannot see this: the smallest matrix scale in `xCHKST`/`xDRVST` is `RTUNFL*N*ULPINV = 6.7e-139 * N`, precisely the scale at which the floor stops mattering.

**Fix.** The threshold is formed as `(|d(j)| ulp) (|d(j+1)| ulp)`, which is bit-identical to `|d(j) d(j+1)| ulp**2` whenever the latter is finite and does not overflow before `e(j)**2` does, so an off-diagonal next to two entries above 1.3e154 is no longer declared negligible by an infinite threshold (that case is outside the range `xSTEBZ` documents, but the old test turned it into a wrong answer with `INFO = 0`). `e(j) = 0` splits as before. For a matrix in the normal range the only entries whose treatment changes are off-diagonals below 1.5e-154 sitting between two diagonal entries below 2e-138: they now stay in their block instead of being zeroed, their squares take part in the Sturm counts as subnormal numbers, and the twin sweep below shows the counts unaffected. A nonzero `e(j)` whose square underflows to zero still splits, which the Sturm sequence could not distinguish from zero anyway.

**Minimal reproducer**

```fortran
! DSTEVX on T = 1e-150 * tridiag(1, 1e-9, 1), n = 6, ABSTOL = 2*DLAMCH('S').
program minimal
  implicit none
  integer, parameter :: n = 6
  double precision :: d(n), e(n-1), w(n), z(n,n), work(5*n), abstol, dlamch, pi, wex(n)
  integer :: iwork(5*n), ifail(n), info, m, k
  external dlamch
  pi = 4*atan(1d0)
  d = 1d-150
  e = 1d-159
  abstol = 2*dlamch('S')
  call dstevx('N', 'A', n, d, e, 0d0, 0d0, 0, 0, abstol, m, w, z, n, work, iwork, ifail, info)
  do k = 1, n
    wex(n+1-k) = 1d-150*(1 + 2d-9*cos(k*pi/(n+1)))
  end do
  print '(a,i2)', 'DSTEVX info = ', info
  print '(a,6f14.10)', '(w - 1e-150) * 1e159, computed: ', (w - 1d-150)*1d159
  print '(a,6f14.10)', '                        exact: ', (wex - 1d-150)*1d159
end program
```

```
BEFORE (master):
DSTEVX info =  0
(w - 1e-150) * 1e159, computed:   0.0000000000  0.0000000000  0.0000000000  0.0000000000  0.0000000000  0.0000000000
                        exact:  -1.8019377711 -1.2469795971 -0.4450419127  0.4450417770  1.2469795971  1.8019377711
AFTER (this branch):
DSTEVX info =  0
(w - 1e-150) * 1e159, computed:  -1.8019377711 -1.2469797328 -0.4450419127  0.4450417770  1.2469795971  1.8019374998
                        exact:  -1.8019377711 -1.2469795971 -0.4450419127  0.4450417770  1.2469795971  1.8019377711
```

The remaining differences are one `ulp` of `|T|`.

**Regression test.** `xCHKST` gets a matrix type 22: a tridiagonal matrix with equal diagonal entries and an off-diagonal entry one tenth of the square root of an ulp of them, at `sqrt( SAFMIN / EPS )`, the smallest norm the eigenvalue drivers scale a matrix up to. The square of that off-diagonal entry underflows, which is exactly what the absolute term of the old test hid and what the relative test sees.

Two existing ratios catch it, the comparison of `xSTEBZ` against `xSTERF` and the residual of the eigenvectors `xSTEIN` builds from its blocks: on the parent commit the type fails 7 ratios per run in single precision and 8 in double, at 1e6 and above against a threshold of 60, and with the fix all of them pass. The `'I'` against `'V'` comparison is skipped for this type, because the interval that test builds is floored at twice the square root of the underflow threshold, which is wider than the whole spectrum of such a matrix, so the two ranges cannot agree either way. All four `xCHKST` files carry the type, since the complex ones call the real `xSTEBZ` as well.

Adding the type raises the number of types in `sep.in` past the count that made `ALAREQ` read its list, so `MAXTYP` for the symmetric problem goes to 22 in all four `xCHKEE` files; the two-stage checker clamps to its own 21 and ignores the new type.

**Validation**

- The full LAPACK test suite passes on this branch: 5446341 LAPACK tests, 0 numerical errors, 0 other errors (`ctest`: 100% of 215 tests passed). The 4440 tests above the parent commit are the new type.
- The sweep above (7 scales x 6 ratios x 11 routines, eigenvalues and eigenvector residuals against the closed form) is clean for every driver at every scale on this branch; the only rows still above tolerance are direct `DSTEBZ` calls at `|T| = 1e-290`, outside the range `xSTEBZ` documents (`e(j)**2` underflows to zero there), which the drivers handle by scaling.
- A power-of-two twin sweep of `DSTEBZ` + `DSTEIN` on `T * 2**k` for 42 values of `k` from -1070 to 1020, compared line by line with master: eigenvalues, block structure, `INFO` and eigenvectors are identical for every `k` at which master was correct (`-508 <= k <= 507`); at `k = -510 ... -520` master splits the matrix completely and returns the diagonal, this branch returns the eigenvalues to working precision; at `k = 512, 513` master aborts in `DSTEIN` (`XERBLA`, illegal `IBLOCK`), this branch returns `INFO = 1` from `DSTEBZ` like master does for larger `k`.

Found while auditing the tridiagonal bisection and inverse-iteration routines; the routines themselves are correct within their documented range, this is the one place where the drivers' scaling and `xSTEBZ`'s splitting disagree.


rmlarsen:

PR #1387, opened 2026-09-07
Return INFO = 1 from xLASQ1 instead of stopping in XERBLA when SIGMX is a NaN

rmlarsen:

**Summary**

`xLASQ1` scales the bidiagonal matrix by its largest entry `SIGMX` before calling the dqds algorithm. When `D` or `E` contains a NaN that survives the `MAX` reductions that compute `SIGMX` (with gfortran: a NaN in `D(N)`), `SIGMX` is a NaN and `xLASCL` rejects it by stopping the process in `XERBLA` (`On entry to DLASCL parameter number 4 had an illegal value`). `xBDSQR` without singular vectors takes this path, and so does `xGESVD` with `JOBU = JOBVT = 'N'`. This PR returns `INFO = 1` from `xLASQ1` instead, which is the same mechanism #1382 uses for `xLALSD` and `xBDSDC`. `xBDSQR` treats every nonzero `INFO` from `xLASQ1` as "try the QR algorithm", so it then finishes the QR iteration on the NaN data and reports the non-convergence through its own `INFO`, as it does for other non-finite input. Two files: `{s,d}lasq1.f`.

**Description**

```fortran
      DO 20 I = 1, N
         SIGMX = MAX( SIGMX, D( I ) )
   20 CONTINUE
      ...
      CALL DLASCL( 'G', 0, 0, SIGMX, SCALE, 2*N-1, 1, WORK, 2*N-1, IINFO )
```

Which NaN positions reach `DLASCL` depends on the compiler's `MAX` (gfortran returns the second argument when the first is a NaN, so only a NaN in `D(N)` propagates; a NaN in `D(1)` or `D(3)` is dropped and the routine returns `INFO = 0` with NaN output). The guard tests exactly the quantity `DLASCL` will reject, so it removes the abort on every compiler without changing the outcome for a NaN that does not reach `SIGMX`. `INFO = 1` is documented ("a split was marked by a positive value in E, or the input contains a NaN").

**Minimal reproducer**

```fortran
! DBDSQR without singular vectors on a 5x5 upper bidiagonal matrix with D(5) = NaN,
! and DGESVD('N', 'N') on the same matrix.
program minimal
  implicit none
  integer, parameter :: n = 5
  double precision :: d(n), e(n), a(n,n), s(n), u(n,n), vt(n,n), work(8*n*n+40), zero, c(1,1)
  integer :: info, i
  zero = 0
  d = [1d0, 2d0, 3d0, 4d0, 5d0]; e = 0.3d0; e(n) = 0
  d(n) = zero/zero
  a = 0
  do i = 1, n
    a(i,i) = d(i)
  end do
  do i = 1, n-1
    a(i,i+1) = e(i)
  end do
  call dbdsqr('U', n, 0, 0, 0, d, e, vt, n, u, n, c, 1, work, info)
  print '(a,i3)', 'DBDSQR (no vectors): info = ', info
  call dgesvd('N', 'N', n, n, a, n, s, u, n, vt, n, work, 8*n*n+40, info)
  print '(a,i3)', 'DGESVD(N, N):        info = ', info
end program
```

```
BEFORE (master):      ** On entry to DLASCL parameter number  4 had an illegal value   (process stops)
AFTER (this branch):  DBDSQR (no vectors): info =   4
                      DGESVD(N, N):        info =   4
```

**Regression test.** `xERRBD` gets the case: a 4 by 4 bidiagonal with a NaN in `D(N)`, passed to `xBDSQR` without singular vectors, which must return without calling `XERBLA`. All four files carry it, since the complex `xBDSQR` reaches the same real `xLASQ1`. It belongs with the error-exit tests rather than in `xCHKBD`, because the routine returns no meaningful output for such a matrix and the point of the test is that it returns at all; as a matrix type it would also be misleading, since `xCHKBD` counts the fixed behaviour (`IINFO > 0`) as a failure while master's `INFO = 0` with NaN output passes every `RESULT( J ) .GE. THRESH` comparison. On the parent commit the four sections report the two `xLASCL` illegal-argument lines and `*** xBDSQR called XERBLA for a matrix with a NaN ***`; with the fix they pass.

**Validation**

- 120 cases (`DBDSQR` without vectors and `DGESVD`; NaN, `+Inf`, `-Inf` in five positions; `N = 2, 3, 5, 8`), one process per case: master stops in `XERBLA` in 3 of them (NaN in `D(N)`, `N >= 3`, through `DBDSQR`), this branch in none; every other case returns the same `INFO` as on master.
- `DLASQ1` itself: NaN in `D(1)` and `D(3)` return `INFO = 0` on both; NaN in `D(5)` stops the process on master and returns `INFO = 1` here.
- Finite input never takes the new branch (`DISNAN( SIGMX )` is false), so results are unchanged.

`DGESDD` is not affected: it checks its input norm for a NaN up front and returns `INFO = -4`.

The `INFO = 0` returns for a NaN in `D(1)` or `D(3)` above are a separate defect, reported as #1392: `xLASQ3`'s deflation tests are written as "keep iterating while `Z(...) .GT. TOL2*(...)`", so a NaN operand makes every comparison false and the NaN is deflated as a converged singular value. Widening the check here from `SIGMX` to the whole bidiagonal would cover that too; this PR keeps to the position that stops the process.

This PR touches `TESTING/EIG/{s,d,c,z}errbd.f`, as #1390 does; the two blocks are independent and the textual conflict between them is a trivial one.


rmlarsen:

Difference from x86: the defect does not occur on this machine. gfortran compiles MAX to `fmaxnm`, which returns the non-NaN operand, so SIGMX never becomes a NaN and xLASCL is never reached. On master and on this branch alike, DLASQ1 returns INFO = 0 with NaN output for the reproducer (the behavior of #1392). The new xERRBD case therefore passes even without the fix here. On x86, where `maxsd` propagates the NaN, it should fail as described.

PR #1389, opened 2026-09-07
Bound the bracket widening in xLARRB and xLARRJ so that a NaN matrix cannot hang xSTEMR

rmlarsen:

**Summary**

`xLARRB` refines eigenvalues of a relatively robust representation $`LDL^T`$ by bisection, and `xLARRJ` does the same on the tridiagonal matrix itself when only eigenvalues are wanted. Before bisecting each makes sure the initial interval `[W - WERR, W + WERR]` contains the wanted eigenvalue by widening it: it moves the endpoint by `BACK`, doubles `BACK`, and repeats until the Sturm count agrees with the index. In exact arithmetic that ends quickly, because the count is 0 below the spectrum and `N` above it. A NaN pivot is never counted, so with a NaN in the representation the count stays short of the index and the loop runs forever: `DSTEMR`, and through it `xSTEVR`, `xSYEVR` and `xHEEVR` with `RANGE = 'A'`, never return for a symmetric tridiagonal matrix with a NaN in one of its first two rows, with `JOBZ = 'V'` through `xLARRB` and with `JOBZ = 'N'` through `xLARRJ`. This PR stops the widening in both routines once `BACK` has overflowed, at which point no further step can move the endpoint, and returns `INFO = 1`, which the `xLARRB` callers already propagate (`xLARRV` returns -1, `xLARRE` -4, `xSTEMR` 21 or 11, and the drivers fall back to `xSTEBZ` + `xSTEIN`). It also checks `xLARRB`'s return code at the one call site in `xLARRV` that ignored it, and `xLARRJ`'s return code in `xSTEMR`, which never looked at it and now returns `INFO = 3X` for a failure there. Twelve files: `{s,d}larrb.f`, `{s,d}larrj.f`, `{s,d,c,z}larrv.f`, `{s,d,c,z}stemr.f`.

**Description**

The loops are

```fortran
         BACK = WERR( II )
 20      CONTINUE
         NEGCNT = DLANEG( N, D, LLD, LEFT, PIVMIN, R )
         IF( NEGCNT.GT.I-1 ) THEN
            LEFT = LEFT - BACK
            BACK = TWO*BACK
            GO TO 20
         END IF
```

and its mirror image for `RIGHT`. `DLANEG` deliberately does not count a NaN pivot (`DPLUS.LT.ZERO` is false) and replaces a NaN quotient by 1 to continue the recurrence, so for a representation with a NaN in `D(1)` the count never exceeds `N - 1` for any shift, and the `RIGHT` loop for the last eigenvalue widens through `+Inf` without end. The widening also stalls when `WERR` is exactly zero, which `xLARRE` produces for an eigenvalue that the dqds algorithm returns as exactly zero (`WERR(I) = RTOL*ABS(W(I))`): then `BACK = 0`, the endpoint never moves, and the loop terminates only if the Sturm count at `W` already agrees with the index.

`xLARRJ` has the same two loops, written with a factor `FAC` that doubles and a step `WERR( II )*FAC`, and the same Sturm count (`DPLUS.LT.ZERO`), so `DSTEMR( 'N', 'A', ... )` on the same matrices never returns either: it skips `xLARRV` and refines the `xLARRE` eigenvalues with `xLARRJ` instead. `xSTEMR` ignores the `INFO` of `xLARRJ`.

**Fix.** In both routines `BACK` starts at `MAX( WERR( II ), MNWDTH )`, where `MNWDTH = 2*PIVMIN` is the minimum interval width the bisection already uses, so a zero `WERR` still makes progress; and each loop gives up with `INFO = 1` when `BACK.LT.TWO*BACK` is false, i.e. when `BACK` is infinite (or NaN), because no further step can move the endpoint. The test is false only after `BACK` has already overflowed, which a finite representation never reaches: the count is 0 once `LEFT` is below the Gershgorin bound and `N` once `RIGHT` is above it, and `BACK` doubles from at least `2*PIVMIN` to beyond the spectral diameter in at most `MAXITR` steps. `INFO` is documented in both (`= 1`: the interval could not be widened to contain the eigenvalue; the counts are inconsistent, as they are for a NaN or an Inf). `xSTEMR` now checks `xLARRJ`'s `INFO` and returns `INFO = 30 + ABS( IINFO )`, documented next to the existing `1X` (`xLARRE`) and `2X` (`xLARRV`) codes. The second `xLARRB` call in `xLARRV` (refining the extremal eigenvalues of a child cluster) now returns `INFO = -1` on failure like the first one instead of continuing with an unrefined bracket, in all four precisions: `CLARRV` and `ZLARRV` call the real `xLARRB` at the same three places as `SLARRV` and `DLARRV`, and without the check `CSTEMR` and `ZSTEMR` on the NaN matrices below go on into the representation construction and report the `xLARRF` failure (`INFO = 22`) where `SSTEMR` and `DSTEMR` report the refinement failure (`INFO = 21`).

**Minimal reproducer**

```fortran
! DSYEVR('V', 'A') on a 3x3 symmetric tridiagonal matrix with A(1,1) = NaN.
program minimal
  implicit none
  integer, parameter :: n = 3
  double precision :: a(n,n), w(n), z(n,n), work(26*n), zero, d(n), e(n), dd(n), ee(n)
  integer :: iwork(10*n), isuppz(2*n), info, m, i
  logical :: tryrac
  zero = 0
  a = 0
  do i = 1, n
    a(i,i) = i
  end do
  do i = 1, n-1
    a(i,i+1) = 0.3d0; a(i+1,i) = 0.3d0
  end do
  a(1,1) = zero/zero
  d = [a(1,1), a(2,2), a(3,3)]
  e = [0.3d0, 0.3d0, 0d0]
  dd = d; ee = e; tryrac = .true.
  call dstemr('V', 'A', n, dd, ee, 0d0, 0d0, 1, n, m, w, z, n, n, isuppz, tryrac, work, 26*n, iwork, 10*n, info)
  print '(a,i3,a,i2)', 'DSTEMR(V): info = ', info, '  m = ', m
  dd = d; ee = e; tryrac = .true.
  call dstemr('N', 'A', n, dd, ee, 0d0, 0d0, 1, n, m, w, z, n, n, isuppz, tryrac, work, 26*n, iwork, 10*n, info)
  print '(a,i3,a,i2)', 'DSTEMR(N): info = ', info, '  m = ', m
  call dsyevr('V', 'A', 'U', n, a, n, 0d0, 0d0, 1, n, 0d0, m, w, z, n, isuppz, work, 26*n, iwork, 10*n, info)
  print '(a,i3,a,i2)', 'DSYEVR: info = ', info, '  m = ', m
end program
```

```
BEFORE (master):     (no output; the first DSTEMR call never returns)
AFTER (this branch): DSTEMR(V): info =  21  m =  3
                     DSTEMR(N): info =  31  m =  3
                     DSYEVR: info =   0  m =  0
```

`DSYEVR` returns through its `DSTEBZ` + `DSTEIN` fallback, which reports no eigenvalue for the NaN matrix, the same result it gives on master for a NaN that is not in the first two rows.

**Regression test.** `xCHKST` gets the case: after the size and type loops it calls `xSTEMR` with `RANGE = 'A'` on a tridiagonal matrix whose first diagonal entry is a NaN, once with `JOBZ = 'V'` and once with `JOBZ = 'N'`, which covers `xLARRB` and `xLARRJ` respectively. Only a negative `INFO` counts as a failure, since the routine returns no meaningful output for such a matrix and the point of the test is that it returns at all. On the parent commit all four `xeigtst` binaries stop responding at that call, so `LAPACK-xeigtst{s,d,c,z}_sep_in` fail by timeout; with the fix `sep.in` passes unchanged (13464 tests for the real precisions, 11016 for the complex).

**Validation**

- The full LAPACK test suite passes on this branch: 5441941 LAPACK tests, 0 numerical errors, 0 other errors (`ctest`: 100% of 215 tests passed). The 40 extra tests over the parent commit are the new NaN calls, 5 sizes times 2 job options times 4 precisions.
- 100 NaN cases (`DSTEMR`, `SSTEMR`, `DSYEVR`, `DSTEVR`, `ZHEEVR`; NaN in `D(1)`, `D(2)`, `E(1)`, `D(N)`, `E(N-1)`; `N = 3, 5, 8, 20`), one process per case with a 10 s timeout: master hangs in 60 of them (every case with the NaN in the first two rows); this branch returns in all 100, `xSTEMR` with `INFO = 21` (48 cases) or `INFO = 11` (32 cases, when `xLARRE`'s own refinement hits the loop first), the drivers with `INFO = 0` and no eigenvalues.
- `xSTEMR` called directly in all four precisions on the 12 NaN positions in rows 1-2 (`D(1)`, `D(2)`, `E(1)`; `N = 3, 5, 8, 20`) returns `INFO = 21` in every case; `CSTEMR` and `ZSTEMR` returned `INFO = 22` before the `{c,z}larrv.f` check.
- `xSTEMR( 'N', 'A', ... )` in all four precisions on five NaN positions (`D(1)`, `D(2)`, `E(1)`, `D(N)`, `E(N-1)`; `N = 3, 5, 8, 20`), one process per case with a 10 s timeout: master hangs in 48 of the 80 cases; this branch returns in all 80, with `INFO = 31` in every case that hung (`repro/probe8n.f90`, `run8n.sh`).
- 200 finite matrices (`N = 1 ... 40`; random, glued-Wilkinson clusters, scaled to 1e-150 and to 1e100, nearly diagonal with off-diagonals 1e-9) give bit-identical eigenvalues from `DSTEMR` on master and on this branch, with `JOBZ = 'V'` and with `JOBZ = 'N'`, so the `MNWDTH` floor never changed a computed bracket in that sample.

Found while auditing the tridiagonal bisection routines (the same NaN sweep that produced #1384); `xLARRB` and `xLARRJ` are the only routines in the symmetric eigenvalue path that hang rather than returning garbage for a NaN. The `xLARRJ` half was found afterwards, when preparing a NaN test for the suite that calls `DSTEMR` with `JOBZ = 'N'` as `dchkst.f` does.


rmlarsen:

PR #1390, opened 2026-09-07
Let an exact zero always split the matrix in xBDSQR so that an Inf cannot hang it

rmlarsen:

**Summary**

`xBDSQR` with singular vectors never returns when the bidiagonal matrix contains an `Inf` in one of its first two diagonal entries, and neither do `xBDSDC` with `COMPQ = 'I'` and the SVD drivers `xGESVD` and `xGESDD` built on them. The relative-accuracy threshold `THRESH` is computed from a recurrence that produces `Inf/Inf = NaN` for such input, and with `THRESH = NaN` the block search `IF( ABSE.LE.THRESH ) GO TO 80` never recognises a split. The convergence test further down zeroes `E( M-1 )` and jumps back to the top of the loop without touching the iteration counter, the block search does not see the zero, the test zeroes it again, and so on. This PR makes an exact zero in `E` always a split, which is what it already is for every finite `THRESH` (`THRESH >= MAXITR*N*N*UNFL > 0`), so the change is a no-op for finite input and closes the cycle for a NaN threshold. Four files: `{s,d,c,z}bdsqr.f`.

**Description**

With `TOL >= 0` (relative accuracy, the default), the threshold is

```fortran
         MU = SMINOA
         DO 40 I = 2, N
            MU = ABS( D( I ) )*( MU / ( MU+ABS( E( I-1 ) ) ) )
            SMINOA = MIN( SMINOA, MU )
   ...
         THRESH = MAX( TOL*SMINOA, MAXITR*(N*(N*UNFL)) )
```

For `D(1) = Inf` the first step evaluates `Inf/Inf`, `SMINOA` becomes NaN and so does `THRESH`. In the main loop every test against `THRESH` is then false. The block search (`DO 70`) therefore never finds a split, so `LL = 1` and the whole matrix is the working block; the standard test `ABS( E( M-1 ) ).LE.ABS( TOL )*ABS( D( M ) )` (or its forward/backward relative variants) zeroes an off-diagonal entry and executes `GO TO 60`, and the loop returns to the block search, which still does not see the zero. `ITER` is only advanced by a QR sweep, so the iteration limit is never reached. Without singular vectors `xBDSQR` calls `xLASQ1` instead and returns.

The fix changes the split test to `IF( ABSE.LE.THRESH .OR. ABSE.EQ.ZERO )`. For a finite `THRESH` the added clause is implied by the first (`THRESH` is at least `MAXITR*N*N*UNFL`, which is positive), so the singular values and vectors of every finite matrix are unchanged, bit for bit. For a NaN `THRESH` every path back to the loop head now makes progress: a split shrinks the working block, a 2-by-2 block is finished by `xLASV2`, and a QR sweep advances `ITER`, so the loop ends within the existing `MAXITR*N*N` bound and the routine returns like it does for any other non-finite input.

**Minimal reproducer**

```fortran
! DBDSQR with singular vectors on a 3x3 upper bidiagonal matrix with D(1) = Inf.
program minimal
  implicit none
  integer, parameter :: n = 3
  double precision :: d(n), e(n), vt(n,n), u(n,n), c(1,1), work(4*n), zero
  integer :: info, i
  zero = 0
  d = [1d0, 2d0, 3d0]
  e = [0.3d0, 0.3d0, 0d0]
  d(1) = 1/zero
  vt = 0; u = 0
  do i = 1, n
    vt(i,i) = 1; u(i,i) = 1
  end do
  call dbdsqr('U', n, n, n, 0, d, e, vt, n, u, n, c, 1, work, info)
  print '(a,i3,a,3es11.2)', 'DBDSQR: info = ', info, '  d = ', d
end program
```

```
BEFORE (master):     (no output; the call never returns)
AFTER (this branch): DBDSQR: info =   0  d =    Infinity   3.03E+00   1.98E+00
```

**Regression test.** `xERRBD` gets the case: a 4 by 4 bidiagonal with an infinite `D(1)`, passed to `xBDSQR` with both sets of singular vectors, which must return without calling `XERBLA`. It belongs with the error-exit tests rather than in `xCHKBD`, because the routine returns no meaningful output for such a matrix and the point of the test is that it returns at all; a matrix type would additionally feed the infinity to `xBDSDC` and `xBDSVDX`, and its residual ratios would be NaN, which the `RESULT( J ) .GE. THRESH` comparison passes silently. On the parent commit all four `xeigtst` binaries stop responding at that call, so `LAPACK-xeigtst{s,d,c,z}_svd_in` fail by timeout; with the fix the four error-exit sections pass (`SBD`/`DBD` 56 tests, `CBD`/`ZBD` 39).

**Validation**

- 480 cases (`DBDSQR` with and without vectors, `SBDSQR`, `ZBDSQR`, `DBDSDC('I')`, `DGESVD`, `DGESDD`, `ZGESVD`; `+Inf`, `-Inf`, NaN in `D(1)`, `D(2)`, `D(N)`, `E(1)`, `E(N-1)`; `N = 2, 3, 5, 8`), one process per case with a 3 s timeout: master hangs in 66 (the three `xBDSQR` and `DBDSDC` for `+-Inf` in `D(1)` or `D(2)` with `N >= 3`, the three drivers for `+-Inf` in `D(1)`); this branch returns in all 480 with the same `INFO` values master produces for the non-finite cases it does return from (0, or the number of unconverged off-diagonals).
- 160 finite bidiagonal matrices (`N = 1 ... 40`; random, scaled to 1e-150 and to 1e100, nearly diagonal with an exact zero off-diagonal) give bit-identical singular values and vectors from `DBDSQR` on master and on this branch.

`xBDSQR` without vectors takes the `xLASQ1` path; that path stops the process in `XERBLA` (`DLASCL`, parameter 4) for a NaN in `D(N)` and is not changed here; #1387 fixes it.


rmlarsen:

Difference from x86: the hang does not occur on this machine. gfortran compiles MIN to `fminnm`, which returns the non-NaN operand, so SMINOA and THRESH stay finite. On master and on this branch alike, DBDSQR returns INFO = 0 with d = (Inf, 3.00, 2.00) for the reproducer. The new xERRBD case therefore passes even without the fix here. On x86, where `minsd` propagates the NaN, it should hang as described.

PR #1391, opened 2026-09-08
Undo the scaling of the least-squares solution in one step when A and B were scaled to the same end of the range

rmlarsen:

**Summary**

The six least-squares drivers `xGELS`, `xGELST`, `xGETSLS`, `xGELSY`, `xGELSD` and `xGELSS` scale `A` and `B` into `[SMLNUM, BIGNUM]` before solving and undo the two scalings on the solution one after the other. When `A` and `B` were both scaled to the same end of the range the two factors share a constant that cancels, but the first step is applied on its own and can flush a solution entry to zero (both above `BIGNUM`) or overflow it (both below `SMLNUM`) before the second step would have brought it back. For `A = 2^1023 I`, `b = (2^1023, 2^-27)`, whose solution is `x = (1, 2^-1050)`, all 24 drivers return `x = (1, 0)` with `INFO = 0`. This PR applies the quotient of the two factors in one step in that case; nothing changes when at most one of `A`, `B` is scaled or they are scaled to opposite ends.

**Description**

With `IASCL = IBSCL = 2` the drivers compute `x = x' * (BIGNUM/ANRM) * (BNRM/BIGNUM)`. The first factor is as small as 2^-54, so any entry of `x'` below 2^-1020 (2^-125 in single precision) is rounded to zero, and the second factor cannot recover it. The mirror case `IASCL = IBSCL = 1` multiplies by `SMLNUM/ANRM`, up to 2^104, first; an intermediate overflow there needs `x'` above 2^920 with `b` below `SMLNUM`, which a full-rank problem cannot produce, so it is merged for symmetry. The two mixed cases apply two factors in the same direction and are left as they are. The change was found while reviewing #1383, whose first revision copied this block.

| | `DGELS` on `A = 2^1023 I`, `b = (2^1023, 2^-27)` |
|---|---|
| master | `x = (1, 0)`, `INFO = 0` |
| this branch | `x = (1, 2^-1050)`, `INFO = 0` (exact) |

**Fix.** In each driver, when `IASCL .EQ. IBSCL .AND. IASCL .NE. 0`, call `xLASCL( 'G', 0, 0, ANRM, BNRM, ... )` once on the solution instead of the two calls; otherwise the existing per-flag calls run unchanged. `xLASCL` never overshoots its target, so the single call cannot flush or overflow an entry whose final value is representable. In `xGELSY`, `xGELSD` and `xGELSS` the rescaling of `R` or `S`, which involves the factor of `A` alone, is moved into its own `IF` after the solution block; it is unchanged. Where both factors apply, the solution is now rounded once instead of twice, so it can differ from master in the last bit. 24 files: `{s,d,c,z}{gels,gelst,getsls,gelsy,gelsd,gelss}.f`.

**Minimal reproducer** (`repro/undo_cases.f90` runs the case, and its mirror below `SMLNUM`, through all 24 drivers)

```fortran
program flush
  implicit none
  double precision :: a(2,2), b(2), work(100)
  integer :: info
  a = 0d0; a(1,1) = scale(1d0, 1023); a(2,2) = a(1,1)
  b = [scale(1d0, 1023), scale(1d0, -27)]           ! x = (1, 2^-1050)
  call dgels('N', 2, 2, 1, a, 2, b, 2, work, 100, info)
  print '(a,i0,a,2es12.4)', 'DGELS: info = ', info, '  x = ', b
end program
```

```
BEFORE (master):      DGELS: info = 0  x =   1.0000E+00  0.0000E+00
AFTER (this branch):  DGELS: info = 0  x =   1.0000E+00  8.2890-317
```

**Validation**

<details>
<summary>Exponent sweep, six drivers, four precisions, 2028 cases</summary>

`repro/ls_sweep.f90` runs each driver on a well-conditioned 6-by-4 problem with `A` scaled by 2^ka and `b` by 2^kb over a grid of exponents from the subnormal range to the overflow threshold (double: -1070 to 1021; single: -148 to 125), skipping pairs whose exact solution is not representable, and compares the solution with the unscaled twin after rescaling; `x(1)` is printed in hex for a bit-for-bit diff against master. `run_sweep.sh` reproduces the table.

| | cases | code path unchanged | of which bit-identical to master | both scaled to the same end | of which bit-identical | max deviation from the twin, master / branch | failures, master / branch |
|---|---|---|---|---|---|---|---|
| `S`/`C` drivers (12) | 86 each | 68 | 68 | 18 | 2 to 14 | 1.3e-6 / 1.3e-6 | 0 / 0 |
| `D`/`Z` drivers (12) | 83 each | 65 | 65 | 18 | 5 to 16 | 2.8e-14 / 2.8e-14 | 0 / 0 |

Every case whose code path is unchanged is bit-identical to master. In the merged cases the solution is rounded once instead of twice; the cases that still agree with master to the bit are those whose two norms have a power-of-two quotient. The twin problem has no solution entry near the underflow threshold, so the flush itself is covered by `undo_cases.f90`, not by the sweep.

</details>

**Test suite.** The full LAPACK test suite passes on this branch: 215 of 215 CTest entries, 5441901 LAPACK tests and 315872 BLAS tests with 0 numerical errors and 0 other errors, the same totals as the parent commit `f96546fc9` built and run the same way (GCC 13.3, `CMAKE_BUILD_TYPE=Release`, `BUILD_INDEX64_EXT_API=ON`).

**Regression test.** `xQRT13` scales its "scaled up" matrix to `1/(SFMIN/EPS)` = 2^969, and `xDRVLS` now scales it to a fixed 2^1016 instead, above the drivers' threshold, so that the scaling block runs at all. In the scaled-up types `xDRVLS` also multiplies the last column of the exact solution by `256*SFMIN` before forming the right-hand side, which puts it below what the first of the two undo steps can represent. `xQRT16` normalizes its residual per right-hand side, so the flushed column is not hidden by the other columns: on the parent commit the `xGELS`, `xGELST` and `xGETSLS` blocks fail between 2395 and 5041 ratios per precision, and with the fix all four precisions pass. Raising the scale also gives `IASCL` and `IBSCL` their first coverage in the six drivers.

**Why the test suite never saw it.** The least-squares tests in `xDRVLS` take their "scaled up" and "scaled down" matrices from `xQRT13`, which scales the largest entry to `1/(SFMIN/DLAMCH('Epsilon'))` = 2^969 (down: 2^-969), while the drivers scale when the largest entry is outside `[DLAMCH('S')/DLAMCH('P'), 1/that]` = [2^-970, 2^970]. The test matrices sit one binade inside the window on both sides, so `IASCL` and `IBSCL` are 0 for every matrix the suite generates and the undo block at the end of the drivers has never run under the suite. Moving the test scale one binade outward and giving one right-hand side a solution below 2^-1020 makes `xQRT16`'s per-column residual catch the flush (ratios of 10^3 to 10^5 on master against a threshold of 30); that test-suite change is prepared separately so that this fix stays a 24-file source change.

**Performance.** No change on the common path: the merged call replaces two `xLASCL` calls by one and runs only when both matrices were scaled to the same end of the range.

**Checklist**

- [x] The documentation has been updated. (No interface change; the routines' documentation does not describe the scaling.)


rmlarsen:
For the reviewers, on ordering: this PR is independent of #1383 (it was found while reviewing that PR's first revision, which had copied the two-step undo) and of #1381 and #1382, and merges cleanly with each of them in either order. It shares `{s,d,c,z}gelsd.f` with #1382, which only changes the documentation of `INFO` at the top of those files; this PR changes the undo block at the bottom, and a test merge of the two branches has no conflicts.

rmlarsen:

PR #1393, opened 2026-09-08
Test that a NaN pivot is reported by the packed symmetric-indefinite factorizations

rmlarsen:
Adds 144 tests — 24 in each of the `SSP`, `DSP`, `CSP`, `ZSP`, `CHP` and `ZHP` paths — for the NaN pivot guard merged in #1378 (f96546fc9). With the six factorizations restored to their state before that commit, every one of those six paths now dies of `SIGSEGV`, so what the suite previously could not see at all is a hard failure.

`xSPTRF` and `xHPTRF` rejected a column that cannot be pivoted on with

```fortran
IF( MAX( ABSAKK, COLMAX ).EQ.ZERO ) THEN
```

alone until #1378 added the `DISNAN( ABSAKK )` term that their dense twins `xSYTF2` and `xHETF2` have carried since 2006. Nothing in the test suite covered the difference: no matrix type contains a NaN, so the routines were only ever asked about a pivot that `MAX` can classify.

### What the test does

After the matrix-type loops of `xCHKSP` and `xCHKHP`, for each `N` in `NVAL` and each `UPLO`, factor a matrix whose only NaN sits on the diagonal of the last column the pivot loop visits — column `N` for `UPLO = 'L'`, column 1 for `UPLO = 'U'`. That column has no off-diagonal candidate, so `COLMAX` is zero and `IMAX` is never assigned; unless the pivot itself is tested for a NaN, every later comparison is false, the code reaches the 2-by-2 branch, and it writes `IPIV( K+1 )` or `IPIV( K-1 )`, one element past an end of `IPIV`, with `INFO` left at zero.

The NaN column is otherwise zero and the rest of the matrix is $I + \mathbf{1}\mathbf{1}^T$, which is positive definite, so the leading part is factored without an interchange and `INFO` must come back as the index of the NaN column. Two things are checked per case:

* `INFO` equals that index;
* `IPIV` was written only inside its own bounds. `IPIV` is passed as `IWORK( 2 )`, so the write below its first element lands in `IWORK( 1 )`, where a sentinel catches it; a second sentinel in `IWORK( N+2 )` catches the write above its last.

### Evidence

With `{s,d,c,z}sptrf.f` and `{c,z}hptrf.f` restored to f96546fc9^ and everything else at master, each of the six paths crashes on its first case, `N = 1`, where the uninitialized `IMAX` is dereferenced immediately. Running the cases one process each shows what the remaining ones do — gfortran 13.3, `-O2`:

<details>
<summary><code>DSPTRF</code>, before and after #1378, one process per case</summary>

```
===== DSPTRF, SRC at f96546fc9^ =====
 UPLO=U N=1    *** SIGSEGV ***
 UPLO=L N=1    *** SIGSEGV ***
 UPLO=U N=  2 INFO=     0 expected=   1 guards=     -1   -9999
 UPLO=L N=  2 INFO=     0 expected=   2 guards=  -9999      -2
 UPLO=U N=  3 INFO=     0 expected=   1 guards=     -1   -9999
 UPLO=L N=  3 INFO=     0 expected=   3 guards=  -9999      -3
 UPLO=U N=  5 INFO=     0 expected=   1 guards=     -1   -9999
 UPLO=L N=  5 INFO=     0 expected=   5 guards=  -9999      -5
 UPLO=U N= 10 INFO=     0 expected=   1 guards=     -1   -9999
 UPLO=L N= 10 INFO=     0 expected=  10 guards=  -9999     -10
 UPLO=U N= 50 INFO=     0 expected=   1 guards=     -1   -9999
 UPLO=L N= 50 INFO=     0 expected=  50 guards=  -9999     -50

===== DSPTRF, SRC at master =====
 UPLO=U N=  1 INFO=     1 expected=   1 guards=  -9999   -9999
 UPLO=L N=  1 INFO=     1 expected=   1 guards=  -9999   -9999
 UPLO=U N=  2 INFO=     1 expected=   1 guards=  -9999   -9999
 UPLO=L N=  2 INFO=     2 expected=   2 guards=  -9999   -9999
 UPLO=U N=  3 INFO=     1 expected=   1 guards=  -9999   -9999
 UPLO=L N=  3 INFO=     3 expected=   3 guards=  -9999   -9999
 UPLO=U N=  5 INFO=     1 expected=   1 guards=  -9999   -9999
 UPLO=L N=  5 INFO=     5 expected=   5 guards=  -9999   -9999
 UPLO=U N= 10 INFO=     1 expected=   1 guards=  -9999   -9999
 UPLO=L N= 10 INFO=    10 expected=  10 guards=  -9999   -9999
 UPLO=U N= 50 INFO=     1 expected=   1 guards=  -9999   -9999
 UPLO=L N= 50 INFO=    50 expected=  50 guards=  -9999   -9999
```

`-9999` is the sentinel; `IPIV( 0 ) = -1` for `UPLO = 'U'` and `IPIV( N+1 ) = -N` for `UPLO = 'L'` are the out-of-bounds writes. Both of the new checks fire in all ten non-crashing cases.

</details>

Whether `MAX( ABSAKK, COLMAX ).EQ.ZERO` catches a NaN is compiler- and optimization-dependent, so a build on which the old code happens to classify the pivot correctly would see this test pass. That is the same dependence #1378 removed; the test asserts the behavior the guard now makes unconditional.

`xCHKSP` and `xCHKHP` were added to the NAG `-Onopropagate` list beside the `xERRCXX` files, for the same reason: they construct a NaN.

### Validation

Test-only change; nothing under `SRC/` is touched. The full suite passes — 5442045 LAPACK and 315872 BLAS tests, 0 errors, 215 of 215 ctest cases, against 5441901 LAPACK tests on the parent. The 144 new tests are the difference.


rmlarsen:
Verified on an Apple M4 (macOS, Homebrew gfortran 16.2, Release build with the CI flags). With this branch merged onto current master, the full test suite passes. With the six factorizations restored to their state before #1378, the new tests fail in all six paths: the double and complex test binaries crash, and the single-precision one reports the wrong INFO and the write outside IPIV.

PR #1402, opened 2026-09-10
Recursive LAUUM

jprhyne:
**Description**
As discussed in PR #1373, here is the implementation of a recursive LAUUM. Since some may prefer a blocked algorithm, we are proposing that we use `*lauum.f` as a dispatch to whatever implementation is desired in a given environment. It's possible to somehow utilize `ilaenv` to do the dispatch or to have this logic inside `lauum.f` directly. 

In addition, since it is hardware dependent, I default to NX=64 as a crossover to lauu2.  Due to how ILAENV parses the (sub)-routine names, any LAUUM routine that queries for NX will return the same value. This could be relevant if an NX parameter is to be introduced into the blocked variant. I'm not sure what the preferred method to deal with this is, so I propose that we defer this decision until it becomes relevant.

But, if it is better to solve this issue now, I propose adding another variable that checks for whatever comes after underscores in ILAENV. However, since this type of naming is not convention, I do not believe it is a good idea unless there becomes a standardized way of having different variant names in the main source directory. 

Since this is a change to an auxiliary routine without a change in any public facing APIs, I do not think it is important to solve at this moment.
**Checklist**

### DIFF
--- a/SRC/CMakeLists.txt
+++ b/SRC/CMakeLists.txt
@@ -113,7 +113,7 @@ set(SLASRC
    slarz.f  slarzb.f slarzt.f slasy2.f
    slasyf.f slasyf_rook.f slasyf_rk.f slasyf_aa.f
    slatbs.f slatdf.f slatps.f slatrd.f slatrs.f slatrs3.f slatrz.f
-   slauu2.f slauum.f sopgtr.f sopmtr.f sorg2l.f sorg2r.f
+   slauu2.f slauum.f slauum_blocked.f slauum_recursive.f sopgtr.f sopmtr.f sorg2l.f sorg2r.f
    sorgbr.f sorghr.f sorgl2.f sorglq.f sorgql.f sorgqr.f sorgr2.f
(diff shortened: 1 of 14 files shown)


ACSimon33:
Hi @jprhyne,

could you please merge the current master back into all your PRs? We fixed numerical issues and errors on all platforms that are tested in CI, and the CI jobs are now configured to fail if new errors occur. That will help us to quickly determine if the changes in the PRs introduce new errors or can be merged.

mohawk2:
I don't have visibility of the setting for this PR, but the default on this repo is to give permission to those with write permission, to push (including force-push) to a PR branch. That would mean that someone with permissions on this repo could rebase this branch past the `master` updates, and force-push it. (Rebasing is often considered better practice than merging, though there is debate on this)

jprhyne:
I am working on lallm (another variant similar to this routine) and noticed that there is no statement on the unused triangular component of A inside LAUUM. I have opened an issue asking for clarification (Issue #1415). I'm more than happy to merge changes for that into this PR (after I update this branch and ensure all tests pass of course)

jprhyne:
I still need to fix documentation, I'm just running the tests after merging the new master branch before doing any other changes.

PR #1371, opened 2026-08-25
TESTING: Make the test failures fatal and resolve all of them

ACSimon33:
**Description**

The test suite could not fail. A driver that found a numerical error printed it and exited with status 0, and every fatal `STOP` was either bare or carried a character stop code, which also exits with 0 — so `ctest` and `lapack_testing.py`, which both check the status, saw a passing run. This branch closes those paths and fixes what the CI then reported.

Merge this last: the numerical failures it exposes are fixed on their own branches (see table below).

## Failures now fail

- `LAPACK_Test_Summary` runs with `--fail-on-error`, so a numerical failure or error reported by any driver fails the run.
- 177 fatal `STOP`s become `STOP 1`, across 45 files in `BLAS/TESTING`, `CBLAS/testing`, `TESTING/EIG` and `TESTING/LIN`.
- The `ALLOCATE` checks passed their message as the stop code, which goes to stderr where the script never sees it; they now write it to the output unit.

## Reports name their job

`lapack_testing.py --junit-job NAME` prefixes the JUnit suite (`job / suite`) and class (`job.class`) names, exposed to CMake as `LAPACK_TESTING_JUNIT_JOB_ID`. Codecov keys a test case on those names alone, so the reports of ~40 jobs collapsed onto one another, and a failure did not say where it came from.

## Error-exit tests

- On a Windows shared build, the drivers' own `XERBLA` cannot replace the library's, so the first deliberate illegal argument killed the driver before any test ran (CBLAS already skips its `xerbla` tests for this). `LAPACK_SKIP_ERROR_EXIT_TESTS` plus the new `lapack_test_input()` feed such a build a copy of the input with `TSTERR` off, written where `lapack_testing.py` looks for it so a hand-run driver uses the same input `ctest` does. This could be fixed by #1407.
- `?ec.in` had no `TSTERR` line at all — `?chkee` hard-coded `.TRUE.`. It is now read from the file, the line is added to `sec.in`/`dec.in`/`cec.in`/`zec.in`, and the input-format docs in `[sdcz]chkee.F` are updated.

## The numerical failures, fixed elsewhere

| Fix | Clears |
|---|---|
| #1396 | `macos-26-nagfor`, `sec`/`dec`; closes #598 | 
| #1397 | `ubuntu-26.04-gfortran`, the 803-row `cnep`/`ced`/`znep`/`zed` baseline | 
| #1398 | `STFSM` 1 of 7776, every x86-64 gfortran job; closes #679
[...]

PR #1410, opened 2026-09-11
TESTING: Improve eigenvalue consistency tests (chkhs)

ACSimon33:
## The failing test

ATfL 22.1.0, which is what the `ubuntu-24.04-arm` job uses, on the extended `_64` API shows the following error:

```
 Matrix order=   16, type=10, seed=2201,3345,1279,1121, result   8 is 3677.60
 ZHS:    1 out of  2016 tests failed to pass the threshold
```

## Summary

`xCHKHS` test 8 is documented as `| W(Z computed) - W(Z not computed) | / ( |W| ulp )`, and in the real drivers that is exactly what it is. The complex drivers have no `W2` array and compared against `W3`, which comes from `JOB = 'E'`, so they were varying `JOB` as well as `COMPZ` -- a stronger check than the description, and the one that fails on every LLVM Flang from 21 onwards. This PR restores test 8 to the `COMPZ` comparison in all four drivers, then adds the `JOB` comparison back as a new test 11 with a normalization that suits it.

## Commit 1: compare the test 8 eigenvalues at equal JOB

| drivers | before | after |
|---|---|---|
| `schkhs`, `dchkhs` | `WR1` from `('S','V')` vs `WR2` from `('S','N')` | unchanged |
| `cchkhs`, `zchkhs` | `W1` from `('S','V')` vs `W3` from `('E','N')` | `W1` vs a new `W2` from `('S','N')` |

`COMPZ` decides only whether `Z` is accumulated and never feeds back into the Hessenberg data, so the two lists agree exactly. Measured over 200 seeds and every matrix type, 126,000 ratios per precision: all zero, and `max = 0.00` with no scaling of any kind.

## Commit 2: check JOB = 'E' against JOB = 'S' as test 11

With test 8 restored, nothing compared the eigenvalues-only path against the full Schur path. The real drivers never did and the complex ones no longer do, so the new test adds that check to all four -- the first time the real drivers have had it.

It cannot be normalized by `ulp` times the largest eigenvalue. On ZHS type 10 at n=16 a relative O(ulp) perturbation of the Hessenberg matrix moves the spectrum by 1e12 to 2e14 in those units, with **200 of 200 trials over the threshold of 20**, and the eigenvalue that moves most is not a denormal straggler: it sits at 6 to 9% of the spectral radius with its own relative error 4.6% median, up to 57%.

So each difference is weighted by `s(j) = |y(j)**H x(j)| / ( ||y(j)|| ||x(j)|| )`, the reciprocal condition number of eigenvalue `j`, built from the eigenvectors tests 9 and 10 already compute. A backward error `eps` in `H` moves that eigenvalue by `eps / s(j)`, so `|dW(j)| s(j)` is the backward-error-scale quantity to test. `s(j) = 1` for a normal matrix, so the diagonal and Jordan types get the unscaled comparison.

Eigenvalues with `s(j) < sqrt(ulp)` are skipped. Those are the only ones the two paths have been seen to return in a **different order**, and the comparison is index-wise, so including them would mean comparing eigenvalues that do not correspond.

## Results

200 seeds x 21 types x 5 parameter sets x 7 orders = 126,000 ratios per precision, `THRESH` lowered to 0 so every value is recorded, flang 21.1.8 with default contraction:

| precision | old test 8 | new test 8 | new test 11 |
|---|---|---|---|
| S | 0 / 0.00 / 0 | 0 / 0.00 / 0 | 0 / 0.00 / 0 |
| D | 0 / 0.00 / 0 | 0 / 0.00 / 0 | 0 / 0.00 / 0 |
| C | 1408 / 4.868e+06 / **35** | 0 / 0.00 / 0 | 8 / 0.03 / 0 |
| Z | 1838 / 6.327e+14 / **36** | 0 / 0.00 / 0 | 0 / 0.00 / 0 |

(nonzero / max / at-or-over-20)

## Notes

- The `sqrt(ulp)` cutoff rests on measurement and on a mechanism, **not on a bound**. There is no theorem of the form "`s(j) >= c` implies the two paths agree on the ordering": the ordering follows from deflation decisions taken across the whole sweep, and a reordering among admitted eigenvalues would report a spurious difference. Every reordered eigenvalue found in 200 seeds and four precisions lay below the cutoff: by a factor of 8,900 in double but only **11 in single** precision.

### DIFF
--- a/TESTING/EIG/cchkee.F
+++ b/TESTING/EIG/cchkee.F
@@ -1842,10 +1842,10 @@
             CALL CCHKHS( NN, NVAL, MAXTYP, DOTYPE, ISEED, THRESH, NOUT,
      $                   A( 1, 1 ), NMAX, A( 1, 2 ), A( 1, 3 ),
      $                   A( 1, 4 ), A( 1, 5 ), NMAX, A( 1, 6 ),
-     $                   A( 1, 7 ), DC( 1, 1 ), DC( 1, 2 ), A( 1, 8 ),
-     $                   A( 1, 9 ), A( 1, 10 ), A( 1, 11 ), A( 1, 12 ),
-     $                   DC( 1, 3 ), WORK, LWORK, RWORK, IWORK, LOGWRK,
-     $                   RESULT, INFO )
+     $                   A( 1, 7 ), DC( 1, 1 ), DC( 1, 4 ), DC( 1, 2 ),
+     $                   A( 1, 8 ), A( 1, 9 ), A( 1, 10 ), A( 1, 11 ),
+     $                   A( 1, 12 ), DC( 1, 3 ), WORK, LWORK, RWORK,
+     $                   IWORK, LOGWRK, RESULT, INFO )
             IF( INFO.NE.0 )
      $         WRITE( NOUT, FMT = 9980 )'CCHKHS', INFO
   270    CONTINUE
--- a/TESTING/EIG/cchkhs.f
+++ b/TESTING/EIG/cchkhs.f
@@ -21,7 +21,7 @@
 *       .. Array Arguments ..
 *       LOGICAL            DOTYPE( * ), SELECT( * )
 *       INTEGER            ISEED( 4 ), IWORK( * ), NN( * )
-*       REAL               RESULT( 16 ), RWORK( * )
+*       REAL               RESULT( 17 ), RWORK( * )
 *       COMPLEX            A( LDA, * ), EVECTL( LDU, * ),
 *      $                   EVECTR( LDU, * ), EVECTX( LDU, * ),
 *      $                   EVECTY( LDU, * ), H( LDA, * ), T1( LDA, * ),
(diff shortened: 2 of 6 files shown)


langou (approved):

langou:
Hi @ACSimon33, 

Thanks for cleaning this up. This is a great fix. 

> each difference is weighted by s(j) = |y(j)**H x(j)| / ( ||y(j)|| ||x(j)|| ), the reciprocal condition number of eigenvalue j,

Very good idea.

Julien.

langou (approved):

mohawk2:
The CI is failing on `nagfor`, saying:

```
Error: /Users/runner/work/lapack/lapack/build/LAPACKE/testing/xlintsts_lapacke_TEST_sources/serrcxx_TEST.f, line 126: Invalid operand for intrinsic SQRT
Fatal Error: LAPACKE/testing/CMakeFiles/xlintsts_lapacke_core.dir/xlintsts_lapacke_TEST_sources/serrcxx_TEST.f-pp.f: Errors found during constant propagation
```

I don't know how to fix that, I'm afraid.

ACSimon33:
> The CI is failing on `nagfor`, saying:
> 
> ```
> Error: /Users/runner/work/lapack/lapack/build/LAPACKE/testing/xlintsts_lapacke_TEST_sources/serrcxx_TEST.f, line 126: Invalid operand for intrinsic SQRT
> Fatal Error: LAPACKE/testing/CMakeFiles/xlintsts_lapacke_core.dir/xlintsts_lapacke_TEST_sources/serrcxx_TEST.f-pp.f: Errors found during constant propagation
> ```
> 
> I don't know how to fix that, I'm afraid.

Hi, this will be fixed once #1371 is merged (soon). You can ignore that for now.

PR #1412, opened 2026-09-14
Make `valgrind` CI produce output while running

mohawk2:
**Description**

Currently the `valgrind` job is the slowest, which is expected (it's the same for the Perl Data Language CI). But as currently configured, it runs in masterful silence, then shows all its output once finished. This changes things so they are visible as they go along, while still capturing output as an artifact for uploading.

Tested that I have the right order of redirection:

    $ (echo out; echo err >&2) >file 2>&1 # as current for a file
    $ cat file
    out
    err
    $ (echo out; echo err >&2) 2>&1 |less # captures all

**Checklist**


ACSimon33:
@langou I have no right to approve, but this can be merged as far as I'm concerned.

martin-frbg (approved):

mohawk2:
@ACSimon33 Looks like @martin-frbg was kind enough to approve this. Thank you both!

Issue #1200, opened 2026-03-13
refactor: introduce shared CABSSQ/ZABSSQ helpers for complex squared magnitude


nakatamaho:
## Summary

This change replaces duplicated local `ABSSQ` statement-function definitions in
the complex Givens rotation related routines with shared external BLAS helpers:

- `CABSSQ` for `COMPLEX`
- `ZABSSQ` for `DOUBLE COMPLEX`

The following callers are updated:

- `BLAS/SRC/crotg.f90`
- `BLAS/SRC/zrotg.f90`
- `SRC/clartg.f90`
- `SRC/zlartg.f90`
- `SRC/clargv.f`
- `SRC/zlargv.f`

## Motivation

Several BLAS/LAPACK routines currently carry identical local `ABSSQ`
statement-function definitions of the form:

- `REAL(x)**2 + AIMAG(x)**2`
- `DBLE(x)**2 + DIMAG(x)**2`

Centralizing this tiny helper reduces duplication, simplifies maintenance, and
also reduces reliance on local statement functions in the conversion pipeline.

## Numerical behavior

This change is intended to preserve the original numerical behavior.

The helper formula is kept exactly in the same form as before:

- `CABSSQ(t) = REAL(t)**2 + AIMAG(t)**2`
- `ZABSSQ(t) = DBLE(t)**2 + DIMAG(t)**2`

In particular:

- this patch does **not** replace the helper with `ABS(t)**2`
- this patch does **not** move Blue-scaling logic into the helper
- this patch does **not** change Blue constants, scaling thresholds, or branch structure

All Blue-scaling and safe-scaling logic remains local to the existing callers.

## Notes

This is a refactoring change only.  
If later review shows that loop-local statement functions are preferred over
external helpers for performance reasons, the same cleanup can be reworked into
an include-based statement-function form without changing the caller-side logic.

## proposed implimentations
```code
      REAL FUNCTION CABSSQ(T)
*
*  -- Shared helper for complex squared magnitude ------------------------
*     Returns REAL(T)**2 + AIMAG(T)**2.
*     This intentionally does not use ABS(T)**2 and does not perform any
*     scaling. Blue-scaling remains the responsibility of the caller.
*
      COMPLEX T
      CABSSQ = REAL(T)**2 + AIMAG(T)**2
      RETURN
      END
```
```code
      DOUBLE PRECISION FUNCTION ZABSSQ(T)
*
*  -- Shared helper for double complex squared magnitude -----------------
[...]

PR #1418, opened 2026-09-15
BLAS: let SCASUM use SCABS1, as DZASUM uses DCABS1

ACSimon33:
## Description

`DZASUM` accumulates with `STEMP = STEMP + DCABS1(ZX(I))`. `SCASUM`, its single-precision counterpart, spells the same quantity out inline as `STEMP + ABS(REAL(CX(I))) + ABS(AIMAG(CX(I)))`. `SCABS1` computes exactly `ABS(REAL(Z)) + ABS(AIMAG(Z))`, so the two files were doing the same arithmetic by different routes for no stated reason.

The consequence is that `SCABS1` has no caller anywhere in the Fortran library. The only references to it on master are `BLAS/SRC/DEPRECATED/icamax.f`, which no build file lists, and the CBLAS wrappers `cblas_scabs1.c` and `scabs1sub.f`, which no test calls. `DCABS1` is reached through `DZASUM`. This PR makes `SCASUM` call the helper, which brings the two precisions into line and gives `SCABS1` a caller.

### Coverage

Measured locally with gfortran 13.3.0: `BLAS/SRC` built `-O0 -g --coverage -fno-inline`, the four `?blat1` drivers linked against it and run, then `gcov -b -n`.

| Routine | Lines before | Lines after |
| --- | --- | --- |
| `scabs1.f` | 0.00% of 3 | 100.00% of 3 |
| `dcabs1.f` | 100.00% of 3 | 100.00% of 3 |

`scasum.f` itself is unmoved at 100% of 12 lines and 71.43% of 14 branches, which is what `dzasum.f` also scores.

### This is not bit-for-bit identical

The inline form evaluates `(STEMP + |re|) + |im|`, two roundings into the accumulator per element. The helper form evaluates `STEMP + (|re| + |im|)`, one. Over 20000 random vectors of length 1000, 19376 gave a different single-precision result.

The new grouping is slightly the more accurate of the two, which is what the grouping predicts: adding the two same-magnitude parts to each other first, then once into a larger running total, rounds less than folding each part in separately. Against a `DZASUM` reference over 5000 random vectors of length 1000:

| | Inline (before) | Via `SCABS1` (after) |
| --- | --- | --- |
| mean relative error | 4.695e-07 | 3.422e-07 |
| max relative error | 2.434e-06 | 1.546e-06 |

For reference, single-precision epsilon is 1.192e-07. `SCASUM` now also rounds the same way `DZASUM` always has, so the two precisions agree on more than just the expression.

This matters because `SCASUM` is not only a user-facing routine: `CLAEIN`, `CLATRS`, `CLATPS`, `CLATBS`, `CLATDF`, `CTREVC` and `CTREVC3` call it, as do several EIG test helpers. The suite results below cover that.

### Cost

On gfortran 13.3.0, x86-64, n = 100000, timing `SCASUM` three ways:

| Variant | `-O2` | `-O2 -flto` |
| --- | --- | --- |
| inline, as on master | 0.118s | 0.119s |
| via `SCABS1`, this PR | 0.133s | 0.059s |
| inline but parenthesised, for reference | 0.060s | 0.060s |

Two separate effects are at work and they pull opposite ways.

The call itself costs about 12%, because without link-time optimization it cannot be inlined across translation units. That is precisely what `DZASUM` has always paid: substituting the inline expression back into `dzasum.f` and timing the pair the same way gives 1.13x. On that axis this change brings `SCASUM` to parity with `DZASUM` rather than introducing a new cost.

The regrouping is worth a factor of two, and it is the grouping rather than the call or the inlining that earns it. `STEMP + (|re| + |im|)` puts one dependent addition on the critical path where `STEMP + |re| + |im|` puts two, which gfortran emits as one `addss` into the accumulator instead of two. Parenthesising the inline expression, with no call at all, reaches the same 0.060s at plain `-O2` and produces bit-identical results to this PR. So the 2x seen under LTO is the grouping becoming visible once the call is inlined away, not a benefit of LTO as such.

The net is that this PR is about 12% slower than master at `-O2` and about twice as fast with LTO. `DZASUM` sits in the same place for the same reason. Only gfortran on x86-64 was timed.


ilayn:
I think this is the documented behavior and hence breaking contract and possibly backwards compatinility.

ACSimon33:
True, it would change the results. I ran some tests to see if there is some consensus among the vendors, but it seems there isn't:

n = 1000, 2000 random vectors:

| Implementation | SCASUM matches master | SCASUM mean rel. error | DZASUM matches reference | DZASUM mean rel. error |
| --- | --- | --- | --- | --- |
| reference, master | 100.0% | 5.300e-07 | 100.0% | 7.432e-16 |
| reference, this PR | 5.0% | 3.774e-07 | 100.0% | 7.432e-16 |
| OpenBLAS 0.3.34.dev | 5.3% | 4.868e-08 | 8.0% | 1.472e-16 |
| oneMKL 2026.1 | 5.7% | 4.904e-08 | 7.4% | 2.680e-16 |
| AOCL-BLAS 5.3.0 | 100.0% | 5.300e-07 | 4.5% | 1.001e-15 |
| cuBLAS 12.6.3.3 | 6.4% | 3.546e-08 | 8.3% | 7.464e-17 |

So I'm not sure whether this is documented behavior, other than it being like that for over 20 years.

I'd still like the two routines to operate similarly, so we could replace `DCABS1` in DZASUM with `TEMP + (|re| + |im|)`. That would keep the results the same while improving performance through the inlined call. It is the last `DCABS1` call in the entire library, and @langou already suggested removing these helper routines in #1200. And #1218 did exactly that for `?axpy`.

ilayn:
I happen to translate all Fortran code in SciPy and I am feeling like this also relates to the overflow concerns because I saw this definition a bit too often to attribute to mistakes. I wonder if there is any historical reasoning why this 1- norm is adopted instead of 2- for complex number magnitude since norms are equivalent if the value is not important but ordering is I can "sense" the rationale behind it.

ACSimon33:
Closing in favor of #1421

PR #1422, opened 2026-09-15
Clarifying LAUUM documentation

### LINKED ISSUE #1415: Unused Triangular component in LAUUM
For an application I am considering using LAUUM on a triangle matrix where the contract doesn't allow for modifying the part of the triangle that will not hold the final answer.

For example, if my routine computes an upper triangular matrix then the strictly lower triangular component must not be modified, however the documentation for LAUUM would suggest that this lower triangular component could be modified arbitrarily. 

I think regardless of the choice, the documentation for LAUUM should be clarified.

jprhyne:
**Description**
Updating the documentation of LAUUM routines to state the unused strictly triangular component is not referenced

Closes #1415
**Checklist**


SRC/clauum.f
```diff
@@ -67,9 +67,11 @@
 *>          A is COMPLEX array, dimension (LDA,N)
 *>          On entry, the triangular factor U or L.
 *>          On exit, if UPLO = 'U', the upper triangle of A is
-*>          overwritten with the upper triangle of the product U * U**H;
-*>          if UPLO = 'L', the lower triangle of A is overwritten with
-*>          the lower triangle of the product L**H * L.
+*>          overwritten with the upper triangle of the product U * U**T,
```
martin-frbg: Is the unrelated symbol change from H to T intentional ? H is used in section "Purpose" right above the change, so this would introduce an inconsistency in the documentation

jprhyne: That was not intentional. Thanks for the catch! Lazy copy and paste on my part. Should be fixed now!

martin-frbg (approved):

SRC/clauum.f
```diff
@@ -67,9 +67,11 @@
...
 *>          On entry, the triangular factor U or L.
 *>          On exit, if UPLO = 'U', the upper triangle of A is
-*>          overwritten with the upper triangle of the product U * U**H;
-*>          if UPLO = 'L', the lower triangle of A is overwritten with
-*>          the lower triangle of the product L**H * L.
+*>          overwritten with the upper triangle of the product U * U**H,
+*>          and the lower triangular part of A is not referenced.
+*>          If UPLO = 'L', the lower triangle of A is overwritten with
+*>          the lower triangle of the product L**H * L, and the upper
+*>          triangular part of A is not referenced
```
ACSimon33, line 74:
```suggestion
*>          overwritten with the upper triangle of the product U * U**H,
*>          and the strictly lower triangular part of A is not
*>          referenced. If UPLO = 'L', the lower triangle of A is
*>          overwritten with the lower triangle of the product L**H * L,
*>          and the strictly upper triangular part of A is not
*>          referenced.
```
 
I'd suggest saying that the **strictly** uppler/lower triangular parts are not referenced to avoid confusion. Also add a full stop at the end.

ACSimon33 (changes requested):

jprhyne:
The extra files added were from rebasing to the new master, I'll fix this shortly.

SRC/clauum.f
jprhyne: I agree an is the case now!

ACSimon33 (approved):

